PR #6, opened 2017-03-20
Add tchannel server and client

ChuntaoLu:
This PR adds tchannel server and client implementation. 

~I will soon add~ **Update**, following has been added:
- meta handler for health and version info
- integration tests 

r: @Raynos @Matt-Esch @zhenghuiwang


runtime/tchannel/struct.go
```diff
@@ -0,0 +1,67 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package tchannel
+
+import (
+	"bytes"
+	"io"
+
+	"code.uber.internal/rt/tchan-server/thriftrw/buffer"
```
Raynos: cannot reference code.uber.internal ...

ChuntaoLu: oops, good catch

ChuntaoLu: this should have been stashed, not sure how it went in...

Raynos: text editor adds imports on save and sometimes it "infers" the wrong thing to import.

Raynos:
This is a lot for one PR. 

Let's move all the meta related code to a separate PR.

Feel free to use stacked PRs

PR #9, opened 2017-03-21
[tchannel part2] Add meta endpoints and tests

ChuntaoLu:
This PR includes:
- thrift definition of the meta service
- generated thriftrw structs based on above thrift
- implementations of meta server, client and handler
- tests of the meta service

r: @Raynos @Matt-Esch @zhenghuiwang


Raynos:
This PR needs a rebase before review can be started.

PR #12, opened 2017-03-23
scripts/cover.sh: Massive speedup in coverage ( 80s -> 10s )

Raynos:
This PR improves our scripts/cover.sh to be usable.

Previously it would take >1 minute to generate a coverage report
making authoring tests really frustrating and slow.

The time taken was also O(N) with the number of tests written
due to how we integrated with go coverage.

Now the code coverage and test structure has changed to stop
fighting the tool and to get good coverage and fast test runs.

The two main changes here are to have the test_gateway cache the
binary it creates and spawns between test folders so that we 
only compile the gateway once per scripts/cover.sh run. The second
major change is to have an idiomatic go test folder structure, aka
unit tests are next to the units they test and code coverage for
unit tests only gathers code coverage on the unit.

 - Make scripts/cover.sh executable
 - Make scripts/cover.sh pretty printed / cleaned up
 - Ensure we skip non-tests in scripts/cover.sh
 - Move all unit tests out of the ./test/ folder and next to the unit
 - Update Makefile to run all the new tests.
 - Remove the -coverpkg flag that was severely slowing down coverage.
 - Update TestGateway to cache the child process binary.

r: @zhenghuiwang @Matt-Esch @ChuntaoLu @stevededalus


Raynos:
Before : 63 seconds.
After : 13 seconds.

Before : 73.48% coverage.
After : 73.48% coverage.

Raynos:
Before (travis) : 125 seconds
After (travis) : 31 seconds

Raynos:
## Where should I write my test ?

 - Does your test use `testGateway.CreateGateway()` ?
    - Your test is a child process integration test => put it in ./test/...
 - Is it a generated test for a generated endpoint
    - It should use `testGateway.CreateGateway()` => put it in $PREFIX/build/...
 - Is it a test that uses just one package ?
    - Can it be rewritten using `testGateway.CreateGateway()` ?
        - Rewrite it, => move it into ./test/...
    - Else is it a unit test to cover edge cases ?
        - Cool, put it next to the source code as an x_test.go file.

Raynos:
- `make test-only` ~6s
 - `make generate` ~8s`
 - `make lint` ~10s
 - `make cover` ~15s
 - `make test` ~23s

PR #13, opened 2017-03-24
Add a Middleware stack to zanzibar runtime [1/3]

stevededalus:
This PR adds a middleware stack to the zanzibar runtime that operates on the zanzibar requests/response types. It accepts middlewares that implement the `zanzibar.HandlerFn` interface.

`func(ctx context.Context,
		req *zanzibar.ServerHTTPRequest,
		res *zanzibar.ServerHTTPResponse)`

For middlewares that share state, the will be instantiated in the register.go file in a follow up PR. They're built using 
`func NewMiddleWare(
	gateway *zanzibar.Gateway,
	options Options,
	next zanzibar.HandlerFn) zanzibar.HandlerFn `
and state can be passed through the context to the `next` middleware or handler.

For middlewares that do not share state, they can be directly added to the middlestack and are executed in order. This is intended for both simple default middlewares(i.e. logging, tracing)  as well as being used as helper functions for debugging and unit testing.


glide.lock
```diff
@@ -28,17 +28,19 @@ imports:
 - name: github.com/mailru/easyjson
   version: 9d6630dc8c577b56cb9687a9cf9e8578aca7298a
   subpackages:
-  - bootstrap
```
Raynos: Changed from alphabetical to not alphabetical :(

Not sure what tool is doing this...

glide.yaml
```diff
@@ -36,7 +36,3 @@ import:
   version: a689eb3bc4b53af70390acc3cf68c9f549b6b8d6
 - package: github.com/buger/jsonparser
   version: 5b691c8ebc4af5191baa426561d62f1b5115d6e5
-testImport:
```
Raynos: does glide support test imports ? what is the correct way of doing it ?

scripts/gocov-to-istanbul-coverage.js
```diff
@@ -64,6 +63,10 @@ function parseFunction(folderName, functionInfo) {
     };
     fileObj.fnCounter++;
 
+    // TODO(sindelar): Investigate
+    if (functionInfo.Statements == null) {
```
Raynos: Thanks, no idea what that is.

runtime/middlewares.go
```diff
@@ -0,0 +1,196 @@
...
+
+import (
+	"context"
+	"net/http"
+)
+
+// Handler is an interface that compatable middlewares must implement to be used
+// in the middleware stack. Handle should yield to the next middleware in the stack
+// by invoking the next HandlerFn passed in.
+type Handler interface {
```
Raynos: This middleware interface is not very performant or debuggable.

We do not want people to author middlewares as closures. Generally writing closures instead of structs with methods is less performant and harder to debug because state is stored as closed over variables instead of fields on structs.

Having a wrapping style middleware makes reasoning about the performance profile and flamegraphs difficult.

Currently our flamegraphs look something like this :

image

If we wrap each middleware as a closure like this each middleware will add at least one new stack frame, eventually it will look something like this :

image

We want to avoid all of this noise in the stack frames so it's best if middlewares were a struct with a HandleRequest() / HandleResponse() interface. We then need a middleware runner that loops over the middlewares and calls them one by one and then finally calls the endpoint once all the middlewares are done.

runtime/middlewares.go
```diff
@@ -0,0 +1,196 @@
...
+		next HandlerFn) {
+		// TODO(sindelar): Implement this with a recorder and context propogation.
+		handler.ServeHTTP(res.responseWriter, req.httpRequest)
+		next(ctx, req, res)
+	}
+}
+
+// DefaultStack returns a new middleware stack instance with the default middlewares.
+//
+// Logger - Request/Response Logging
```
Raynos: Although logger and tracer are useful examples of middlewares.

we might be better off for performance/debugging reasons to implement them as methods on the ServerReq / ServerRes class.

runtime/middlewares.go
stevededalus: Ok, I am going to setup a version that iteratively calls interfaces instead of passing the functions through.

This approach will break defers and context scoping though.

For instance, a middleware that performs an action with a database connection or file handle and closes it on defer will never be called if a lower level middleware panics or times out.

Raynos: It can still work, just not using `defer` itself.

If a middleware's `HandleRequest()` is called then the middlewares `HandleResponse()` will be called. You can store some state on the middleware handle instance in `HandleRequest()` and explicitly clean it up in `HandleResponse()`

As for panic safety, if something panics then the process is in an undefined state and we have to restart it, I don't know if we can have panic safety for all code in the codebase.

I'm not familiar enough to talk about context scoping but it sounds like the middlewares and endpoints will all have the same context managed by the router/server itself.

runtime/middlewares.go
```diff
@@ -0,0 +1,117 @@
...
+
+// Middlewares returns a list of all the handlers in the current MiddlewareStack.
+func (m *MiddlewareStack) Middlewares() []MiddlewareHandle {
+	return m.middlewares
+}
+
+// MiddlewareHandle used to define middleware
+type MiddlewareHandle interface {
+	// implement HandleRequest for your middleware.
+	HandleRequest(
```
Raynos: We should pass ctx into the `HandleRequest` and `HandleResponse`

examples/example-gateway/middlewares/example/example.go
```diff
@@ -0,0 +1,85 @@
...
+func (m exampleMiddleware) OwnState() interface{} {
+	return m.middlewareState
+}
+
+// HandleRequest handles the requests before calling lower level middlewares.
+func (m exampleMiddleware) HandleRequest(
+	req *zanzibar.ServerHTTPRequest,
+	res *zanzibar.ServerHTTPResponse,
+	shared zanzibar.SharedState) error {
+	m.middlewareState = MiddlewareState{
```
Raynos: This `middlewareState` is re-used for all requests to one endpoint.

We need to find some way to store "per request" state.

runtime/middlewares.go
stevededalus: Done

glide.yaml
stevededalus: updated glide and it seems to be working now.

runtime/middlewares.go
```diff
@@ -0,0 +1,123 @@
...
+		ctx context.Context,
+		res *ServerHTTPResponse,
+		shared SharedState) error
+	// return any shared state for this middleware.
+	JSONSchema() *jsonschema.Document
+	Name() string
+}
+
+// SharedState used to access other middlewares in the chain.
+type SharedState struct {
```
zhenghuiwang: Do we want to add more fields in `SharedState`? why not `type SharedState map[string]interface{}`?

runtime/middlewares.go
```diff
@@ -0,0 +1,123 @@
...
+	middlewareDict map[string]interface{}
+}
+
+func newSharedState(middlewares []MiddlewareHandle) SharedState {
+	sharedState := SharedState{}
+	sharedState.middlewareDict = make(map[string]interface{})
+
+	for i := 0; i < len(middlewares); i++ {
+		sharedState.middlewareDict[middlewares[i].Name()] = middlewares[i]
```
Raynos: Should we initialize this to `nil` ?

runtime/middlewares.go
```diff
@@ -0,0 +1,123 @@
...
+// Handle executes the middlewares in a stack and underlying handler.
+func (m *MiddlewareStack) Handle(
+	ctx context.Context,
+	req *ServerHTTPRequest,
+	res *ServerHTTPResponse) {
+
+	shared := newSharedState(m.middlewares)
+
+	for i := 0; i < len(m.middlewares); i++ {
+		err := m.middlewares[i].HandleRequest(ctx, req, res, shared)
```
Raynos: Add TODO: "A middleware needs to be able to return a bool to say that it has written to the response"

Consider not allowing middlewares to return `error`. When an `error` is returned its ambiguous about who should write to the response.

runtime/middlewares.go
```diff
@@ -0,0 +1,123 @@
...
+			// zanzibar.HandleFn to return an error and let
+			// router process those at that level.
+			return
+		}
+	}
+
+	m.handle(ctx, req, res)
+
+	for i := len(m.middlewares) - 1; i >= 0; i-- {
+		err := m.middlewares[i].HandleResponse(ctx, res, shared)
```
Raynos: Add TODO: "A middleware needs to be able to return a bool to say that it has written to the response"

runtime/middlewares/logger/logger.go
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
Raynos: This file does not implement new middleware interface, either refactor or delete.

examples/example-gateway/middlewares/example/example.go
```diff
@@ -0,0 +1,86 @@
...
+func NewMiddleWare(
+	gateway *zanzibar.Gateway,
+	options Options) zanzibar.MiddlewareHandle {
+	return &exampleMiddleware{
+		options: options,
+	}
+}
+
+// HandleRequest handles the requests before calling lower level middlewares.
+func (m exampleMiddleware) HandleRequest(
```
zhenghuiwang: Should be `m *exampleMiddleware`. `*exampleMiddleware` doesn't implement the middleware interface if you define the method not on pointer.

examples/example-gateway/middlewares/example/example.go
```diff
@@ -0,0 +1,86 @@
...
+	shared zanzibar.SharedState) error {
+	shared.SetState(
+		m.Name(),
+		MiddlewareState{
+			Baz: m.options.Foo,
+		})
+	return nil
+}
+
+func (m exampleMiddleware) HandleResponse(
```
zhenghuiwang: same here

examples/example-gateway/middlewares/example/example_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNoRequestSuccessfulRequestOKResponse(t *testing.T) {
+	schema := (&exampleMiddleware{}).JSONSchema()
+
+	// TODO(sindelar): Read expected from test file
+	expected := "{\n" +
```
zhenghuiwang: you can use ` `` ` as the outmost quotes for raw string to avoid type so many escapes

examples/example-gateway/middlewares/example/example.go
stevededalus: done

examples/example-gateway/middlewares/example/example.go
stevededalus: done

runtime/middlewares/logger/logger.go
stevededalus: fixed

examples/example-gateway/middlewares/example/example_test.go
stevededalus: Thanks for the tip. I'm going to change this to read from the golden files in a follow up, but in the future will use that approach.

PR #16, opened 2017-03-25
Initial module system with HTTP client generation

Matt-Esch:
This is a first sketch of how the module system will work (the tests show an example of how it is defined). It doesn't yet generate a build directory, the generator interface is TBD. It does read the service dir based on the configuration. I want the runner to be exhaustive so that all errors can be accumulated and displayed on read/build errors.
The main idea is that we will use this configuration to generate a build, running code generators during this build step. We will use the module system to generate only those files that are required for the target service build. This isn't yet a full build system as there may be pre/post generation steps. This is a modular build generator, not a full build system.
For closed source code, the default zanzibar module structure will be extended with custom classes and code generators. The current design of the system should allow us to place classes inside nested directories, such as the Entity class in ./business/entities.
### Terminology
<dl>
<dt>Class:</dt> <dd>A class of module i.e. endpoint/client/service/middleware/library.</dd>
<dt>Class "Type":</dt> <dd>Within a class, an implementation specific identifier, such as "http" endpoint, "tchannel" client</dd>
<dt>Class Type "Instance":</dt> <dd>A specific Class Type definition on disk, such as the "health" http endpoint.</dd>
<dt>Instance "Configuration":</dt> <dd>The files inside a module directory. The minimum required configuration is a $class-config.json file.<dd>
<dt>Generator:</dt> <dd>An inteface that can produce files from the module instance configuration</dd>
</dl>


codegen/runner/runner.go
```diff
@@ -59,6 +59,23 @@ func main() {
 		err, fmt.Sprintf("cant get abs path of config dir %s", configDirName),
 	)
 
+	packageHelper, err := codegen.NewPackageHelper(
```
Raynos: Can we have only one package helper ? `gatewaySpec` also creates one.

codegen/runner/runner.go
```diff
@@ -77,8 +94,9 @@ func main() {
 		err, fmt.Sprintf("can't create gateway spec %#v", gatewaySpec),
 	)
 
-	fmt.Printf("Generating clients for gateway \n")
-	err = gatewaySpec.GenerateClients()
```
Raynos: What about the "clients.go" init file ?

codegen/module_system.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, errors.Wrapf(
+			err,
+			"Error generating HTTP client from template for %s",
+			instance.InstanceName,
+		)
+	}
+
+	structsBuffer := bytes.NewBuffer(nil)
+	structsWriter := bufio.NewWriter(structsBuffer)
+	if err := generator.templates.template.ExecuteTemplate(
```
Raynos: nitpick: We should have a consistent err pattern in the codebase.

Currently we both use

```
err = x();
if err != nil {
    return nil, err
}
```

and

```
if err := x(); err != nil {
    return nil, err
}
```

module/module.go
```diff
@@ -0,0 +1,501 @@
...
+type JSONClassConfig struct {
+	Name         string              `json:"name"`
+	Config       interface{}         `json:"config"`
+	Dependencies map[string][]string `json:"deps"`
+	Type         string              `json:"type"`
+}
+
+// GeneratorOptions provides extra information from the buid system
+// to the code generators
+type GeneratorOptions struct {
```
Raynos: This structs appears to be unused.

module/module.go
```diff
@@ -0,0 +1,501 @@
...
+
+// Instance is a configured module on disk inside a module classes
+// directory. For example, this could be
+//     ClassName:    "Endpoint,
+//     ClassType:    "http",
+//     InstanceName: "health",
+type Instance struct {
+	ClassName            string
+	ClassType            string
+	BaseDirectory        string
```
Raynos: some more comments about what `BaseDirectory` and `Directory` are exactly would be useful.

codegen/module_system.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, errors.Wrapf(
+			err,
+			"Error generating HTTP client structs from template for %s",
+			instance.InstanceName,
+		)
+	}
+
+	clientDirectory := filepath.Join(
+		instance.BaseDirectory,
+		"build",
```
Raynos: This is hard coded here but actually comes from gateway.json.

We should avoid duplicate source of truth, either zanzibar respects "targetGenDir" ( https://github.com/uber/zanzibar/blob/master/examples/example-gateway/gateway.json#L6 ) or is always hard coded to write to "./build/"

module/module.go
```diff
@@ -0,0 +1,501 @@
...
+				buildPath,
+				moduleIndex,
+				moduleCount,
+			)
+
+			classGenerators := moduleSystem.classes[classInstance.ClassName]
+			generator := classGenerators.types[classInstance.ClassType]
+
+			if generator == nil {
+				fmt.Printf(
```
Raynos: Should this skip warning be an error in the future once we've ported everything onto the module system ?

module/module.go
```diff
@@ -0,0 +1,501 @@
...
+					classInstance.ClassType,
+					err.Error(),
+				)
+				return err
+			}
+
+			for filePath, content := range files {
+				filePath = filepath.Clean(filePath)
+
+				if strings.HasPrefix(filePath, "..") {
```
Raynos: This guards for relative paths. What if a generator returns an absolute path ?

module/module.go
```diff
@@ -0,0 +1,501 @@
...
+					return errors.Errorf(
+						"Module %s generated a file outside the build dir %s",
+						classInstance.Directory,
+						filePath,
+					)
+				}
+
+				resolvedPath := filepath.Join(
```
Raynos: What does `filepath.Join()` do if the second argument is an absolute file path ?

codegen/runner/runner.go
Matt-Esch: I imagine that gatewaySpec goes away when the module system fully generates the required code. Open to suggestions about how to share the packageHelper but it doesn't seem unusual to have 2 while we have 2 separate build systems working together.

codegen/module_system.go
Matt-Esch: I don't actually see why it would help to stadardize this. Both patterns are equally valid. I tend to use the if wrapper when the statement returns an error only and avoid declaring variables. This means in cases where you have multi return you generally avoid the wrapped if, and in cases where you only return err, you use it.

module/module.go
Matt-Esch: It gets joined to a base path, so you can't return absolute paths. I'm cleaning the filePath which puts all ../ at the front of the path. This prevents circumventing the fact you can only write files to the build directory. Both .. and absolute paths don't work.

module/module.go
Matt-Esch: You can always verify this for yourself if you're uncertain https://play.golang.org/p/NIj6Iibg3Z

module/module.go
Matt-Esch: It's a question of whether or not the module system defined has to be exhaustive or not. In this case I am skipping "custom" clients. It originally caused a panic after rebasing on a change that added that client type. I think it makes sense to assume module directories are "fully managed" and therefore this would be an error in future.

codegen/runner/runner.go
Raynos: can pass this into GatewaySpec constructor.

Creates less confusion refactors etc, for example zhenghui has a pending PR that refactors the `NewPackageHelper()` constructor.

codegen/module_system.go
```diff
@@ -0,0 +1,199 @@
...
+	if err := clientWriter.Flush(); err != nil {
+		return nil, errors.Wrapf(
+			err,
+			"Error generating HTTP client from template for %s",
+			instance.InstanceName,
+		)
+	}
+
+	structsBuffer := bytes.NewBuffer(nil)
+	structsWriter := bufio.NewWriter(structsBuffer)
```
zhenghuiwang: bytes.Buffer is already a io.writer. you don't need `bufio.NewWriter`?

module/module.go
zhenghuiwang: A suggestion on handle errors exhaustive errors:

define a customized error type `AggregatedErrors`which implements `Error()` interface and holds several errors:

type struct AggregatedErrors {
   Errors []error
}

You can make it add non-nil errors and `Error()` print out all errors.

module/module.go
```diff
@@ -0,0 +1,501 @@
...
+	types             map[string]BuildGenerator
+}
+
+// BuildGenerator provides a function to generate a module instance build
+// artifact from its configuration as part of a build step. For example, an
+// Endpoint module instance may generate endpoint handler code
+type BuildGenerator interface {
+	Generate(
+		instance *Instance,
+	) (map[string][]byte, error)
```
zhenghuiwang: Since `Instance` has all info needed, why hold all file contents in bytes instead of writing on disk directly? 

The check whether the file path is valid can be moved into Generate() too.

Matt-Esch: Generate is hand-written code per module definition. If we exemplify writing stuff to disk here, it will encourage others to do the same. With this structure we can say (in code review) "never manually write files to disk in the code generators". If we do it ourselves, it will be much harder to take this stance.

Also it will also make testing easier as you can test code generators and templates without having to write to disk necessarily. Decoupling these responsibilities yields this advantage as we're unlikely to want to DI a filesystem when we realize that writing lots of content to disk at test time inflates the test times or causes parallel test conflicts.

module/module.go
Matt-Esch: Yeah I was thinking about using an error aggregating module. I want this to be as clear as possible. I actually think the error system might not be the best approach. I want to know what errored out, why it happened and what I can do to fix it if there are reasonable suggestions. It's a bit more involved than your average production code runtime errors.

module/module.go
```diff
@@ -0,0 +1,501 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// moduleType enum defines whether a ModuleClass is a singleton or contains
+// multiple directories with multiple configurations
+type classType int
+
+const (
```
Matt-Esch: I'm not sure what you are suggesting here, this isn't valid go syntax. "Const declaration cannot have type without expression"

Matt-Esch:
I have separated this pr so the module system code will land before it is integrated with the client generator.

module/module.go
Matt-Esch: It's also wrong to say the Instance has all the information required. I'm actually building to a hard coded directory composed from the BaseDirectory, the Directory and the "build" directory.. This is a slight regression compared with how we currently do it, it's a TBD.

I would like to remove BaseDirectory from the instance as it shouldn't be required, but it is necessary to produce the clients at this point I believe.

module/module.go
zhenghuiwang: I missed the second `iota`. those two constants you defined are `int` not `classType`

const (
// SingleModule defines a module class type that has 1 directory
SingleModule classType = iota
// MultiModule defines a module class type with multiple nested directories
MultiModule classType = iota
)

Matt-Esch: good catch

PR #17, opened 2017-03-27
Implement resp.PeekBody()

Raynos:
Adds a method to `ServerHTTPResponse` that allows for a 
middleware to peek at the body that is about to be flushed.

The intent is that the endpoint will write the body,
the response middlewares will run.
And the middleware runner will Flush() the res after
all middlewares run.

 - Refactored `WriteJSON()` to not write immediately.
 - Added `Flush()` which will flush body.

r: @uber/zanzibar-team


runtime/middlewares.go
```diff
@@ -107,7 +107,7 @@ func (m *MiddlewareStack) Handle(
 		// then abort the rest of the stack and evaluate the response
 		// handlers for the middlewares seen so far.
 		if ok == false {
-			for j := i; i >= 0; i-- {
+			for j := i; j >= 0; j-- {
```
stevededalus: embarrassing

runtime/middlewares.go
```diff
@@ -117,6 +117,9 @@ func (m *MiddlewareStack) Handle(
 	m.handle(ctx, req, res)
 
 	for i := len(m.middlewares) - 1; i >= 0; i-- {
-		m.middlewares[i].HandleResponse(ctx, res, shared)
+		ok := m.middlewares[i].HandleResponse(ctx, res, shared)
+		if ok == false {
```
stevededalus: remove and remove bool from signature.

Raynos: Fixed.

PR #18, opened 2017-03-27
Register middlewares [2/3]

stevededalus:
Register middlewares on the endpoint handler stack based upon the endpoint json configuration files.

Endpoint configuration files can now take:
```
`"middlewares": [
		{"name" : "example",
		 "options" : {
			 "Foo": "\"test\""
		 }
		},
		{"name" : "logger"}
		]
````

Gateway config files can now define a middleware directory that contains their custom middlewares
`	"middlewareDir": "./middlewares"
`

### CODE REVIEW

File: examples/example-gateway/gateway.json [hunk 1]
```diff
@@ -6,5 +6,6 @@
 	"targetGenDir": "./build",
 
 	"clientConfig": "./clients",
-	"endpointConfig": "./endpoints"
+	"endpointConfig": "./endpoints",
+	"middlewareDir": "./middlewares"
```
Raynos: This folder will come out of ./vendor/ in the future.

codegen/runner/runner.go
```diff
@@ -86,4 +88,9 @@ func main() {
 	fmt.Printf("Generating main.go for gateway \n")
 	err = gatewaySpec.GenerateMain()
 	checkError(err, "Failed to generate main.go.")
+
+	fmt.Printf("Generating middleware schemas for gateway \n")
+	err = gatewaySpec.GenerateMiddlewareSchemas()
```
Raynos: What does generating middleware schemas mean ?

codegen/gateway.go
```diff
@@ -526,6 +584,36 @@ func NewGatewaySpec(
 		return nil, errors.Wrap(err, "cannot create template")
 	}
 
+	// TODO(sindelar): Filter test files.
+	middlewareFiles, err := filepath.Glob(filepath.Join(
```
Raynos: We may not have any "open source" middlewares, we can remove this if that is the case.

codegen/gateway.go
```diff
@@ -526,6 +584,36 @@ func NewGatewaySpec(
...
+	}
+	// Build custom middleware definitions
+	customMidFiles, err := filepath.Glob(filepath.Join(
+		configDirName,
+		middlewareDir,
+		"*",
+		"*.go",
+	))
+	if err != nil {
+		return nil, errors.Wrap(err, "Cannot load middlewares.")
```
Raynos: nitpick: Cannot load custom middlewares

codegen/gateway.go
```diff
@@ -206,6 +208,40 @@ func NewHTTPClientSpec(jsonFile string, clientConfigObj map[string]string, h *Pa
...
+	// Read the go middleware definitions.
+	fset := token.NewFileSet()
+	f, err := parser.ParseFile(fset, goFile, nil, parser.PackageClauseOnly)
+	if err != nil {
+		return nil, errors.Wrapf(
+			err, "Could not read go file %s: ", goFile,
+		)
+	}
+
+	// TODO(sindelar): Consider adding other middleware validation here.
```
Raynos: We might want to have a json module definition file for each middleware instead of parsing the go source code.

zhenghuiwang: +1. The goFile doesn't have to be in zanzibar, the middleware can come from another package.

codegen/gateway.go
```diff
@@ -362,10 +398,29 @@ func NewEndpointSpec(
 		)
 	}
 
-	// TODO(sindelar): Use structures and validate middleware are defined in codebase.
-	middlewareStrings := make([]string, len(endpointConfigObj["middlewares"].([]interface{})))
+	// TODO(sindelar): Validate middleware are defined in codebase.
+	middlewares := make([]MiddlewareSpec, len(endpointConfigObj["middlewares"].([]interface{})))
 	for idx, middleware := range endpointConfigObj["middlewares"].([]interface{}) {
```
zhenghuiwang: there are several type assertions here, if anyone fails, the program panic. it is not ideal: this function is supposed to throw an error.

Use assertion like
`middlewares, ok := endpointConfigObj["middlewares"].([]interface{})` and throw specific error if `!ok`

codegen/gateway.go
```diff
@@ -650,3 +749,16 @@ func (gateway *GatewaySpec) GenerateMain() error {
 	)
 	return err
 }
+
+// GenerateMiddlewareSchemas will generate the middleware schema files
+// for the gateway
+func (gateway *GatewaySpec) GenerateMiddlewareSchemas() error {
```
zhenghuiwang: I may not have enough context here, but I don't think zanzibar has to handle middleware schema. Here is what i think will happen:

1. Middleware developer creates a go package for a middleware(can put it any repo), and put the middleware options schema in a known repo for UI to locate the its `name`, `Go package path` and `schema`.

2. endpoint creators are shown a page for editing middleware `options` based on schema, verify it is a valid schema(possibly run Zanzibar also) and save it as a JSON file in the gateway repository.

Zanzibar only needs to generate code with  `name`, `Go package path` and `options` regardless where the option comes from.

codegen/runner/runner.go
stevededalus: I wanted to generate the json schemas from the middleware.Options structs, but based on Zhenghui's feedback that we should allow external dependencies, I'm shifting the configuration format around. It will be up to the middleware author to generate the schema and include it in the gateway configs.

codegen/gateway.go
stevededalus: Done.

codegen/gateway.go
stevededalus: Done.

codegen/gateway.go
stevededalus: removed.

codegen/gateway.go
stevededalus: Refactored to use a schema instead of trying to generate one so we can rely on external packages, as you recommended. PTAL.

examples/example-gateway/middlewares/middleware-config.json
```diff
@@ -0,0 +1,17 @@
+[
```
Raynos: Let's make this file an object with a nested field.

The work @Matt-Esch is doing on the module system requires configs to be objects.

```
{
  middlewares: [...]
}
```

PR #19, opened 2017-03-28
Fixup header validation semantics

Raynos:
Currently on master our header semantics were broken.

We have an annotation for "mandatory" headers that was
implemented in code generation to mean "headers to propagate".

I've refactored the client to temporarily remove header
propagation for now and updated the endpoint to verify
that the mandatory headers exist.

 - add support for headers in `TestGateway#MakeRequest()`
 - add `ServerRequest#CheckHeaders`
 - refactor tests to pass

r: @uber/zanzibar-team

### DIFF
--- a/codegen/templates/endpoint.tmpl
+++ b/codegen/templates/endpoint.tmpl
@@ -39,7 +39,7 @@ func Handle{{title .Name}}Request(
 	clients *clients.Clients,
 ) {
 	{{ if .Headers -}}
-	if ok := req.CheckHeaders({{.Headers | printf "%#v" }}); !ok {
+	if !req.CheckHeaders({{.Headers | printf "%#v" }}) {
 		return
 	}
 	{{- end}}
--- a/codegen/templates/endpoint.tmpl
+++ b/codegen/templates/endpoint.tmpl
@@ -38,17 +38,15 @@ func Handle{{title .Name}}Request(
 	res *zanzibar.ServerHTTPResponse,
 	clients *clients.Clients,
 ) {
-	// Handle request headers.
-	h := http.Header{}
 	{{ if .Headers -}}
-	for _, header := range {{.Headers | printf "%#v"}} {
-		h.Set(header, req.Header.Get(header))
+	if ok := req.CheckHeaders({{.Headers | printf "%#v" }}); !ok {
+		return
 	}
 	{{- end}}
 
 	// Handle request body.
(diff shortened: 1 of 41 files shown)


test/lib/bench_gateway/bench_gateway.go
```diff
@@ -189,13 +189,16 @@ func (gateway *BenchGateway) Backends() map[string]*testBackend.TestBackend {
 
 // MakeRequest helper
 func (gateway *BenchGateway) MakeRequest(
-	method string, url string, body io.Reader,
+	method string, url string, headers map[string]string, body io.Reader,
```
zhenghuiwang: Nit: better use `http.Header` instead of map[string]string

codegen/templates/endpoint.tmpl
```diff
@@ -38,25 +38,23 @@ func Handle{{title .Name}}Request(
 	res *zanzibar.ServerHTTPResponse,
 	clients *clients.Clients,
 ) {
-	// Handle request headers.
-	h := http.Header{}
 	{{ if .Headers -}}
-	for _, header := range {{.Headers | printf "%#v"}} {
-		h.Set(header, req.Header.Get(header))
+	if ok := req.CheckHeaders({{.Headers | printf "%#v" }}); !ok {
```
zhenghuiwang: Nit: Simplify it to`if !req.CheckHeader()` ?

runtime/server_http_response_test.go
```diff
@@ -163,7 +163,7 @@ func TestCallWriteJSONWithBadJSON(t *testing.T) {
 		),
 	)
 
-	resp, err := gateway.MakeRequest("GET", "/foo", nil)
+	resp, err := gateway.MakeRequest("GET", "/foo", nil, nil)
```
stevededalus: This method signature is starting to take a lot of arguments. If we add any more we should consider accepting an Options argument that contains theses fields. That was when you call Make request
```
MakeRequest(&Options{
   method: "Get",
   path: "/foo"
}) 
```
which gives us better readability and nullable arguments.

test/lib/bench_gateway/bench_gateway.go
Raynos: Want to make all the header interfaces for http/tchannel the same where possible so going to use `map[string]string` in all 6 locations.

runtime/server_http_response_test.go
Raynos: Good idea, will refactor to options next time i touch it.

PR #21, opened 2017-03-28
change client config shape to match new module system

Matt-Esch:
The new module system will expect class type configs to have a nested "config" property. Putting this up as a separate clean PR.

### DIFF
--- a/codegen/gateway.go
+++ b/codegen/gateway.go
@@ -87,6 +87,23 @@ type ClientSpec struct {
 	ThriftServiceName string
 }
 
+// moduleClassConfig represents the generic JSON config for
+// all modules. This will be provided by the module pacakge.
+type moduleClassConfig struct {
+	Name   string      `json:"name"`
+	Type   string      `json:"type"`
+	Config interface{} `json:"config"`
+}
+
+// httpClientClassConfig represents the specific config for
+// an http client. This is a downcast of the moduleClassConfig.
+// Config here is a map[string]string but could be any type.
+type httpClientClassConfig struct {
+	Name   string            `json:"name"`
+	Type   string            `json:"type"`
+	Config map[string]string `json:"config"`
+}
+
 // NewClientSpec creates a client spec from a json file.
 func NewClientSpec(jsonFile string, h *PackageHelper) (*ClientSpec, error) {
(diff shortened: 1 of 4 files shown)


codegen/gateway.go
```diff
@@ -101,32 +101,46 @@ func NewClientSpec(jsonFile string, h *PackageHelper) (*ClientSpec, error) {
...
 	}
 
-	clientType := clientConfigGeneric["clientType"].(string)
-	if clientType == "http" {
+	className := classConfig["name"].(string)
+	classType := classConfig["type"].(string)
+
+	clientConfig := classConfig["config"].(map[string]interface{})
+
+	if classType == "http" {
```
Raynos: Should this type be global ? aka "httpClient" or is the type per module aka "http" type for the client module.

codegen/gateway.go
```diff
@@ -101,32 +101,46 @@ func NewClientSpec(jsonFile string, h *PackageHelper) (*ClientSpec, error) {
...
 		// A better solution for client configs is needed here.
 		// Assuming a string map is too restrictive
-		clientConfigObj := map[string]string{}
-		err := json.Unmarshal(bytes, &clientConfigObj)
+		httpClientConfig := map[string]string{}
 
+		serializedConfig, err := json.Marshal(clientConfig)
```
zhenghuiwang: It is not efficient to marshal and then unmarshal to get the `clientConfig` here.

Define `classConfig` with explicit type info:

` 
type ClassConfig struct {
    Name string
    Type string
    ClientConfig map[string]string
}
`

You can get all info without type assertions

Matt-Esch: The JSON library doesn't offer an intermediate object with an Unmarshal interface. It would be great if it did. What I am trying to do is a downcast from the outer config type, and without writing the manual reflection code inline here, it's not really possible without the pre-serialize. We could obviously write a function that did the reflection generically, but I don't see much point when this isn't exactly hot path code.

The best thing I can do is supply the raw json bytes for the config and then try re-marshaling the entire config into a specific type.

codegen/gateway.go
Matt-Esch: types are per class, so think of class and type as a composite primary key

PR #23, opened 2017-03-28
Refactor client generation to take gateway as arg.

Raynos:
This prep allows us to get access to the `gateway.Logger`
in follow up PRs where we add a `HTTPClientRequest` class.

r: @uber/zanzibar-team

### DIFF
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -47,7 +47,6 @@ func NewClient(
 // {{title .Name}} calls "{{.HTTPPath}}" endpoint.
 func (c *{{$clientName}}) {{title .Name}}(ctx context.Context, r *{{.RequestType}}) (*http.Response, error) {
 	// Generate full URL.
-	// TODO: (jakev) insert params if needed here.
 	fullURL := c.client.BaseURL
 	{{- range $k, $segment := .PathSegments -}}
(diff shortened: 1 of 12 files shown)


codegen/templates/http_client.tmpl
```diff
@@ -67,15 +64,15 @@ func (c *{{$clientName}}) {{title .Name}}(ctx context.Context, r *{{.RequestType
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/json")
-	return c.Client.Do(req.WithContext(ctx))
+	return c.client.Client.Do(req.WithContext(ctx))
```
zhenghuiwang: Nit:`c.client.Client` repeats `client`. using 'zanzibar.HTTPClient' as anonymous field could remove one word 'client'.

Raynos: Good point, I'd rather avoid inheriting http.Client as public API for now. 

naming these things is a bit hard.

ChuntaoLu:
Is passing in a gateway pointer necessary if all we need is logger access? Do we predict other use cases with gateway other than logger?

Raynos:
@ChuntaoLu `gateway.MetricsScope` as well, its used for logger, metrics, jaeger tracing, etc.

PR #24, opened 2017-03-28
Add ClientHTTPRequest class.

Raynos:
This PR is refactoring the client so that it becomes transport
agnostic. As a first step we are adding a ClientHTTPRequest
class that can managed the sending component.

As a follow PR we will add a ClientHTTPResponse class that
can parse the response and the body.

In the future, the client API will no longer return a 
*http.Response but just return the struct itself.

r: @uber/zanzibar-team

### DIFF
--- a/runtime/client_http_request.go
+++ b/runtime/client_http_request.go
@@ -1,0 +1,126 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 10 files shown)


runtime/client_http_request.go
```diff
@@ -0,0 +1,126 @@
...
+func NewClientHTTPRequest(
+	clientName string, methodName string,
+	client *HTTPClient,
+) *ClientHTTPRequest {
+	req := &ClientHTTPRequest{
+		Logger: client.Logger,
+		client: client,
+	}
+
+	req.start(clientName, methodName)
```
ChuntaoLu: Is the start time intended to be when a client request is created, rather than when it's sent?

Raynos: There are many timers that are interesting to us.

For example "when an httpclientrequest is started, including serialize/deserializer" and "when a http request is send until a response is received, just network IO".

TChannel adds the start timer at request creation time ( https://github.com/uber/tchannel-go/blob/3abb4c025c1663b383452339a22d918cf9d0be0b/outbound.go#L119 ).

ChuntaoLu: Makes sense.

runtime/client_http_request.go
```diff
@@ -0,0 +1,126 @@
...
+
+	req.ClientName = clientName
+	req.MethodName = methodName
+
+	req.started = true
+	req.startTime = time.Now()
+}
+
+// WriteJSON will send a json http request out.
+func (req *ClientHTTPRequest) WriteJSON(
```
zhenghuiwang: The naming is interesting here:

`NewClientHTTPRequest` doesn't construct a request with body, destination url etc, but `WriteJSON` does. It suggests `NewClientHTTPRequest` actually returns a "client" (you can call `WriteJSON` multiple times to send out different requests.

Consider rename "ClientHTTPRequest" to "HTTPClient" or make the 
`NewClientHTTPRequest` takes additional arguments to create one request, similar to `http.NewRequest`

Raynos: I'm looking to have symmetry with HTTPServerResponse class.

A client http request is definitely 1-1 with a single outbound http.Request.

I can rename this to `OutgoingHTTPRequest` instead of `ClientHTTPRequest` if that reduces confusion.

zhenghuiwang: My question is why separate the construction of a request into two functions: `NewClientHTTPRequest` and `WriteJSON`? (The definition of ClientHTTPRequest indicates all it needs to construct a request.)

Based only on this API without looking at the code,
```
request := NewClientHTTPRequest(...)
request.Do(...) 
```
is a valid usage. but it is actually not.  If `WriteJSON` is required to call exactly once after `NewClientHTTPRequest`, it should be inside `NewClientHTTPRequest` and not be exposed  publicly.

Raynos: I don't like having so much logic / side-effects in the `NewX()` allocator. 

A constructor function should be thin in it's behaviour. `request.Do()` will throw a panic if you didnt call `WriteJSON`.

Maybe what makes sense is `req := httpClient.request(...); req.Do();` helper.

PR #25, opened 2017-03-29
Add copyrights to generated Go files in Zanzibar

stevededalus:
Add the ability to define a copyright-header.txt file in gateway.json that will be applied to the top of generated code.


codegen/template.go
```diff
@@ -637,11 +662,26 @@ func (t *Template) GenerateMainFile(
 	}, nil
 }
 
-func (t *Template) execTemplateAndFmt(templName string, filePath string, data interface{}) error {
+func (t *Template) execTemplateAndFmt(
+	templName string,
+	filePath string,
+	data interface{},
+	copyrightHeader string,
```
zhenghuiwang: Strictly speaking, it is `copyRightHeaderFile` not the "copyRightHeader" string itself.

Also, we should allow the file to be `""` and set "// Generated by Zanzibar ..." by default.

To be more robust, zanzibar should throw error if the header file doesn't have `//` at the beginning of each line.

examples/example-gateway/gateway.json
```diff
@@ -7,5 +7,6 @@
 
 	"clientConfig": "./clients",
 	"endpointConfig": "./endpoints",
-	"middlewareConfig": "./middlewares/middleware-config.json"
+	"middlewareConfig": "./middlewares/middleware-config.json",
+	"copyrightHeader": "./copyright_header.txt"
```
Raynos: Putting this in gateway.json seems to "leak" zanzibar responsibility to user.

We can just have one copyright that we default to.

codegen/runner/runner.go
```diff
@@ -70,7 +70,7 @@ func main() {
 		config.MustGetString("clientConfig"),
 		config.MustGetString("endpointConfig"),
 		config.MustGetString("middlewareConfig"),
-
+		config.MustGetString("copyrightHeader"),
```
Raynos: We can just hardcode a path to copyright_txt here.

codegen/template.go
```diff
@@ -637,11 +662,26 @@ func (t *Template) GenerateMainFile(
...
+) error {
 	file, err := openFileOrCreate(filePath)
 	if err != nil {
 		return errors.Wrapf(err, "failed to open file: %s", err)
 	}
+
+	_, err = io.WriteString(file, copyrightHeader)
```
Raynos: It looks like we need to write the "generated" by comments first and then the copyright otherwise it breaks the github UI of collapsing generated code.

codegen/template.go
stevededalus: Rename to File when it's the file and string when its the string.

codegen/template.go
stevededalus: Done.

codegen/runner/runner.go
stevededalus: I wanted this to be definable by the user so that if someone uses the open sourced zanzibar to generate a gateway, then the files build based on their specifications would have their copyright attached instead of the zanzibar/uber copyright header (just for the generated files, not for the generators or runtime).

examples/example-gateway/gateway.json
stevededalus: I want whoever's using zanzibar to generate gateways to be able to attach their own copyright header, since the copyright for the generated code should be attached to whoever defines the specification (but not fro the runtime or generators)

codegen/gateway.go
```diff
@@ -768,8 +769,19 @@ func NewGatewaySpec(
...
 	middlewareConfig string,
+	copyrightHeaderFile string,
 	gatewayName string,
 ) (*GatewaySpec, error) {
+
+	copyright, err := ioutil.ReadFile(filepath.Join(
+		configDirName,
+		copyrightHeaderFile,
+	))
+	if err != nil {
```
Raynos: Let's make this optional, if it doesnt exist then copy right header is empty string.

examples/example-gateway/build/clients/bar/bar.go
```diff
@@ -1,6 +1,3 @@
-// Code generated by zanzibar
-// @generated
-
 package barClient
```
Raynos: Why is this gone ^ we need to add this back.

codegen/test_data/clients/bar_structs.gogen
```diff
@@ -1,6 +1,3 @@
-// Code generated by zanzibar
```
Raynos: We want to add this back here.

codegen/gateway.go
stevededalus: Done.

codegen/test_data/clients/bar_structs.gogen
stevededalus: Done.

examples/example-gateway/build/clients/bar/bar.go
stevededalus: Done.

codegen/templates/http_client.tmpl
```diff
@@ -1,6 +1,6 @@
 {{- /* template to render edge gateway http client code */ -}}
-// Code generated by zanzibar
-// @generated
+// Code generated by zanzibar
```
Raynos: some of the templates have comments other donts :( this is confusing.

codegen/template.go
```diff
@@ -200,12 +201,21 @@ func (t *Template) GenerateClientFile(
 		IncludedPackages: m.IncludedPackages,
 		ClientID:         c.ClientID,
 	}
-	err := t.execTemplateAndFmt("http_client.tmpl", c.GoFileName, clientMeta)
```
Raynos: This is not being used and is instead used by the module system. 

As part of cleanup, can you delete this method and add the copyrightHeader to the module system ?

stevededalus: Done

codegen/templates/http_client.tmpl
stevededalus: Its because the client code was executed through a different path. Fixed now.

PR #26, opened 2017-03-30
[TChannel part0] Add tchannel test backend

ChuntaoLu:
This PR adds the tchannel test backend to facilitate testing endpionts backend by tchannel downstream.

r: @Raynos

### DIFF
--- a/test/lib/test_backend/test_tchannel_backend.go
+++ b/test/lib/test_backend/test_tchannel_backend.go
@@ -1,0 +1,127 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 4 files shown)


test/lib/test_backend/test_tchannel_backend.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Close closes the underlying channel
+func (backend *TestTChannelBackend) Close() {
+	backend.Channel.Close()
+}
+
+// Wait keeps backend process live
+func (backend *TestTChannelBackend) Wait() {
+	backend.WaitGroup.Wait()
```
Raynos: We do not need this `WaitGroup`, or at least tchannel server doesn't have a way to wait on the server.

ChuntaoLu: right, I thought we need it for the usage as https://github.com/uber/zanzibar/blob/master/benchmarks/benchserver/main.go#L52

Raynos: I see, that only makes sense if somebody calls `.Add(1)` and `.Done()` 

`.Wait()` immediately returns if no-one calls `.Add()`

ChuntaoLu: Done

PR #27, opened 2017-03-30
[TChannel part1] Add Baz tchannel client

ChuntaoLu:
This PR adds Baz client/endpoint thrift files, generated code, and manual Baz tchannel client/server implementation.

This PR is stacked on top of https://github.com/uber/zanzibar/pull/26

r: @Raynos @zhengguiwang


codegen/templates/init_clients.tmpl
```diff
@@ -9,6 +9,9 @@ import (
 	"{{$pkg}}"
 	{{end}}
 	"github.com/uber/zanzibar/runtime"
+
+	// TODO: (lu) remove this, to be generated
```
Raynos: This is a neat way of adding hand written code without breaking the code gen :)

examples/example-gateway/config/production.json
```diff
@@ -20,6 +20,9 @@
 	"clients.bar.port": 4001,
 	"clients.bar.ip": "127.0.0.1",
 
+	"clients.SimpleService.port": 4002,
```
Raynos: lower case for consistency ?

examples/example-gateway/idl/github.com/uber/zanzibar/endpoints/baz/baz.thrift
```diff
@@ -0,0 +1,29 @@
...
+
+exception SimpleErr {
+  1: required string message
+}
+
+exception NewErr {
+  1: required string message
+}
+
+service SimpleService {
```
Raynos: This endpoint will need http annotations for it to work ?

runtime/tchannel_server.go
```diff
@@ -27,15 +27,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/uber-go/zap"
 	"github.com/uber/tchannel-go"
-)
```
Raynos: Should we move this file into the tchannel folder ?

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,237 @@
+// Code generated by zanzibar
+// @generated
```
Raynos: If this file is not generated then dont add the generated comments ;)

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// NewClient returns a new http client for service Bar.
+func NewClient(config *zanzibar.StaticConfig, gateway *zanzibar.Gateway) *BazClient {
+	// TODO: (lu) get clientName from config
+	// clientName := config.MustGetString("clients.baz.clientName")
+	clientName := "BazClient"
+
+	// TODO: (lu) client channel opts
+	ch, err := tchannel.NewChannel(clientName, nil)
```
Raynos: We do not want to allocate new channels here.

There needs to be a `gateway.Channel` and we should allocate a sub channel here.

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		gateway.Logger.Error("Failed to create tchannel",
```
Raynos: If creating a client can fail then we need to refactor our client interfaces.

We should have something like

```
NewX(config, gateway) *XClient
(*XClient) Bootstrap() error
(*XClient) Close()
```

Aka, if a client has some bootstrapping logic that can fail we need an explicit bootstrap method ( or init method, methodname TBD. )

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,237 @@
...
+			zap.String("error", err.Error()))
+	}
+
+	ip := config.MustGetString("clients.SimpleService.ip")
+	port := config.MustGetInt("clients.SimpleService.port")
+	ch.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+
+	// TODO: (lu) get serviceName from config
+	// serviceName := config.MustGetString("clients.baz.serviceName")
```
Raynos: ServiceName is a thrift detail so its known from the thrift file and/or client-config.json.

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,237 @@
...
+	return &BazClient{
+		SimpleServiceClient{
+			thriftService: serviceName,
+			client:        client,
+		},
+	}
+}
+
+// SimpleServiceClient is the client to talk to SimpleService backend.
+type SimpleServiceClient struct {
```
Raynos: Can we merge this client with `BazClient` ?

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,237 @@
...
+		default:
+			err = errors.New("received no result or unknown exception for SimpleFuture")
+		}
+	}
+
+	return respHeaders, err
+}
+
+// SimpleServiceServer is the Baz backend service
+type SimpleServiceServer struct {
```
Raynos: Who uses `SimpleServiceServer` ? It looks like this code doesn't belong here.

Can we move it to `endpoints/baz/baz.go` or something ?

examples/example-gateway/clients/baz/baz.go
ChuntaoLu: yeah, I removed the comment in the following PR https://github.com/uber/zanzibar/pull/30/files#diff-50f740ccde9b3f6a874ee45ca46b1613L53

examples/example-gateway/clients/baz/baz.go
ChuntaoLu: make sense, will make changes accordingly

examples/example-gateway/clients/baz/baz.go
ChuntaoLu: if we pass in a top level channel via gateway and not allocate new channel  here, we can probably don't have failure cases.

examples/example-gateway/clients/baz/baz.go
ChuntaoLu: Yes

examples/example-gateway/clients/baz/baz.go
ChuntaoLu: the test backend use it to create mock backends. Yeah, it does not belong here, but I don't think `endpoints/baz/baz.go` is a good fit either. I should have put a todo here.

examples/example-gateway/config/production.json
ChuntaoLu: I imagine we will pull out the service name from thrift or client config, and use it as key to get from the above config, https://github.com/uber/zanzibar/pull/27/files#diff-50f740ccde9b3f6a874ee45ca46b1613R49. We can follow convention here, but a conversion is needed in the middle somewhere, which I am not in favor of.

And the `clients` prefix in `clients.xxxx.port` does not really make sense, it's clearly downstream service configs. In this sense, it's arguably more consistent as is the same as the service name defined in thrift.

examples/example-gateway/idl/github.com/uber/zanzibar/endpoints/baz/baz.thrift
ChuntaoLu: Right, I am deferring it a bit

runtime/tchannel_server.go
ChuntaoLu: Yeah, we can even merge it into `runtime/tchannel/server.go`, which is probably even better?

examples/example-gateway/config/production.json
Raynos: We need to use something from client-config.json, like the "name" field ( https://github.com/uber/zanzibar/blob/master/examples/example-gateway/clients/googlenow/client-config.json#L2 ).

The thrift file is not the source of truth, the client-config is.

runtime/tchannel_server.go
Raynos: cool, sounds good.

examples/example-gateway/clients/baz/baz.go
Raynos: cool.

runtime/tchannel_server.go
ChuntaoLu: I will make the merge when updating part 3 PR.

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,215 @@
...
+// NewClient returns a new http client for service Bar.
+func NewClient(config *zanzibar.StaticConfig, gateway *zanzibar.Gateway) *BazClient {
+	// TODO: (lu) get serviceName from thrift
+	serviceName := "SimpleService"
+
+	ch := gateway.Channel
+
+	ip := config.MustGetString("clients.baz.ip")
+	port := config.MustGetInt("clients.baz.port")
+	ch.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
```
Raynos: We cannot mutate the `Peers()` of the top channel.

We need to allocate a sub channel.

runtime/gateway.go
```diff
@@ -91,12 +94,19 @@ func CreateGateway(
 		logWriter = opts.LogWriter
 	}
 
+	// TODO: (lu) channel name and opts
+	ch, err := tchannel.NewChannel("Gateway", nil)
```
Raynos: This needs to be allocated in `setupChannel()` not here.

runtime/gateway.go
```diff
@@ -353,11 +356,19 @@ func (gateway *Gateway) setupHTTPServer() error {
 }
 
 func (gateway *Gateway) setupTChannel(config *StaticConfig) error {
+	// TODO: (lu) channel name and opts
+	ch, err := tchannel.NewChannel("Gateway", nil)
```
Raynos: That's not really what I meant. This is duplicate with `runtime/tchannel_server.go:56`

we should use `tchannelServer.Channel`.

Never allocate two channels.

examples/example-gateway/clients/baz/baz.go
```diff
@@ -0,0 +1,215 @@
...
+	// TODO: (lu) get serviceName from thrift
+	serviceName := "SimpleService"
+
+	sc := gateway.Channel.GetSubChannel(serviceName)
+
+	ip := config.MustGetString("clients.baz.ip")
+	port := config.MustGetInt("clients.baz.port")
+	sc.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+
+	client := zt.NewClient(gateway.Channel, serviceName)
```
Raynos: Got to pass the sub channel in here otherwise it wont work.

Raynos: Apparently we have to pass both in because `ch.RunWithRetry` ...

ChuntaoLu: the allocated sub channel is acquired by calling `channel.GetSubSchannel(serviceName)` in `zt.NewClient` call

Raynos: I guess that's good enough.

runtime/tchannel_server.go
```diff
@@ -72,8 +68,7 @@ func NewTChannelServer(opts *TChannelServerOptions) (*TChannelServer, error) {
 			err)
 	}
 
-	return &TChannelServer{
-		Channel: channel,
-		Logger:  opts.Logger,
-	}, nil
+	gateway.Channel = channel
```
Raynos: mutating the gateway here is naughty, can we move this into `setupChannel()`.

Keeping mutation of the gateway to gateway.go makes debugging easier.

Raynos: Sounds good.

PR #28, opened 2017-03-30
Add logger, endpoint and handler name to response to use in middlewares

stevededalus:
Add logger, endpoint and handler name to response so that these can use used in the 

`	HandleResponse(
		ctx context.Context,
		res *ServerHTTPResponse,
		shared SharedState,
	)`

methods.

### DIFF
--- a/runtime/server_http_response.go
+++ b/runtime/server_http_response.go
@@ -43,7 +43,11 @@ type ServerHTTPResponse struct {
 	pendingBodyObj    interface{}
 	pendingStatusCode int
 
-	StatusCode int
+	Logger zap.Logger
+
+	StatusCode   int
+	EndpointName string
+	HandlerName  string
 }
 
 // NewServerHTTPResponse is helper function to alloc ServerHTTPResponse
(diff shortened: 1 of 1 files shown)


Raynos:
Should the middleware runner just pass the `req` in ? 

Having duplicate data fields between `req` and `res` will create confusion.

stevededalus:
That's an option, but right now the Response object holds a reference to the Request object, so we could just expose these as methods in the response body. 

You asked me on the last diff to remove Request from HandleResponse's arguments, was the reason for that other than cleaning the interface? It does seem like sneaking the logger and endpoint properties into the request object is obfuscating some flow since zap.logger isn't request scoped.

Raynos:
I dont remember saying removing request from HandleResponse.

To avoid duplication we can do : 

 - Add `request` to `HandleResponse()`
 - Add `Response.Request` field.

the fact that zap.logger is not request scoped is just not implemented yet, need to do it.

stevededalus:
PTAL

PR #29, opened 2017-03-31
Add ClientHTTPResponse, refactor http client API

Raynos:
This PR adds the ClientHTTPResponse class.

It also refactors our generated http clients to return
structs instead of `*http.Response`. This creates a thrift
based API for our http clients.

In passing I had to clean up our aliased imports statementes.

Doesn't pass tests yet, needs a litte more work. 

r: @uber/zanzibar-team


codegen/templates/endpoint.tmpl
```diff
@@ -67,7 +67,7 @@ func Handle{{title .Name}}Request(
 	)
 	{{end}}
 	if err != nil {
-		req.Logger.Error("Could not make client request",
+		req.Logger.Warn("Could not make client request",
```
ChuntaoLu: what's the rationale here?

Raynos: @ChuntaoLu 

Currently if we get a 400/500 "application error". Instead of parsing it as a struct we just return an "error".

So not all errors here should be Error().

Once we can type switch on the err and get some of the structs out then we upgrade it back to an Error() for a subset of cases.

```
	Messages:   	{"level":"error","ts":"2017-04-01T00:46:22Z","msg":"Could not make client request","error":"Could not parse client(bar) json: parse error: invalid character 'b' after top-level value near offset 0 of 'bad bytes'"}
```

Take for example this error. One of our downstream clients returned an invalid payload. That is a warning because it's not an error in OUR code.

```
error() is meant to be used when you want to print error messages that concern application or business logic. These messages should just record that a "wrong thing" has happened.

You should use error() whenever something incorrect or unhandlable happens.

If your in a code path that is uncommon but still correct consider using warn() instead.
```

When a client returns bad data we should "handle it", in this case by returning a 500. 

An "error" logline is all about something unhandled / buggy / broken that needs to go on our backlog / issue / fix it now plz list.

runtime/client_http_request.go
```diff
@@ -82,7 +85,7 @@ func (req *ClientHTTPRequest) start(
 
 // WriteJSON will send a json http request out.
 func (req *ClientHTTPRequest) WriteJSON(
```
stevededalus: Maybe change method name to  WriteJSONReq or WriteHTTPJSON, to make it clear you're writing the whole request with headers and not just the json body.

stevededalus: Consider renaming method to WriteJSONReq or WriteHTTPJSON to make it clearer that it's the whole request with headers and not just the json body.

PR #30, opened 2017-03-31
[TChannel part2] Add manual baz endpoint and tests

ChuntaoLu:
This PR addes the baz endpoint implementation and tests, client initialization and endpoint registration are temporarily manually added to the template, and will be removed in the future.

This PR is stacked on top of https://github.com/uber/zanzibar/pull/27


examples/example-gateway/clients/baz/baz.go
```diff
@@ -66,11 +68,21 @@ func NewClient(config *zanzibar.StaticConfig, gateway *zanzibar.Gateway) *BazCli
...
 	client        zt.TChanClient
+	// timeout is the default timeout if upstream does not have one
+	timeout time.Duration
 }
 
 // Call ...
 func (c *SimpleServiceClient) Call(ctx context.Context, reqHeaders map[string]string, args *baz.SimpleService_Call_Args) (map[string]string, *baz.BazResponse, error) {
 	var result baz.SimpleService_Call_Result
+
+	_, ok := ctx.Deadline()
```
Raynos: This looks like a timeout check.

Every `ctx` should have a `Deadline()` I think it's best if we move this logic here ( https://github.com/uber/zanzibar/blob/03e905277787c3ec2670e12cd000d4d7878fbd18/runtime/router.go#L166 ).

When we have a new incoming request to the server. we should have a default timeout for the entire flow.

Now as for tchannel itself,  we want to allocate a `ContextBuilder` ( https://godoc.org/github.com/uber/tchannel-go#ContextBuilder ).

We want to set timeout and timeoutPerAttempt as well as the ParentContext.

So even if the ctx has a known Deadline() we still want to set a timeout on the context builder.
We will have to write a test for timeout propagation ( aka if ParentContext Deadline() is 2s and you set timeout to 5s does it pick 5s or 2s ).

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call.go
```diff
@@ -0,0 +1,66 @@
...
+			)
+			res.SendError(500, errors.Wrap(err, "could not make client request:"))
+			return
+		}
+	}
+
+	response := convertCallClientResponse(clientResp)
+
+	// TODO: (lu) decide response status code
+	statusCode := 200
```
Raynos: this comes from the thrift file.

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call_test.go
```diff
@@ -0,0 +1,96 @@
...
+			getDirName(), "..", "..", "build", "main.go",
+		),
+	})
+	if !assert.NoError(t, err, "got bootstrap err") {
+		return
+	}
+	defer gateway.Close()
+
+	server := bazServer.NewSimpleServiceServer(&testCallHandler{})
+	gateway.TChannelBackends()["SimpleService"].Register(server)
```
Raynos: It would be nice if we can register a single method instead of an entire service.

examples/example-gateway/endpoints/baz/baz_simpleservice_method_simple_test.go
```diff
@@ -0,0 +1,83 @@
...
+	headers := map[string]string{}
+
+	res, err := gateway.MakeRequest(
+		"GET",
+		"/baz/simple-path",
+		headers,
+		bytes.NewReader([]byte(`{}`)),
+	)
+
+	if !assert.NoError(t, err, "got http error") {
```
Raynos: duplicate assert here.

ChuntaoLu: good catch, the one below should not be there

examples/example-gateway/clients/baz/baz.go
ChuntaoLu: right, besides the default endpoint timeout, we probably should also have configurable timeouts for different endpoints and clients, thoughts?

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call.go
ChuntaoLu: right, it should be from the annotation

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call_test.go
ChuntaoLu: good point, I am on it

examples/example-gateway/clients/baz/baz.go
Raynos: Agreed.

ChuntaoLu:
@Raynos updated to address the comments in this and previous PRs, ready for another round of review.

examples/example-gateway/clients/baz/baz.go
```diff
@@ -34,23 +31,42 @@ func NewClient(config *zanzibar.StaticConfig, gateway *zanzibar.Gateway) *BazCli
...
+
+	timeout           time.Duration
+	timeoutPerAttempt time.Duration
 }
 
 // Call ...
 func (c *BazClient) Call(ctx context.Context, reqHeaders map[string]string, args *baz.SimpleService_Call_Args) (map[string]string, *baz.BazResponse, error) {
 	var result baz.SimpleService_Call_Result
+
+	retryOpts := &tchannel.RetryOptions{
```
Raynos: We should move this logic into the `Call()` function. 

That way generated code does not import tchannel, only `zanzibar/runtime`.

codegen/templates/endpoint_register.tmpl
```diff
@@ -88,4 +91,36 @@ func Register(g *zanzibar.Gateway, router *zanzibar.Router) {
 		),
 	)
 	{{end}}
+
+	// TODO: (lu) remove below, to be generated
```
Raynos: We cannot put this here as it breaks edge-gateway. We need to use custom endpoints like contacts for this.

codegen/templates/endpoint_register.tmpl
```diff
@@ -12,6 +12,9 @@ import (
 	{{end}}
 
 	"github.com/uber/zanzibar/runtime"
+
+	// TODO: (lu) remove this, to be generated
+	"github.com/uber/zanzibar/examples/example-gateway/endpoints/baz"
```
Raynos: This approach doesn't work with edge-gateway, we actually broke clients.go with edge-gateway.

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call.go
```diff
@@ -0,0 +1,66 @@
+// Code generated by zanzibar
```
Raynos: this is not generated yet.

runtime/router.go
```diff
@@ -163,7 +164,17 @@ func (endpoint *Endpoint) HandleRequest(
...
 
 	fn := endpoint.HandlerFn
-	fn(r.Context(), req, req.res)
+
+	ctx := r.Context()
+	_, ok := ctx.Deadline()
+	if !ok {
+		var cancel context.CancelFunc
+		// TODO: (lu) get timeout from endpoint config
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(100)*time.Millisecond)
```
Raynos: this is not a useful default. This needs to be configurable per endpoint.

Landing this breaks edge-gateway.

test/lib/test_backend/test_tchannel_backend.go
```diff
@@ -47,8 +47,14 @@ func BuildTChannelBackends(
 	n := len(knownTChannelBackends)
 	result := make(map[string]*TestTChannelBackend, n)
 
+	var clientName, serviceName string
 	for i := 0; i < n; i++ {
-		serviceName := knownTChannelBackends[i]
+		clientName = knownTChannelBackends[i]
+
+		// TODO: (lu) tmp, get backend thrift service name from config
+		if clientName == "baz" {
```
Raynos: dont know how we are going to get this information.

This problem might go away if we have the ability to register individual endpoints in the backend.

examples/example-gateway/config/production.json
```diff
@@ -21,6 +21,9 @@
 
 	"clients.baz.port": 4002,
 	"clients.baz.ip": "127.0.0.1",
+	"clients.baz.serviceName": "Qux",
```
Raynos: This service name is a bit weird, can we just use `"baz"`

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/uber-go/zap"
+	"github.com/uber/zanzibar/examples/example-gateway/build/clients"
+	zanzibar "github.com/uber/zanzibar/runtime"
+
+	clientTypeBaz "github.com/uber/zanzibar/examples/example-gateway/build/gen-code/github.com/uber/zanzibar/clients/baz/baz"
+	"github.com/uber/zanzibar/examples/example-gateway/build/gen-code/github.com/uber/zanzibar/endpoints/baz/baz"
+)
+
+// HandleCallRequest handles "/baz/call-path".
+func HandleCallRequest(
```
Raynos: I just landed a change to the semantics of "custom" endpoints. you'll need to rebase on master and change this interface.

Raynos:
Land it yo.

PR #31, opened 2017-03-31
Use best ip on localhost

zhenghuiwang:
use tchannel ListenIP() to find the best IP of localhost to listen.

r: @uber/zanzibar-team


codegen/templates/main.tmpl
```diff
@@ -29,16 +30,22 @@ func getConfigDirName() string {
...
 		filepath.Join(getDirName(), "zanzibar-defaults.json"),
 		filepath.Join(getConfigDirName(), "production.json"),
 		filepath.Join(os.Getenv("CONFIG_DIR"), "production.json"),
-	}, nil)
+	}, seedConfig)
 }
 
 func createGateway() (*zanzibar.Gateway, error) {
-	config := getConfig()
+    listenIP, err := tchannel.ListenIP()
```
Raynos: We want to listen on both IPs.

Listening on just one IP is going to make local development frustrating, as well as authoring tests frustrationg.

codegen/templates/main.tmpl
```diff
@@ -29,18 +30,23 @@ func getConfigDirName() string {
...
 		filepath.Join(os.Getenv("CONFIG_DIR"), "production.json"),
-	}, nil)
+	}, seedConfig)
 }
 
 func createGateway() (*zanzibar.Gateway, error) {
-	config := getConfig()
-	
-	gateway, err := zanzibar.CreateGateway(config, nil)
+    listenIP, err := tchannel.ListenIP()
```
Raynos: Can we move this logic into `gateway.Bootstrap()` instead of being in generated code ?

runtime/gateway.go
```diff
@@ -128,14 +130,25 @@ type RegisterFn func(gateway *Gateway, router *Router)
...
+	_, err := gateway.localServer.JustListen()
 	if err != nil {
 		gateway.Logger.Error("Error listening on port",
 			zap.String("error", err.Error()),
 		)
 		return errors.Wrap(err, "error listening on port")
 	}
-
+	if gateway.localServer.RealIP != gateway.IP {
+		_, err := gateway.server.JustListen()
```
Raynos: Missing a `Close()` call somewhere.

PR #32, opened 2017-04-03
Change benchGateway.CreateGateway to accept endpoint and clients funcs

stevededalus:
Change benchGateway.CreateGateway to accept endpoint and clients funcs. This allows us to decouple the benchGateway used in testing from the example gateway's generated clients and endpoints/register files. Now new gateway repos (ie edge gateway) can use the benchmarkGateway for testing without pulling in the example gateway as a dependency.

### DIFF
--- a/codegen/templates/init_clients.tmpl
+++ b/codegen/templates/init_clients.tmpl
@@ -23,7 +23,7 @@ type Clients struct {
 func CreateClients(
 	config *zanzibar.StaticConfig,
 	gateway *zanzibar.Gateway,
-) *Clients {
+) interface{} {
 	return &Clients{
 		{{range $idx, $cinfo := .ClientInfo -}}
 		{{$cinfo.FieldName}}: {{$cinfo.PackageName}}.NewClient(
--- a/test/lib/bench_gateway/bench_gateway.go
+++ b/test/lib/bench_gateway/bench_gateway.go
@@ -60,21 +60,6 @@ func getZanzibarDirName() string {
 	return filepath.Join(getDirName(), "..", "..", "..")
 }
 
-type clientsInterface interface {
-	CreateClients(
-		config *zanzibar.StaticConfig,
-		gateway *zanzibar.Gateway,
-	) *Clients
-}
-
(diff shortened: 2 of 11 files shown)


codegen/templates/init_clients.tmpl
```diff
@@ -23,7 +23,7 @@ type Clients struct {
 func CreateClients(
 	config *zanzibar.StaticConfig,
 	gateway *zanzibar.Gateway,
-) *Clients {
+) interface{} {
```
Raynos: This would look like it doesn't work, but it does. We might want to refactor this in the future to be a bit more typesafe.

test/lib/bench_gateway/bench_gateway.go
```diff
@@ -62,9 +60,27 @@ func getZanzibarDirName() string {
 	return filepath.Join(getDirName(), "..", "..", "..")
 }
 
+type clientsInterface interface {
```
Raynos: This is unused code.

test/lib/bench_gateway/bench_gateway.go
```diff
@@ -62,9 +60,27 @@ func getZanzibarDirName() string {
...
+	CreateClients(
+		config *zanzibar.StaticConfig,
+		gateway *zanzibar.Gateway,
+	) *Clients
+}
+
+// Clients interface is a placeholder for the generated clients
+type Clients interface {
```
Raynos: Unused code.

test/lib/bench_gateway/bench_gateway.go
```diff
@@ -62,9 +60,27 @@ func getZanzibarDirName() string {
...
+}
+
+type endpointsInterface interface {
```
Raynos: unused code.

codegen/templates/init_clients.tmpl
stevededalus: Agreed, I think we need to change the generated clients struct and then we can have an interface that is generic and shared between cleints and benchmarkgateway, but we can do that in the future.

PR #33, opened 2017-04-03
Improve import statements in generate code.

Raynos:
We have multiple different packages that have the same basename
for example `examples/bar/bar` and `clients/bar/bar`.

This PR adds support for using AliasName for included packages.
For each thrift struct generated by thriftrw we generate
a unique "alias" to refer to it in our generated code.

r: @uber/zanzibar-team

### DIFF
--- a/codegen/package.go
+++ b/codegen/package.go
@@ -162,8 +162,8 @@ func (p PackageHelper) TypePackageName(thrift string) (string, error) {
 	// Strip the leading / and strip the .thrift on the end.
 	thriftSegment := thrift[idx+len(root)+1 : len(thrift)-7]
 
-	thriftPackageName := strings.Replace(thriftSegment, "/", "_", 100)
-	return thriftPackageName, nil
+	thriftPackageName := strings.Replace(thriftSegment, "/", "_", -1)
+	return camelCase(thriftPackageName), nil
 }
 
 func (p PackageHelper) getRelativeFileName(thrift string) (string, error) {
--- a/codegen/package.go
+++ b/codegen/package.go
@@ -159,6 +159,7 @@ func (p PackageHelper) TypePackageName(thrift string) (string, error) {
 		)
 	}
 
+	// Strip the leading / and strip the .thrift on the end.
 	thriftSegment := thrift[idx+len(root)+1 : len(thrift)-7]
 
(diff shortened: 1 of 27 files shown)


codegen/package.go
```diff
@@ -149,8 +149,20 @@ func (p PackageHelper) TypePackageName(thrift string) (string, error) {
...
+
+	idx := strings.Index(thrift, root)
+	if idx == -1 {
+		return "", errors.Errorf(
+			"file %s is not in thrift dir (%s)",
+			thrift, p.gatewayThriftRootDir,
+		)
+	}
+
+	thriftSegment := thrift[idx+len(root)+1 : len(thrift)-7]
```
ChuntaoLu: maybe comment a bit here so 1 and 7 are not mysterious?

stevededalus:
We should avoid using underscores in package names and overrides. We already use camelcase, which isn't great style, but I think it's better than mixing and matching both.
https://blog.golang.org/package-names

Usually underscore is reserved for _test or deployment specific code (i.e. _docker or _unix)

codegen/package.go
```diff
@@ -149,8 +149,21 @@ func (p PackageHelper) TypePackageName(thrift string) (string, error) {
...
+		return "", errors.Errorf(
+			"file %s is not in thrift dir (%s)",
+			thrift, p.gatewayThriftRootDir,
+		)
+	}
+
+	// Strip the leading / and strip the .thrift on the end.
+	thriftSegment := thrift[idx+len(root)+1 : len(thrift)-7]
+
+	thriftPackageName := strings.Replace(thriftSegment, "/", "_", 100)
```
zhenghuiwang: use -1 instead of 100 in `strings.Replace(...)`?

codegen/service.go
```diff
@@ -91,7 +97,7 @@ func (ms *ModuleSpec) AddImports(module *compile.Module, packageHelper *PackageH
 		return errors.Wrapf(err, "can't add import %s", ms.ThriftFile)
 	}
 
-	sort.Strings(ms.IncludedPackages)
+	// sort.Strings(ms.IncludedPackages)
```
zhenghuiwang: delete it?

PR #34, opened 2017-04-03
Cleanup usage of status code thrift annotation

Raynos:
- Use client status code to verify client status code.
 - Only allow one status code per method
 - Use endpoint status code in the response writing.

This change allows us to cleanup the endpoint code to
not need the client status code when the client becomes
a thrift based interface.

The client status code checking logic will move to the
generated client itself.

r: @uber/zanzibar-team

### DIFF
--- a/codegen/method.go
+++ b/codegen/method.go
@@ -47,7 +47,7 @@ type MethodSpec struct {
 	Headers             []string
 	RequestType         string
 	ResponseType        string
-	OKStatusCode        []StatusCode
+	OKStatusCode        StatusCode
 	ExceptionStatusCode []StatusCode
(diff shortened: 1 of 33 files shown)


zhenghuiwang:
Is it possible to have two valid responses(`OKResponse`s) for an endpoint? For example, `checkCredentials` endpoint can return 200 or 404 to indicate the credential is valid or not. They can be both valid responses and we wouldn't fire alerts if 404 is too high.

Raynos:
@zhenghuiwang 

A thrift method can have thrift exceptions. Each thrift exception is "valid" and not to be alerted on.

So once we add thrift exception support we can add a status code (404) annotation to the thrift exception.

PR #37, opened 2017-04-04
Annotate baz client/endpoint thrift

ChuntaoLu:
Very simple PR that annotates the baz client/endpoint thrift

r: @zhenghuiwang

### DIFF
--- a/examples/example-gateway/idl/github.com/uber/zanzibar/clients/baz/baz.thrift
+++ b/examples/example-gateway/idl/github.com/uber/zanzibar/clients/baz/baz.thrift
@@ -22,30 +22,15 @@ service SimpleService {
 
   BazResponse Call(
     1: required BazRequest arg
-  ) (
-    zanzibar.http.status = "200"
-    zanzibar.http.method = "POST"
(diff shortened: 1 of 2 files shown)


examples/example-gateway/idl/github.com/uber/zanzibar/clients/baz/baz.thrift
```diff
@@ -19,9 +19,34 @@ exception NewErr {
 }
 
 service SimpleService {
-  BazResponse Call(1: required BazRequest arg)
-  void Simple() throws (1: SimpleErr simpleErr)
-  void SimpleFuture() throws (1: SimpleErr simpleErr, 2: NewErr newErr)
+
+  BazResponse Call(
```
Raynos: This is a tchannel/thrift client ? Does it need annotations?

ChuntaoLu: right, this does not need annotation, will remove to avoid confusion

ChuntaoLu:
@zhenghuiwang the generation part is not yet there, the annotation is a more like a goal state

PR #39, opened 2017-04-05
Add support for multiple workflows

Raynos:
This refactors the endpoint.tmpl to separate the HTTP server serialization from the "client calling / conversion" logic.

This prepares the http endpoint for tchannel/thrift client support and means the custom endpoints are based on thrift interface instead of raw HTTP interface.

 - refactor template so that Downstream is optional
 - refactor template so that HTTP handler calls a "thrift workflow"
 - refactor dependency passing to create a struct with { clients, logger }

r: @uber/zanzibar-team


examples/example-gateway/endpoints/contacts/save-contacts.go
```diff
@@ -3,58 +3,54 @@ package contacts
...
 
-	body.UserUUID = req.Params[0].Value
-	body.AppType = req.Header.Get("x-uber-client-name")
-	body.DeviceType = req.Header.Get("x-uber-device")
-	body.AppVersion = req.Header.Get("x-uber-client-version")
+	r.UserUUID = endpointContacts.UUID(w.Request.Params[0].Value)
 
-	clientBody := convertToClient(&body)
-	cres, _, err := clients.Contacts.SaveContacts(ctx, nil, clientBody)
+	clientBody := convertToClient(r)
```
ChuntaoLu: nit: `convertToClientRequest` is arguably a better name

examples/example-gateway/endpoints/contacts/save-contacts.go
```diff
@@ -3,58 +3,54 @@ package contacts
...
 	}
 
 	// TODO: verify IsOKResponse() on client response status code
 
 	response := convertToResponse(cres)
-	res.WriteJSON(202, response)
+	return response, nil, nil
 }
 
 func convertToResponse(
```
ChuntaoLu: nit: not a change in this pr, maybe this should have been named `convertToEndpointResponse` as compared to the previous `convertToClient` function

PR #42, opened 2017-04-07
runtime/gateway: read the datacenter file only once

Raynos:
This change reads the datacenterFile once and stores
it in the StaticConfig instance, instead of every
other component re-reading this file.

r: @uber/zanzibar-team @Matt-Esch

### DIFF
--- a/config/production.json
+++ b/config/production.json
@@ -2,9 +2,9 @@
 	"serviceName": "my-gateway",
 	"port": 4000,
 	"env": "production",
+	"datacenterFile": "/etc/datacenter",
 
 	"logger.fileName": "/var/log/my-gateway/my-gateway.log",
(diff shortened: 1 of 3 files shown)


Matt-Esch:
What if this is deployed without a dc file (i.e. not at uber)?

Raynos:
dc defaults to "unknown" if it does not exist.

Matt-Esch:
I was referring to the fact you expect it to exist in the config. Seems a little clunky to hold configuration for features which are unused.

Raynos:
@Matt-Esch It exists in the default config ( https://github.com/uber/zanzibar/pull/42/files#diff-e3e3709caf1b0db2882b08abef48b870 ).

So if an application does not specify a datacenterFile it defaults to "/etc/datacenter". If that file does not exist on disk then "unknown" is used.

PR #43, opened 2017-04-07
Allow non-http client thrift to not have annotations

ChuntaoLu:


codegen/service.go
```diff
@@ -61,7 +61,7 @@ type ServiceSpec struct {
 }
 
 // NewModuleSpec returns a specification for a thrift module
-func NewModuleSpec(thrift string, packageHelper *PackageHelper) (*ModuleSpec, error) {
+func NewModuleSpec(thrift string, wantAnnot bool, packageHelper *PackageHelper) (*ModuleSpec, error) {
```
Raynos: Rather then passing `wantAnnot` along as a parameter ? can we make it a field on `ModuleSpec` and `ServiceSpec` ?

ChuntaoLu: good idea

PR #44, opened 2017-04-08
Add support for thrift exception parsing in client.

Raynos:
This PR adds initial thrift exception support.

 - Read thrift exceptions in the HTTP client.
 - Add support for thrif exceptions in Method class.
 - Add tests for thrift exception usage.

Next up in future PRs:

 - TODO: Support serializing thrift exceptions in server
 - TODO: Generate struct converter functions for client/server
        exceptions.
 - TODO: Ensure http endpoints convert exceptions properly.

r: @uber/zanzibar-team

### DIFF
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -98,7 +98,11 @@ func (c *{{$clientName}}) {{title .Name}}(
 		respHeaders[k] = res.Header.Get(k)
 	}
 
-	res.CheckOKResponse({{.OKStatusCode.Code}})
+	res.CheckOKResponse([]int{
+		{{- range $index, $code := .ValidStatusCodes -}}
+		{{if $index}},{{end}}{{$code}}
+		{{- end -}}
+	})
 
 	{{if eq .ResponseType ""}}
 	_, err = res.ReadAll()
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -106,27 +106,8 @@ func (c *{{$clientName}}) {{title .Name}}(
 	})
 
 	{{if and (eq .ResponseType "") (eq (len .Exceptions) 0)}}
+	// TODO: log about unexpected body bytes?
 	_, err = res.ReadAll()
-	if err != nil {
-		return respHeaders, err
-	}
-	return respHeaders, nil
-	{{else if eq .ResponseType ""}}
-	switch res.StatusCode {
-		{{range $idx, $exception := .Exceptions}}
-		case {{$exception.StatusCode.Code}}:
-			var exception {{$exception.Type}}
-			err = res.ReadAndUnmarshalBody(&exception)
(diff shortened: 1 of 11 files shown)


codegen/templates/http_client.tmpl
```diff
@@ -98,22 +99,79 @@ func (c *{{$clientName}}) {{title .Name}}(
 		respHeaders[k] = res.Header.Get(k)
 	}
 
-	res.CheckOKResponse({{.OKStatusCode.Code}})
+	res.CheckOKResponse([]int{
```
ChuntaoLu: Is this still necessary? It's covered by the default case in the switch check below

Raynos: There are 4 cases, only two of them have a switch.

codegen/test_data/clients/bar.gogen
```diff
@@ -216,14 +303,38 @@ func (c *BarClient) TooManyArgs(
...
+
+	case 403:
+		var exception clientsBarBar.BarException
+		err = res.ReadAndUnmarshalBody(&exception)
+		if err != nil {
+			return nil, respHeaders, err
+		}
+		return nil, respHeaders, &exception
+
+	default:
```
stevededalus: how about removing this from default and then you don't need to worry about the dead code check below? Then anyone adding a status code but forgetting to return will see "Unexpected http client response" errors when testing/building.

codegen/templates/http_client.tmpl
```diff
@@ -98,22 +99,79 @@ func (c *{{$clientName}}) {{title .Name}}(
...
+		default:
+			// TODO: log about unexpected body bytes?
+			_, err = res.ReadAll()
+	}
+
+	if err != nil {
+		return respHeaders, err
+	}
+	return respHeaders, nil
+	{{else if eq (len .Exceptions) 0}}
```
stevededalus: Might be slightly more readable if this was the second case in your if-else-if-else chain. That way the two switch statements are together and the two "base" cases are together.

codegen/test_data/clients/bar.gogen
Raynos: Good suggestion.

codegen/templates/http_client.tmpl
ChuntaoLu: I see, maybe it can be removed for the cases that have switch.

Raynos: This method still logs a warning to disk where as the default block only returns an error.

I want to keep the warning logic here.

zhenghuiwang: It is simpler to change `CheckOKResponse` to take `.ValidStatusCodes` as input than serialization here.

Raynos: There is no reference to .ValidStatusCodes anywhere. it only exists on EndpointSpec which is not loaded in the gateway binary only in the code generation runner.

PR #45, opened 2017-04-10
Remove the requirement that a thrift file must be under a gateway IDL directory.

zhenghuiwang:
This PR makes the generated Go File path determined only by its thrift file path.

For example, after this PR, 
`<idl root>/<path to a thrift>/foo.thrift` generates client or endpoint Go code to `build/<path to a thrift>/foo/foo.go`

Previously, all clients code are generated from `<idl root>/<path_to_a_gateway>/<sub path>` to `build/clients/<sub path>`, which unnecessarily requires the thrift files to be under a gateway IDL dir.

r: @uber/zanzibar-team


codegen/gateway.go
```diff
@@ -427,18 +427,20 @@ func NewEndpointSpec(
...
-		dirName,
-	)
 
-	goStructsFileName := filepath.Join(
-		h.CodeGenTargetPath(),
-		"endpoints",
-		dirName,
-		dirName+"_structs.go",
-	)
+	goFolderName, err := h.GoFileGenPath(thriftFile, "")
```
Raynos: Why ? We don't want this change. The file name should be based on `client-config.json` and `endpoints-config.json` , not the thrift file.

codegen/gateway.go
```diff
@@ -210,25 +210,26 @@ func NewHTTPClientSpec(jsonFile string, clientConfigObj map[string]string, h *Pa
 
 	baseName := filepath.Base(filepath.Dir(jsonFile))
 
-	goFileName := filepath.Join(
-		h.CodeGenTargetPath(),
-		"clients",
-		baseName,
-		baseName+".go",
-	)
+	goFileName, err := h.GoFileGenPath(thriftFile, baseName+".go")
```
Raynos: This creates a deeply nested folder. we want to avoid that, the file name should be based on the `client-config.json` not the thrift file.

PR #46, opened 2017-04-10
Remove checking a thrift file must be under a gateway IDL directory

zhenghuiwang:
Previously, all clients code are generated from thrift files that must be under a gateway IDL dir. This PR removes the check.

1. Remove `gatewayThriftRoot` dir since it is no longer important. use `gatewayNamespace` for Go package path of the gateway.

2. Move googleNow client&endpoint out of the `idl/<gatewayNamespace>` folder to show the generation still works.

r: @uber/zanzibar-team


Raynos:
The generated names are getting pretty crazy `"githubComUberZanzibarEndpointsBarBar"`

To clean up our example code can we move all thrift files to:

 - `idl/endpoints/*.thrift`
 - `idl/clients/*.thrift`

This will flatten out a lot of the generated code and make things cleaner.

PR #48, opened 2017-04-10
Add zap logger to tchannel logger adapter

ChuntaoLu:


runtime/tchannel_logger.go
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+
+	"github.com/uber/tchannel-go"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// NewTChannelLogger creates a TChannel logger given a zap logger
+func NewTChannelLogger(logger *zap.Logger) tchannel.Logger {
+	return &TChanLogger{logger}
```
Raynos: Can we return struct by value ? `TChanLogger{logger: logger}`

Also nitpick: always use {logger: logger} for no ambiguity

runtime/tchannel_logger.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Debug logs a message at debug priority.
+func (l TChanLogger) Debug(msg string) {
+	l.logger.Debug(msg)
+}
+
+// Fields returns the fields that this logger contains.
+func (l TChanLogger) Fields() tchannel.LogFields {
+	// TODO: (lu) is it ok to return nil?
```
Raynos: I can't find any callers to Fields() in tchannel-go other then unit tests.

ChuntaoLu: Nice

runtime/tchannel_logger.go
ChuntaoLu: will do

PR #49, opened 2017-04-12
Generate clients with module system

Matt-Esch:
Next step for integrating the module system is to generate the client files. Note that the old method for generating the clients index still exists and will be generated by the module system later.


module/module.go
```diff
@@ -515,3 +517,41 @@ func (jsonConfig *JSONClassConfig) Read(
 
 	return configFile, nil
 }
+
+func writeFile(filePath string, bytes []byte) error {
```
Raynos: Consider comment // like ioutil.WriteFile but also mkdirp()

PR #50, opened 2017-04-13
Generate HTTP endpoints for tchannel client work flow

ChuntaoLu:
This PR adds support for HTTP endpoints generation with tchannel client.


codegen/method.go
```diff
@@ -191,7 +199,22 @@ func isStructType(spec compile.TypeSpec) bool {
...
+	var requestType string
+	// TODO: (lu) the assumption here is 'no annotation == tchannel', good enough until new protocol is introduced
+	if ms.WantAnnot {
+		requestType = strings.Title(f.Name) + "HTTPRequest"
+	} else {
+		pkgName, err := h.TypePackageName(curThriftFile)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to generate new request type")
+		}
+		requestType = fmt.Sprintf(
```
Raynos: Can we leave a comment here.

This is specifically generating the "Args" type that thriftrw generates.

ChuntaoLu: Done.

PR #51, opened 2017-04-13
Fix header propogation from endpoint to client

stevededalus:
Modify WriteJSON and WriteJSONBytes to accept header maps. Modify the template map to propagate all required headers to the client and all required headers from the client response to the the server response. We can not blindly copy all headers, because we don't want the content-length, content-type, etc from the client impacting the server.

Future work and considerations:
Currently, we treat headers as map[string]string although in go's http they are map[string][]string. We should decide whether we want to support duplicate headers in zanzibar or not, they seem to create a lot of issues with different libraries so I am leaning against that option.

We define the required headers as annotations on the thrift idl method definitions. We should determine if we want to support different sets of headers for request and response. We should also add optional headers as an annotation, currently all headers are required.


codegen/templates/endpoint.tmpl
```diff
@@ -9,6 +9,8 @@ import (
 	"io/ioutil"
 	"net/http"
```
zhenghuiwang: extra blank line here?

codegen/test_data/endpoints/bar_bar_method_argnotstruct.gogen
```diff
@@ -26,14 +26,18 @@ func HandleArgNotStructRequest(
 	}
 
 	headers := map[string]string{}
+	// TODO(sindelar): Add optional headers in addition to required.
+	for _, k := range []string(nil) {
```
zhenghuiwang: It is strange to have generated code ranging over nil. Can we make template copy headers only if there are headers in request?

runtime/gateway.go
```diff
@@ -31,13 +31,14 @@ import (
 	"sync"
 	"time"
```
zhenghuiwang: extra blank line?

codegen/templates/endpoint.tmpl
```diff
@@ -50,20 +52,25 @@ func Handle{{title .Name}}Request(
 	{{end}}
 
 	headers := map[string]string{}
+	// TODO(sindelar): Add optional headers in addition to required.
+	for _, k := range {{.Headers | printf "%#v" }} {
```
Raynos: This is not what .Headers means. We already have `req.CheckHeaders()`

We want to get all headers here, all of them.

codegen/templates/endpoint.tmpl
```diff
@@ -73,10 +80,16 @@ func Handle{{title .Name}}Request(
 		return
 	}
 
+	// TODO(sindelar): Add response headers as an thrift spec annotation.
+	endRespHead := map[string]string{}
+	for _, k := range {{.Headers | printf "%#v" }} {
```
Raynos: There is no header white list here, if the workflow returns headers we should write all headers to the response.

codegen/templates/endpoint.tmpl
```diff
@@ -125,18 +138,18 @@ func (w {{$workflow}}) Handle(
 	{{end}}
 
 	{{if and (eq $clientReqType "") (eq $clientResType "")}}
-	_, err := w.Clients.{{$clientName}}.{{$clientMethodName}}(ctx, nil)
+	respHeaders, err := w.Clients.{{$clientName}}.{{$clientMethodName}}(ctx, headers)
```
Raynos: We do not want to proxy headers here.

We need a new whitelist in the endpoint json config for which headers to propagate. This is different from `.Headers` on the `MethodSpec`.

codegen/templates/endpoint.tmpl
```diff
@@ -147,10 +160,10 @@ func (w {{$workflow}}) Handle(
 	}
 
 	{{if eq .ResponseType "" -}}
-	return nil, nil
+	return respHeaders, nil
```
Raynos: We do not want to proxy all response headers. we need a response header whitelist.

The request header proxy whitelist & response header proxy whitelist should be two seperate configs in the endpoint json config.

runtime/server_http_response.go
```diff
@@ -110,26 +110,30 @@ func (res *ServerHTTPResponse) SendErrorString(
...
 // WriteJSONBytes writes a byte[] slice that is valid json to Response
 func (res *ServerHTTPResponse) WriteJSONBytes(
-	statusCode int, bytes []byte,
+	statusCode int, headers map[string]string, bytes []byte,
 ) {
 	// TODO: mark header as pending ?
 	res.responseWriter.Header().
 		Set("content-type", "application/json")
 
+	for k, v := range headers {
```
Raynos: Let's set headers first and then finally set the content-type.

We do not want application headers to overwrite the content-type.

codegen/templates/endpoint.tmpl
stevededalus: I don't think we want all of them. Content-length definitely breaks downstream requests, could potentially content-type.

Raynos: We want to parse all the headers here.

We do not want to *propagate* all the headers.

stevededalus:
PTAL

codegen/gateway.go
```diff
@@ -484,6 +489,31 @@ func NewEndpointSpec(
 		}
 	}
 
+	reqHeaderMap := make(map[string]string)
+	m, ok := endpointConfigObj["reqHeaderMap"]
+	if ok {
```
Raynos: else error ?

codegen/gateway.go
```diff
@@ -484,6 +489,31 @@ func NewEndpointSpec(
...
+		// Do a deep cast to enforce a string -> string map
+		castMap := m.(map[string]interface{})
+		for key, value := range castMap {
+			switch value := value.(type) {
+			case string:
```
Raynos: default error ?

codegen/service_test.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/uber/zanzibar/codegen"
 )
 
-var updateGoldenFile = flag.Bool("update", false, "Updates the golden files with expected response.")
+var updateGoldenFile = flag.Bool("update", true, "Updates the golden files with expected response.")
```
Raynos: You can do `go test ./codegen/ -update` ;)

We want this flag to stay defaulted to false.

codegen/templates/endpoint.tmpl
```diff
@@ -103,41 +104,75 @@ type {{$workflow}} struct {
 func (w {{$workflow}}) Handle(
 {{- if and (eq .RequestType "") (eq .ResponseType "") }}
 	ctx context.Context,
-	headers map[string]string,
+	// TODO(sindelar): Switch to zanzibar.Headers when tchannel 
+ 	// generation is implemented.
+	headers http.Header,
```
Raynos: We want `map[string]string` here. Let's not use `http.Header`

codegen/templates/endpoint.tmpl
```diff
@@ -50,21 +52,20 @@ func Handle{{title .Name}}Request(
 	}
 	{{end}}
 
-	headers := map[string]string{}
```
Raynos: We need to parse http headers into a map.

codegen/templates/endpoint.tmpl
```diff
@@ -75,9 +76,9 @@ func Handle{{title .Name}}Request(
 	}
```
Raynos: No validation of `RespHeaders` ?

codegen/templates/endpoint.tmpl
```diff
@@ -146,11 +181,19 @@ func (w {{$workflow}}) Handle(
 		return {{if eq .ResponseType ""}}nil, err{{else}}nil, nil, err{{end}}
 	}
 
+	// Filter and map response headers from client to server response.
+	endRespHead := map[string]string{}
+	{{if (ne (len $resHeaderMap) 0) -}}
+	for k, v := range {{$resHeaderMap | printf "%#v" }} {
+		endRespHead[v] = respHeaders[k]
```
Raynos: This semantics is confusing.

In the reqHeaderMap key is SERVER header and value is CLIENT header.

Here the key is CLIENT header and the value is SERVER header.

Should we do `endRespHead[k] = respHeaders[v]` ?

codegen/templates/endpoint.tmpl
```diff
@@ -146,11 +181,19 @@ func (w {{$workflow}}) Handle(
 		return {{if eq .ResponseType ""}}nil, err{{else}}nil, nil, err{{end}}
```
Raynos: Want to return partial headers here in error case if we have headers. 

Add a TODO for now ?

codegen/test_data/endpoint_tests/bar_bar_method_toomanyargs_test.gogen
```diff
@@ -41,8 +41,6 @@ func TestTooManyArgsSuccessfulRequestOKResponse(t *testing.T) {
 	)
 
 	headers := map[string]string{}
-	headers["x-uuid"] = "placeholder"
```
Raynos: Can we update thrift files to use new annotations ?

codegen/test_data/endpoints/bar_bar_method_toomanyargs.gogen
```diff
@@ -23,23 +24,18 @@ func HandleTooManyArgsRequest(
 	res *zanzibar.ServerHTTPResponse,
 	clients *clients.Clients,
 ) {
-	if !req.CheckHeaders([]string{"x-uuid", "x-token"}) {
```
Raynos: Can we update thrift annotation ?

examples/example-gateway/idl/endpoints/googlenow/googlenow.thrift
```diff
@@ -16,6 +16,7 @@ service GoogleNow {
         zanzibar.http.path = "/googlenow/check-credentials"
         zanzibar.http.status = "202"
         // comma sparated list for required headers
-        zanzibar.http.headers = "x-uuid,x-token"
+        zanzibar.http.reqHeaders = "x-uuid,x-token"
```
Raynos: missed old annotation on `addCredentials`.

stevededalus: I don't understand this comment.

stevededalus: ah, nm, it was meant for a higher up line :)

codegen/gateway.go
stevededalus: Added, but this means these fields are required in the endpoint config. Not sure if that will complicate tchannel endpoints when Matt implements them.

codegen/gateway.go
stevededalus: Done

codegen/service_test.go
stevededalus: Whoops, done

codegen/templates/endpoint.tmpl
stevededalus: What's the advantage to doing that here?

That way we're creating three header objects and parsing/filtering them twice. The headers are inspected and mapped/filtered in the workflow. Seems like we should just pass the http.Header to the workflow and when we want to support non-http endpoints we can create a zanzibar.Header and use that throughout serving.

codegen/test_data/endpoints/bar_bar_method_toomanyargs.gogen
stevededalus: Done.

codegen/test_data/endpoint_tests/bar_bar_method_toomanyargs_test.gogen
stevededalus: Done.

codegen/templates/endpoint.tmpl
stevededalus: I'm not sure that's less confusing. The map is taking header X and putting it in header Y. On the request path that server->client, on the response that's client->server.

codegen/templates/endpoint.tmpl
stevededalus: TODO added

codegen/gateway.go
Raynos: Good point, this field should be optional, my bad.

codegen/templates/endpoint.tmpl
Raynos: A workflow is transport agnostic. we dont want to regress back to having http specific workflows

It's interface should not include `http.Header`, this means it should have either `zanzibar.Header` or `map[string]string`

codegen/templates/endpoint.tmpl
stevededalus: added a todo for now.

codegen/templates/endpoint.tmpl
```diff
@@ -75,6 +75,10 @@ func Handle{{title .Name}}Request(
 		return
 	}
 
+	{{- if .ResHeaders -}}
```
Raynos: nitpick, we want a new line, can we do `}}` instead of `-}}`

codegen/templates/endpoint.tmpl
stevededalus: I think moving to map[string]string is going to hurt things in the long run because we lose canonicalization of headers, so the request response can become unmatched based on capitalization

stevededalus: I'll add a zanzibar.Header to runtime

examples/example-gateway/endpoints/googlenow/check_credentials_test.json
```diff
@@ -3,12 +3,28 @@
...
         "clientRequest": {"authcode": "test"},
-        "clientResponse": {"status": "200 OK"}
+        "clientReqHeaders":  {
+            "X-Uuid": "test-uuid", 
+            "X-Token": "test-token"
+        },
+        "clientResponse": {"status": "200 OK"},
+        "clientResHeaders":  {
+            "X-Uuid": "test-uuid", 
+            "X-Token": "test-token"
```
Raynos: We do not actually want to proxy tokens around for googlenow. Just the uuid.

runtime/server_header.go
```diff
@@ -0,0 +1,59 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zanzibar
+
+// ServerHeader struct manages request
+type ServerHeader map[string][]string
```
Raynos: Can we make `ServerHeader` an interface ? That way it can be backed by both `http.Header` and `map[string]string` with two implementations, one for http and one for tchannel.

Raynos:
Server header interface is not being used in server_http_request.go yet.

runtime/server_header.go
```diff
@@ -0,0 +1,73 @@
...
+package zanzibar
+
+import (
+	"net/http"
+)
+
+// ServerHeaderInterface defines methods on ServerHeaders
+type ServerHeaderInterface interface {
+	Get(key string) (string, bool)
+	Add(key string, value string)
```
Raynos: Is there a semantic difference between `Add()` and `Set()`

stevededalus: Add appends a header, set overwrites.

Raynos: TChannel does not allow duplicate headers.

If we want transport agnostic headers that support both TChannel and HTTP then what is the difference between `Add()` and `Set()` for tchannel ?

codegen/templates/endpoint.tmpl
```diff
@@ -126,35 +127,46 @@ func (w {{$workflow}}) Handle(
 	ctx context.Context,
 	// TODO(sindelar): Switch to zanzibar.Headers when tchannel 
  	// generation is implemented.
-	headers http.Header,
+	headers zanzibar.ServerHeaderInterface,
```
Raynos: nitpick: ending something in `Interface` seems not idiomatic.

codegen/templates/endpoint.tmpl
```diff
@@ -214,18 +202,18 @@ func (w {{$workflow}}) Handle(
 	}
 
 	// Filter and map response headers from client to server response.
-	endRespHead := map[string]string{}
-	{{if (ne (len $resHeaderMap) 0) -}}
+
+	// TODO: Add support for TChannel Headers with a switch here
+	resHeaders := zanzibar.ServerHTTPHeader{}
```
Raynos: Ideally a workflow should be transport agnostic without a switch.

Maybe we need `zanzibar.ServerHeader{}` type ?

PR #52, opened 2017-04-17
Refactor endpoints to have an init pattern

Raynos:
Currently our clients module all have a stable pattern
of each client being a struct with a NewX() function.

However our endpoint handlers are just flat functions.

This PR refactors each endpoint to be a struct with
a method and a constructor.

This means its dependencies ( in this case clients ) are
passed into the constructor instead of passed as an extra
argument in the handler method.

I've also updated endpoints_register.tmpl to set up
an Endpoints struct and a NewEndpoints to bootstrap all
of the endpoints. 

In the feature both the NewEndpoints and NewClients file
will be generate from the module system.

r: @uber/zanzibar-team @Matt-Esch

### DIFF
--- a/codegen/templates/endpoint_register.tmpl
+++ b/codegen/templates/endpoint_register.tmpl
@@ -14,63 +14,39 @@ import (
 	"github.com/uber/zanzibar/runtime"
 )
 
-type handlerFn func(
-	ctx context.Context,
-	req *zanzibar.ServerHTTPRequest,
-	res *zanzibar.ServerHTTPResponse,
-	clients *clients.Clients,
-)
-
-type myEndpoint struct {
-	HandlerFn handlerFn
-	Clients   *clients.Clients
+// Endpoints is a struct that holds all the endpoints
+type Endpoints struct {
+	{{range $idx, $endpoint := .Endpoints -}}
+	{{$endpoint.HandlerName}} *{{$endpoint.HandlerType}}
+	{{end}}
 }
 
-func (endpoint *myEndpoint) handle(
-	ctx context.Context,
-	req *zanzibar.ServerHTTPRequest,
-	res *zanzibar.ServerHTTPResponse,
-) {
-	fn := endpoint.HandlerFn
-	fn(ctx, req, res, endpoint.Clients)
-}
-
-func makeEndpoint(
-	g *zanzibar.Gateway,
-	endpointName string,
-	handlerName string,
-	middlewares []zanzibar.MiddlewareHandle,
-	handlerFn handlerFn,
-) *zanzibar.Endpoint {
-	myEndpoint := &myEndpoint{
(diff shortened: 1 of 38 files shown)


module/module.go
```diff
@@ -305,12 +305,17 @@ func (moduleSystem *System) GenerateBuild(
 				targetGenDir,
 				classInstance.Directory,
 			)
+			prettyBuildPath := filepath.Join(
```
Matt-Esch: You want path.Rel probably. Can't assume that Base is one level relative to the cwd.

codegen/templates/endpoint_register.tmpl
```diff
@@ -14,63 +14,39 @@ import (
...
-		makeEndpoint(
+		zanzibar.NewEndpoint(
 			g,
 			"{{$endpoint.EndpointID}}",
 			"{{$endpoint.HandlerID}}",
 			{{ if len $endpoint.Middlewares | ne 0 -}}
-			
-			[]zanzibar.MiddlewareHandle{
-			{{range $idx, $middleware := $endpoint.Middlewares -}}
+			zanzibar.NewStack([]zanzibar.MiddlewareHandle{
```
Matt-Esch: This is the first time I have seen this. I don't like that we have detached the middlewares from the endpoint itself. The router should not be endpoint aware and this magic here which is in a global registration code should just be loading blindly into the router.

examples/example-gateway/build/endpoints/register.go
```diff
@@ -17,89 +14,89 @@ import (
...
+// CreateEndpoints bootstraps the endpoints.
+func CreateEndpoints(
+	gateway *zanzibar.Gateway,
+) interface{} {
+	return &Endpoints{
+		ArgNotStructHandler: bar.
+			NewArgNotStructEndpoint(gateway),
+		MissingArgHandler: bar.
+			NewMissingArgEndpoint(gateway),
+		NoRequestHandler: bar.
```
Matt-Esch: avoid breaking lines on ".", consider before "bar"

Matt-Esch:
👍 generally LGTM, I've left a couple of auxiliary comments

module/module.go
Raynos: path.Rel will give example/example-gateway/build which is too much.

Just doing build/... is good enough.

codegen/templates/endpoint_register.tmpl
Raynos: This can be refactored in a seperate diff.

Currently the router is not endpoint or middleware aware, 

We just wrap all the handler functions in a single file instead of one file per endpoint.

examples/example-gateway/build/endpoints/register.go
Raynos: We cant line break before "bar", gofmt will put it back on one line. 

I'll just keep it all on one line for now.

PR #54, opened 2017-04-18
Improve client status code handling

Raynos:
This PR makes the status code validation in the client
mandatory even if there are no exceptions of response body.

This means that if the client returns an unexpected status
code the client will return an error and the endpoint handler
will write a 500 to the caller of edge-gateway

Also cleaned up some things :

 - templates/structs: add json annotations
 - Fix some tests
 - Make testing cache optional

r: @uber/zanzibar-team

### DIFF
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -105,20 +105,26 @@ func (c *{{$clientName}}) {{title .Name}}(
 	})
 
 	{{if and (eq .ResponseType "") (eq (len .Exceptions) 0)}}
-	// TODO: log about unexpected body bytes?
-	_, err = res.ReadAll()
-	if err != nil {
-		return respHeaders, err
+	switch res.StatusCode {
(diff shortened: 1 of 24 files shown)


codegen/templates/http_client.tmpl
```diff
@@ -105,20 +105,26 @@ func (c *{{$clientName}}) {{title .Name}}(
...
 
 	{{if and (eq .ResponseType "") (eq (len .Exceptions) 0)}}
-	// TODO: log about unexpected body bytes?
-	_, err = res.ReadAll()
-	if err != nil {
-		return respHeaders, err
+	switch res.StatusCode {
+		case {{.OKStatusCode.Code}}:
+			// TODO: log about unexpected body bytes?
+			_, err = res.ReadAll()
```
ChuntaoLu: Shouldn't it be considered as an error because response body is expected to contain no bytes?

Raynos: There is an I implemented todo :)

PR #56, opened 2017-04-18
Generate tchannel server

ChuntaoLu:
This PR is on top of https://github.com/uber/zanzibar/pull/55

PR #57, opened 2017-04-19
Refactor baz.thrift and TChannel runtime

ChuntaoLu:
This PR refactors baz.thrift to cover all the req/resp combinations, and moves the thrift method dispatch logic from gencode to runtime.

This PR is stacked on top of https://github.com/uber/zanzibar/pull/56


examples/example-gateway/endpoints/baz/silly_noop.json
```diff
@@ -0,0 +1,14 @@
+{
+	"endpointType": "http",
+	"endpointId": "baz",
+	"handleId": "sillyNoop",
+	"thriftFile": "endpoints/baz/baz.thrift",
+	"thriftFileSha": "{{placeholder}}",
+	"thriftMethodName": "SimpleService::SillyNoop",
+    "workflowType": "tchannelClient",
```
Raynos: you keep mixing spaces vs tabs. 

Can you evaluated [eclint](https://github.com/jedmao/eclint) and add it to zanzibar for .json and .tmpl files ?

ChuntaoLu: fixed, I will look into eclint

runtime/tchannel_server.go
```diff
@@ -42,11 +42,11 @@ import (
 type PostResponseCB func(ctx context.Context, method string, response RWTStruct)
 
 type handler struct {
-	server         TChanServer
+	handler        TChanHandler
```
Raynos: handler.handler is a bit weird.

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call_test.go
```diff
@@ -63,7 +63,7 @@ func TestCallSuccessfulRequestOKResponse(t *testing.T) {
 	}
 	defer gateway.Close()
 
-	gateway.TChannelBackends()["baz"].Register(bazServer.NewServerWithSimpleServiceCall(call))
+	gateway.TChannelBackends()["baz"].Register(bazServer.NewSimpleServiceCallHandler(call))
```
Raynos: Somehow this feels wrong.

`Register({ Service, Method, RequestHandler })` feels wrong compared to `Register("Service::Method", RequestHandler)`

It's weird that we register an interface that contains methods to get the arg1 name.

Definitely not symmetrical with registering http endpoints

examples/example-gateway/endpoints/baz/baz_simpleservice_method_call_test.go
```diff
@@ -25,7 +25,7 @@ package baz
 import (
```
ChuntaoLu: will do

runtime/tchannel_server.go
ChuntaoLu: will rename

PR #58, opened 2017-04-20
Add support for thrift exceptions in http endpoints

Raynos:
This PR adds support for thrift exceptions in http endpoints.
We now serialize thrift exception structs to JSON with proper
status codes.

I've also added support for generating conversion functions in
the generate workflow that converts client exceptions to server
exceptions.

The conversion code has a TODO where it creates the "default"
struct for the server exception and does not copy over the fields.

r: @uber/zanzibar-team


ChuntaoLu:
LGTM, seems a rebase is needed.

codegen/method.go
```diff
@@ -304,12 +309,14 @@ func (ms *MethodSpec) setExceptions(
 		}
 
 		if !ms.WantAnnot {
```
stevededalus: Maybe change WantAnnot to RequireAnnot

codegen/method.go
```diff
@@ -291,6 +293,9 @@ func (ms *MethodSpec) setExceptions(
 		ms.OKStatusCode.Code: true,
 	}
 	ms.Exceptions = make([]ExceptionSpec, len(resultSpec.Exceptions))
+	ms.ExceptionsIndex = make(
```
stevededalus: May Map or Dict instead of index

codegen/method.go
```diff
@@ -342,6 +349,8 @@ func (ms *MethodSpec) setExceptions(
 				Message: e.Name,
 			},
 		}
+		ms.Exceptions[i] = exception
```
stevededalus: You might need to sort Exceptions at the end of constructing it to ensure that the ordering in the templates is fixed. I'm not sure what ordering resultSpec.Exceptions will be in.

Raynos: Seems to not be an issue in practice.

PR #60, opened 2017-04-20
Serve static template files in .go source code

zhenghuiwang:
Problem: For an application using Zanzibar in docker, the vendor folder is used to build binary but not uploaded along with the binary. Zanzibar can't find its template files during runtime.

This PR uses `go-bindata` package (https://github.com/jteeuwen/go-bindata) to compile the templates file into a `.go` source code. Zanzibar will always be able to find templates regardless how static files are served.

r: @uber/zanzibar-team


Makefile
```diff
@@ -87,7 +92,7 @@ install-wrk:
 	cd ./vendor/wrk ; (ls ./wrk 2>/dev/null || make >install_wrk.log)
 
 .PHONY: test
-test: generate lint
+test: template-data generate lint
```
Raynos: Can we run `template-data` as part of `make generate` ?

tmpldata/template_files.go
```diff
@@ -0,0 +1,442 @@
+// Code generated by go-bindata.
```
Raynos: Can we keep this file in `codegen/templates/template_files.go` ?

PR #61, opened 2017-04-21
Rename serverheaderinterface to serverheader

stevededalus:
Drop the "interface" from interface name to be more idiomatic.


runtime/server_header.go
```diff
@@ -24,8 +24,8 @@ import (
 	"net/http"
 )
 
-// ServerHeaderInterface defines methods on ServerHeaders
-type ServerHeaderInterface interface {
+// ServerHeader defines methods on ServerHeaders
+type ServerHeader interface {
```
Raynos: Consider calling it just `Header` :)

stevededalus: Done

PR #62, opened 2017-04-21
Remove cycle with template_files in template dir

stevededalus:
Move template_files.go into its own directory so that it is not bundling an old version of itself when regenerating.


Makefile
```diff
@@ -21,12 +21,12 @@ PROGS = examples/example-gateway/build/example-gateway \
 .PHONY: check-licence
 check-licence:
 	ls ./node_modules/.bin/uber-licence 2>/dev/null || npm i uber-licence
-	./node_modules/.bin/uber-licence --dry --file '*.go' --dir '!vendor' --dir '!examples' --dir '!.tmp_gen' --dir '!templates'
+	./node_modules/.bin/uber-licence --dry --file '*.go' --dir '!vendor' --dir '!examples' --dir '!.tmp_gen' --dir '!templates' --dir '!template_bundle'
```
zhenghuiwang: we can remove `--dir "!templates"`

Makefile
```diff
@@ -21,12 +21,12 @@ PROGS = examples/example-gateway/build/example-gateway \
...
 
 .PHONY: fix-licence
 fix-licence:
 	ls ./node_modules/.bin/uber-licence 2>/dev/null || npm i uber-licence
-	./node_modules/.bin/uber-licence --file '*.go' --dir '!vendor' --dir '!examples' --dir '!.tmp_gen' --dir '!templates'
+	./node_modules/.bin/uber-licence --file '*.go' --dir '!vendor' --dir '!examples' --dir '!.tmp_gen' --dir '!templates' --dir '!template_bundle'
```
zhenghuiwang: we can remove --dir \`!templates\`

PR #66, opened 2017-04-24
document semantics of HTTP+JSON

Raynos:
This adds a document to specify what the semantics of 
our annotations are and how the HTTP+JSON support work.

r: @uber/zanzibar-team

### DIFF
--- a/docs/thrift.md
+++ b/docs/thrift.md
@@ -10,6 +10,8 @@ The structs defined in the thrift file are serialized into
 JSON. When parsing, any fields on the wire not defined
 in the thrift struct are ignored.
 
+## JSON semantics
+
 ## Annotations
 
 ### `zanzibar.http.method`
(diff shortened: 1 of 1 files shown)


docs/thrift.md
```diff
@@ -0,0 +1,84 @@
...
+
+### `zanzibar.http.method`
+
+required.
+
+The HTTP method to use, this is mandatory and can be
+"GET", "POST", "DELETE", "UPDATE", "PATCH"
+
+If the method is GET then the function arguments
+must have http.ref annotations remapping arguments
```
zhenghuiwang: change  `http.ref` to` "zanzibar.http.ref"` to be more clear?

docs/thrift.md
```diff
@@ -0,0 +1,84 @@
...
+to query parameters or params.
```
zhenghuiwang: change this line to `to query parameters, params, or headers.`?

docs/thrift.md
ChuntaoLu: probably even better with relative link, `[zanzibar.http.ref](#zanzibar.http.ref)`

docs/thrift.md
ChuntaoLu: This line might not be necessary as the customization is specified in zanzibar.http.ref section.

docs/thrift.md
```diff
@@ -0,0 +1,84 @@
...
+optional. Annotation on thrift method
+
+`headerGroups` is a comma seperated list of struct names
+which define mandatory headers to apply to this method.
+
+### `zanzibar.http.reqHeaders`
+
+optional. Annotation on thrift method
+
+The list of required headers on the http request.
```
ChuntaoLu: maybe mention the headers that are optional?

Raynos:
Tracking issue for things we need to implement until this document is accurate ( https://github.com/uber/zanzibar/issues/68 ).

docs/thrift.md
Raynos: optional headers do not have annotations. Dont know what the best place is to document optional headers.

Maybe document it once statically for ALL of the gateway instead of "once per thrift method"

ChuntaoLu: sounds good

Matt-Esch:
- We need a way to specify "Number or String" JSON representation
 - We need a way to tell specify when "null" satisfies "optional" and when it does not.

Raynos:
null and a missing field always satisfy optional as far as I can tell.

Matt-Esch:
Are there any annotations for mapping JSON names to struct names? Like maybe the struct uses pascal case but the JSON object has a lower snake case name? Given the public/private semantics I'd imagine that what we'd really want is pascal case names for all struct fields with annotations for the required JSON (de)serialized names.

Raynos:
Each go struct has explicit json annotations for what the json field name is based on the thrift field name.

All go strucst have pascal names for everything but thats just a go thing.

Raynos:
For snake casing there is `zanzibar.http.ref = "body.snake_case"` . the http.ref = body.x allows full customization of json field name on the wire irregardless of thrift field name.

PR #67, opened 2017-04-24
Pin atomic in glide.yaml

stevededalus:
Pin atomic in glide.yaml


glide.yaml
```diff
@@ -15,7 +15,7 @@ import:
 - package: github.com/pkg/errors
   version: ^0.8.0
 - package: github.com/uber-go/atomic
```
Raynos: There are two of them. the other one is `go.uber.org/atomic` ...

PR #71, opened 2017-04-25
Makefile: go-bindata set modtime to 1

Raynos:
This avoids flaps in `go-bindata`. The mod time of our templates
is based on the time you call `git pull`.

r: @uber/zanzibar-team


ChuntaoLu:
That's interesting, why is the mod time based on `git pull`?

PR #72, opened 2017-04-26
Add a GetOrEmpty convenience method to zanzibar.Headers

stevededalus:
Add a GetOrEmpty convenience method to zanzibar.Headers that does not return an error for missing keys. Allows logging and middlewares to inline assignments without if statements for every field.

### DIFF
--- a/runtime/server_header.go
+++ b/runtime/server_header.go
@@ -53,6 +53,18 @@ func (zh ServerHTTPHeader) Get(
 	return "", false
 }
 
+// GetOrEmptyStr retrieves the first string stored on a given header or
+// the empty string (golang's zero vlaue for string types)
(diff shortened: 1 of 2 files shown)


runtime/server_header.go
```diff
@@ -53,6 +53,18 @@ func (zh ServerHTTPHeader) Get(
 	return "", false
 }
 
+// GetOrEmptyStr retrieves the first string stored on a given header or
+// the empty string (golang's zero vlaue for string types)
+func (zh ServerHTTPHeader) GetOrEmptyStr(
+	key string,
```
zhenghuiwang: can the function signature be put in one line?

stevededalus: Done.

PR #74, opened 2017-04-26
Add a request logger to zanzibar

stevededalus:
Add a request logger to zanzibar

- Headers
- Body
- Method
- Host
- Timestamp
- URI Path
- Content-length

### DIFF
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -21,9 +21,7 @@
 package zanzibar
 
 import (
-	"bytes"
 	"context"
-	"io/ioutil"
 	"net/http"
 	"strconv"
 	"time"
@@ -276,25 +274,32 @@ func (router *Router) logRequest(r *http.Request) error {
 
 	fields = append(fields, zap.String("method", r.Method))
 	fields = append(fields, zap.Int64("content-length", r.ContentLength))
-	fields = append(fields, zap.String("host", r.Host))
-	body, err := ioutil.ReadAll(r.Body)
-	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
-
-	fields = append(fields, zap.String("body", string(body)))
+	fields = append(fields, zap.String("remoteAddr", r.RemoteAddr))
 	fields = append(fields, zap.String("pathname", r.URL.RequestURI()))
 	fields = append(fields, zap.String("host", r.Host))
 	fields = append(fields, zap.Time("timestamp", time.Now().UTC()))
+	// TODO add endpoint.id and endpoint.handlerId
+	// TODO log jaeger trace span
+
+	// Do not log body by default because PII and bandwidth.
+	// TODO: Add a gateway level configurable body unmarshaller
+	// to extract only non-PII info.
+
+	// body, err := ioutil.ReadAll(r.Body)
+	// r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	// fields = append(fields, zap.String("body", string(body)))
 
 	router.gateway.Logger.Info(
 		"Incoming Request",
 		fields...,
 	)
-	if err != nil {
-		fields = append(fields, zap.Error(err))
-		router.gateway.Logger.Error(
-			"Failed to Read Body",
-			fields...,
-		)
-	}
-	return err
+	// if err != nil {
+	// 	fields = append(fields, zap.Error(err))
+	// 	router.gateway.Logger.Error(
+	// 		"Failed to Read Body",
+	// 		fields...,
+	// 	)
+	// }
+	// return err
+	return nil
 }
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -210,8 +210,12 @@ func NewRouter(gateway *Gateway) *Router {
 
 func (router *Router) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	err := router.logRequest(r)
+	// TODO: Remove coverage ignore when body unmarshaling supported.
+	/* coverage ignore next line */
 	if err != nil {
+		/* coverage ignore next line */
 		http.Error(w, "can't read body", http.StatusBadRequest)
+		/* coverage ignore next line */
 		return
 	}
 	router.httpRouter.ServeHTTP(w, r)
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -268,7 +268,7 @@ func (router *Router) logRequest(r *http.Request) error {
 	fields := make([]zapcore.Field, 0)
 	for k, v := range r.Header {
 		if len(v) > 0 {
-			fields = append(fields, zap.String(k, v[0]))
+			fields = append(fields, zap.String("Header-"+k, v[0]))
 		}
 	}
 
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -99,6 +99,8 @@ var knownStatusCodes = []int{
 	http.StatusNetworkAuthenticationRequired, // 511
 }
 
+const statusCodeZapName = "statusCode"
+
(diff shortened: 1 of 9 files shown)


runtime/router.go
```diff
@@ -205,6 +211,11 @@ func NewRouter(gateway *Gateway) *Router {
 }
 
 func (router *Router) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	err := router.logRequest(r)
```
Raynos: This will fix an outstanding bad thing in RTAPI because we hook in early enough to log all 404 and 405.

Raynos: I think it would be nice to log res status code, should do this after `ServeHTTP()` ?

runtime/router.go
```diff
@@ -251,3 +262,39 @@ func (router *Router) handleMethodNotAllowed(
...
+		}
+	}
+
+	fields = append(fields, zap.String("method", r.Method))
+	fields = append(fields, zap.Int64("content-length", r.ContentLength))
+	fields = append(fields, zap.String("host", r.Host))
+	body, err := ioutil.ReadAll(r.Body)
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
+	fields = append(fields, zap.String("body", string(body)))
```
Raynos: We do not want to log body by default because PII and bandwidth.

Raynos: If we do want to log body it will be a "debug logger" which will be flipr based on x-uuid

runtime/router.go
```diff
@@ -251,3 +262,39 @@ func (router *Router) handleMethodNotAllowed(
...
+	fields = append(fields, zap.String("pathname", r.URL.RequestURI()))
+	fields = append(fields, zap.String("host", r.Host))
+	fields = append(fields, zap.Time("timestamp", time.Now().UTC()))
```
Raynos: Let's add `// TODO log jaeger trace span`

runtime/router.go
```diff
@@ -251,3 +262,39 @@ func (router *Router) handleMethodNotAllowed(
...
+
+	fields = append(fields, zap.String("method", r.Method))
+	fields = append(fields, zap.Int64("content-length", r.ContentLength))
+	fields = append(fields, zap.String("host", r.Host))
+	body, err := ioutil.ReadAll(r.Body)
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
+	fields = append(fields, zap.String("body", string(body)))
+	fields = append(fields, zap.String("pathname", r.URL.RequestURI()))
+	fields = append(fields, zap.String("host", r.Host))
```
Raynos: Let's add RemoteAddr

Also host header is duplicate.

runtime/router.go
```diff
@@ -251,3 +262,39 @@ func (router *Router) handleMethodNotAllowed(
...
 }
+
+func (router *Router) logRequest(r *http.Request) error {
+	// TODO: Allocating a fixed size array causes the zap logger to fail
+	// with ``unknown field type: { 0 0  <nil>}'' errors. Investigate this
+	// further to see if we can avoid reallocating underlying arrays for slices.
+	fields := make([]zapcore.Field, 0)
+	for k, v := range r.Header {
+		if len(v) > 0 {
+			fields = append(fields, zap.String(k, v[0]))
```
Raynos: Can we prefix this like `Header-k` ?

runtime/router.go
```diff
@@ -251,3 +262,39 @@ func (router *Router) handleMethodNotAllowed(
...
+		}
+	}
+
```
Raynos: Let's add a `// TODO add endpoint.id and endpoint.handlerId`

runtime/router.go
stevededalus: Should we just log that on response?

Raynos: Ideally a single logline with both req and res data would be good

runtime/router.go
stevededalus: Removed for now, we should add the ability for gateway owner to define custom marshallers that don't log PII

runtime/router.go
stevededalus: Done

runtime/router.go
stevededalus: I think that will be tricky without making the headers immutable somehow.

Raynos:
We need logging of response status codes for these logs to add value. They are there to see detailed information about 4xx and 5xx from our graphs.

stevededalus:
Was planning on adding responses in a follow up pull, but here's an update that hooks into a defer to add status code. We can add more response fields in a later PR.

runtime/router.go
```diff
@@ -160,7 +166,15 @@ func NewRouterEndpoint(
 func (endpoint *RouterEndpoint) HandleRequest(
 	w http.ResponseWriter, r *http.Request, params httprouter.Params,
 ) {
+	reqFields := logRequestFields(r)
+	resFields := make([]zapcore.Field, 0)
+
+	defer func() {
+		writeLogs(endpoint.gateway.Logger, reqFields, resFields)
```
Raynos: Can we just `defer writeLogs(...)` ? instead of adding a closure.

runtime/router.go
```diff
@@ -245,9 +264,63 @@ func (router *Router) handleNotFound(w http.ResponseWriter, r *http.Request) {
...
+	return fields
+}
+
+func writeLogs(l *zap.Logger, reqFlds []zapcore.Field, resFlds []zapcore.Field) {
+	fields := reqFlds
+	if resFlds != nil {
+		fields = append(reqFlds, resFlds...)
+	}
+	l.Info(
+		"Request/Response",
```
Raynos: Consider something more descriptive `"Finished an incoming server HTTP request"`

Raynos:
What's the performance impact of this change ? Can you run `benchmarks/compare_to.sh` ?

runtime/router.go
stevededalus: No, if you defer it will pass the current arrays of resFields as an argument. This is reassigned later to a different underlying array later in this method. The anonymous closure will reference the new underlying array.

Raynos: Can we make the variables pointers instead ?

stevededalus:
~10% reduction on the contact test endpoint @ 1KB.

```

    Latency    19.45ms   10.22ms 166.97ms   85.68%
    Req/Sec     1.73k   219.85     3.15k    77.42%
  622072 requests in 30.06s, 68.22MB read
  Socket errors: connect 0, read 221, write 0, timeout 6
  Non-2xx or 3xx responses: 6
Requests/sec:  20693.57
Transfer/sec:      2.27MB
```


Master
```

    Latency    17.89ms    9.08ms 169.31ms   85.96%
    Req/Sec     1.87k   257.76     3.14k    76.53%
  671910 requests in 30.08s, 73.69MB read
  Socket errors: connect 0, read 239, write 0, timeout 1
  Non-2xx or 3xx responses: 1
Requests/sec:  22340.63
Transfer/sec:      2.45MB
```

runtime/router.go
stevededalus: yeah, but I think it makes the code less readable. A defered anonymous function makes it clear that the arguments are modified in the execution of the containing function. It also makes the append messy to use in the build field methods.

Raynos:
Do you have a flamegraph ?

Raynos:
Flamegrpah shows logging overhead is dominated by fs write syscall.

stevededalus:
@Raynos 
Can't seem to share svg on github, here's a link for future reference
https://drive.google.com/open?id=0BwHuGdCRDwISZzkzNjRhWEVZT3c

PR #75, opened 2017-04-26
Generate tests for endpoints that proxy tchannel clients

ChuntaoLu:
@zhenghuiwang @Raynos


examples/example-gateway/endpoints/baz/compare_test.json
```diff
@@ -0,0 +1,33 @@
+[{
+    "testName": "successfulRequest",
+    "endpointId": "baz",
+    "handlerId": "compare",
+    "endpointRequest": {
+		"arg1": {
```
Raynos: why

Raynos:
lgtm.

PR #77, opened 2017-04-27
Switch googlenow handwritten test to use 202 instead to match the idl def

stevededalus:
Switch googlenow handwritten test to use 202 instead to match the idl def


Raynos:
Why did Travis not fail ?

stevededalus:
It seems to be failing in the travis runs but not consistently causing travis to fail. Here's the log from lu's passing branch that shows a failed run https://travis-ci.org/uber/zanzibar/jobs/226215182

PR #81, opened 2017-04-28
Move endpoint and endpoint test generation to module system

ChuntaoLu:


module/module.go
```diff
@@ -537,20 +535,20 @@ func writeFile(filePath string, bytes []byte) error {
...
-			err, "Could not open file for writing: %s", filePath,
+			err, "Could not open file for writing: \"%s\"", filePath,
 		)
 	}
 
 	n, err := file.Write(bytes)
 
 	if err != nil {
-		return errors.Wrapf(err, "Error writing to file %s", filePath)
+		return errors.Wrapf(err, "Error writing to file \"%s\"", filePath)
```
Raynos: There is a %q that will encode with the quotes. Aka `%q is `\"%s\"`

ChuntaoLu: nice, good to know, replaced them

ChuntaoLu:
@Matt-Esch 

> The configuration for all module types is supposed to live in the "config:" key. The "endpoints" array needs to live on that config object. The endpoint generation modules should generate both endpoints and tests from that config.

I should have noted the config arrangement in this PR is tentative, the main focu was to add endpoint generation into the module system with minimum config structure change. As to the actual config structure, we should come up with a proper schema before diving into change those json files.

> I think the actual config structs should remain in the gateway.go file for now, 

Those config structs were originally in template.go, which is going away as the module system takes care of code generation. I agree with the point of what module_system.go files tries to tell, I don't think it matters too much where those structs are for now since things will surely gets moved around a lot.

codegen/module_system.go
```diff
@@ -34,7 +34,11 @@ import (
 
 // NewDefaultModuleSystem creates a fresh instance of the default zanzibar
 // module system (clients, endpoints)
-func NewDefaultModuleSystem(h *PackageHelper) (*module.System, error) {
+func NewDefaultModuleSystem(
+	configDirName string,
+	middlewareConfig string,
+	h *PackageHelper,
```
Matt-Esch: Have you considered putting the middleware config information on the PackageHelper. To me it feels a bit weird to have these arguments here. Right now I have a codebase that I only generate clients in, and the change to this interface would now require me to care about non-existent middleware configs.

codegen/module_system.go
```diff
@@ -92,10 +100,68 @@ func NewDefaultModuleSystem(h *PackageHelper) (*module.System, error) {
 		)
 	}
 
-	// TODO: Register endpoint module class and type generators
+	middlewareSpecs := map[string]*MiddlewareSpec{}
```
Matt-Esch: I think the middlewares need to be modeled as a dependency but we can think about that later. It might make sense to treat them as a global grab bag (i.e. a single module of middlewares)

codegen/module_system.go
ChuntaoLu: Honestly, I didn't know a good way to get a hold of the middleware config, I know it feels awkward to pass in as arguments, it doesn't feel belong to PackageHelper either.

codegen/module_system.go
ChuntaoLu: Agreed, I like them to be explicitly specified as an endpoint's dependencies, so module system can pick them up just like clients. But they are slightly different because middleware config may need to provide input for middlewares to work.

codegen/module_system.go
Matt-Esch: I don't feel like it belongs in either place, but the PackageHelper gives us information about the structure of the package. With middlewares in this current mode, they are a permanent fixture of the package, which is arguably why the package helper does make some sense. It also have the benefit that existing consumers initialize the package helper in the current way and this configuration can be defaulted.

Matt-Esch:
👍 lgtm

PR #83, opened 2017-04-28
generate service binaries with module system

Matt-Esch:
Next step in using the module system universally is to generate the service main files as modules.

This raises a couple of issues:

  - We should be testing the binaries without the files, preferably in a chroot to prevent access to build time files. The zanizibar-defaults.json will not be available to the binary when the source directories are removed.

  - The endpoint tests are currently coupled to a binary, but in a multi-tenanted service configuration the tests need to be generated for the build and not the endpoints themselves. This change hard-couples the generated tests to the example-gateway service. They will need to be generated with a target binary in mind. It's worth thinking about how tests and other files are generated based on build configuration vs assuming that there is one static binary.


Matt-Esch:
out of scope for this change:

 - Using the service config to generate client / endpoint bootstrap code.
 - Fixing the endpoint test generation code to be generated per-binary.

Makefile
```diff
@@ -148,11 +148,11 @@ $(PROGS): $(GO_FILES)
 bins: generate $(PROGS)
 
 .PHONY: run
-run: examples/example-gateway/build/example-gateway
+run: examples/example-gateway/build/services/example-gateway/example-gateway
```
Raynos: Do we want to put the binary in `build/services/{{service}}/{{service}}` or should we have a `build/bin/{{service}}` or `bin/{{service}}` ?

Makefile
```diff
@@ -12,7 +12,7 @@ GO_FILES := $(shell \
 FILTER_LINT := grep -v -e "vendor/" -e "third_party/" -e "gen-code/" -e "codegen/templates/" -e "codegen/template_bundle/"
 
 # list all executables
-PROGS = examples/example-gateway/build/example-gateway \
+PROGS = examples/example-gateway/build/services/example-gateway/example-gateway \
```
Raynos: Please .gitignore the binary as its in this PR.

codegen/module_system.go
```diff
@@ -279,3 +302,84 @@ func readClientConfig(rawConfig []byte) (*clientClassConfig, error) {
...
+				AliasName: "",
+			},
+			{
+				PackageName: generator.packageHelper.GoGatewayPackageName() +
+					"/endpoints",
+				AliasName: "",
+			},
+		},
+		GatewayName:             instance.InstanceName,
+		RelativePathToAppConfig: filepath.Join("..", "..", ".."),
```
Raynos: Is this always hard coded ? I guess it is.

codegen/templates/endpoint_test.tmpl
```diff
@@ -32,7 +32,7 @@ func Test{{.HandlerID | Title}}{{.TestName | Title}}OKResponse(t *testing.T) {
 	gateway, err := testGateway.CreateGateway(t, nil, &testGateway.Options{
 		KnownHTTPBackends: []string{"{{$clientName | camel}}"},
 		TestBinary: filepath.Join(
-			getDirName(), "..", "..", "main.go",
+			getDirName(), "..", "..", "services", "example-gateway", "main.go",
```
Raynos: ^ We need the service name available in the template here. Hard coding it to `example-gateway` will break `edge-gateway`.

For `edge-gateway` it should be `"edge-gateway"`.

We can figure out multiple binaries / gateways in a seperate PR.

codegen/module_system.go
Matt-Esch: Just following the existing setup

codegen/templates/endpoint_test.tmpl
Matt-Esch: I'd actually argue that we need to not hard couple to a binary in the endpoint tests. If you're generating tests that target a specific binary, they should be generated as part of the tests for that binary, not for a general purpose unit.

Raynos: We need to generate tests that pass in this PR.

If you want to move test generation to be part of the binary generation then thats fine too.

Makefile
Matt-Esch: I'm preferring `bin/{{service}}`

Raynos:
A rebase should fix travis.

Matt-Esch:
That should be all issues addressed @Raynos

PR #85, opened 2017-04-28
Makefile: make test-all same as travis

Raynos:
Forgot to add `check-generate`

r: @uber/zanzibar-team


Matt-Esch:
https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html

PR #87, opened 2017-05-02
Use pending status code when logging response

stevededalus:
Use pending status code when logging response


Raynos:
Can you add the binary to .gitignore ?

PR #91, opened 2017-05-05
Add checksum to speed up easyjson

Raynos:
EasyJSON can make `make generate` take upto 60 seconds.

With the new checksum caching / short circuiting the
`make generate` step now takes 10 seconds.

r: @uber/zanzibar-team


Matt-Esch:
So now that you're not cleaning the gen-code dir, how are you going to remove excess files?

Raynos:
I dont know how we are going to remove excess files. travis will fail if you dont remove them by hand. 

We cant clean the directory if we want to re-use checksum and have faster builds :(

Matt-Esch:
Well there is a process that checks the files, can you keep a list of the files that were checked or generated from scratch and remove anything not in that list?

Raynos:
@Matt-Esch thats only for the `gen-code/*.go -> gen-code/*_easyjson.go` pipeline, not the `idl/*.thrift -> gen-code/*.go` pipeline, would only remove half the dead files.

scripts/easy_json/easy_json.go
```diff
@@ -91,38 +94,75 @@ func generate(fname string) error {
 	return nil
 }
 
+func getOldChecksum(easyJSONFile string) string {
+	oldEasyJSONBytes, err := ioutil.ReadFile(easyJSONFile)
+	if err == nil {
+		sliceStart := len(prefixBytes) + len(checksumPrefix)
+		sliceEnd := len(prefixBytes) + len(checksumPrefix) + 24
```
ChuntaoLu: nit: `const checksumLen = 24` could make 24 less magical.

Raynos:
I'll add the ability to remove generated files in a separate PR.

PR #92, opened 2017-05-05
Update clients/endpoints to use thriftrw Arg structs

Raynos:
Currently the "structs.go" that we generate ourselves is
very limited and doesn't really support proper nested
data structures and imports.

I've changed the client and endpoints to use the
thriftrw Service_Arg_Method go structs which are generated
properly and support imports/enums/pointers etc.

This also adds "required" vs "optional" support at
the arg struct level so I had to fix a bunch of tests.

r: @uber/zanzibar-team


codegen/templates/structs.tmpl
```diff
@@ -16,19 +16,3 @@ func getDirName() string {
 	_, file, _, _ := runtime.Caller(0)
 	return zanzibar.GetDirnameFromRuntimeCaller(file)
 }
```
zhenghuiwang: Do we still need this template file? seems we can delete it?

glide.lock
```diff
@@ -28,7 +28,9 @@ imports:
 - name: github.com/kardianos/osext
   version: c2c54e542fb797ad986b31721e1baedf214ca413
 - name: github.com/mailru/easyjson
-  version: 9d6630dc8c577b56cb9687a9cf9e8578aca7298a
+  version: c7e8f2b687054e448f35ff2693bd298c1a2f3a10
+  repo: https://github.com/Raynos/easyjson.git
```
zhenghuiwang: If you pin a specific version of easyjson, then  maybe always point to the master version of the customized repo, because then git repo will sync only with new stable version.

Raynos: I want to wait for bugfix to get merged, hopefully soon :)

codegen/templates/structs.tmpl
Raynos: I think we need this function `getDirName()` will double check.

codegen/method.go
```diff
@@ -64,8 +62,7 @@ type MethodSpec struct {
 	ExceptionsIndex  map[string]ExceptionSpec
 	ValidStatusCodes []int
 	// Additional struct generated from the bundle of request args.
-	RequestBoxed  bool
-	RequestStruct []StructSpec
+	RequestBoxed bool
```
Raynos: I think we kind of support it. We want to get rid of it, but that's a different PR.

codegen/templates/structs.tmpl
Raynos: Ok we only use this for endpoint tests, ill get rid of it for clients :)

PR #95, opened 2017-05-10
Update module system initializer signature

ChuntaoLu:


Raynos:
why are these extra args there in the first place ?

ChuntaoLu:
@Raynos I believe those are overlooked, they should have been included in one of my previous PR

PR #97, opened 2017-05-12
Generate req/resp converters with nested structs

Raynos:
This PR improves our existing request / response converter
functions to copy over nested structs.

I've significantly expanded and refactor the approach for
generating these and moved logic into a seperate file.

There is more work still to do be done that will come in a
seperate PR, for example lists, maps, sets, enums, etc.

r: @uber/zanzibar-team


codegen/type_converter.go
```diff
@@ -0,0 +1,193 @@
...
+// FieldGroups from one to another. It's assumed that the converted code
+// operates on two variables, "in" and "out" and that both are a go struct.
+type TypeConverter struct {
+	Lines  []string
+	Helper *PackageHelper
+}
+
+// ConvertFields will add lines to the TypeConverter for mapping
+// from one go struct to another based on two thriftrw.FieldGroups
+func (c *TypeConverter) ConvertFields(
```
ChuntaoLu: without looking at the code, to me the method signature (name and parameters) reads like converting fields between the two, while it actually generates code for conversion. IMO, `func GenConversionCode(keyPrefix string, from []*compile.FieldSpec, to []*compile.FieldSpec, helper *packageHelper) ([]string, error)` is a more intuitive signature.

codegen/type_converter.go
```diff
@@ -0,0 +1,193 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	"go.uber.org/thriftrw/compile"
+)
+
+// TypeConverter can generate a function body that converts two thriftrw
+// FieldGroups from one to another. It's assumed that the converted code
+// operates on two variables, "in" and "out" and that both are a go struct.
+type TypeConverter struct {
```
ChuntaoLu: Is this converter object necessary? feels like the `convertFields` method can be a standalone function.

codegen/type_converter.go
```diff
@@ -0,0 +1,193 @@
...
+			line = "} else {"
+			c.Lines = append(c.Lines, line)
+
+			line = "	" + prefix + " = nil"
+			c.Lines = append(c.Lines, line)
+
+			line = "}"
+			c.Lines = append(c.Lines, line)
+
+		default:
```
ChuntaoLu: Maybe default should return error?

codegen/type_converter.go
```diff
@@ -0,0 +1,193 @@
...
+			// 	return nil, err
+			// }
+			// typeName := pkgName + "." + toField.Type.ThriftName()
+			// line := prefix + "(*" + typeName + ")" + postfix
+			// c.Lines = append(c.Lines, line)
+		}
+	}
+
+	return nil
+}
```
ChuntaoLu: Can we have unit tests on this?

Raynos: :sob:

codegen/type_converter.go
Raynos: It should once we support other types, implementation is partial atm.

codegen/type_converter.go
Raynos: a standalone function passes too many args around so it put it on an object.

codegen/type_converter.go
Raynos: makes sense.

codegen/runner/post-steps.sh
```diff
@@ -15,20 +15,4 @@ EASY_JSON_BINARY="$EASY_JSON_DIR/easy_json"
...
-echo "Generated easy_json files for clients +$runtime"
-
-for file in $(find "$PREFIX/endpoints" -name "*_structs.go"); do
-    "$EASY_JSON_BINARY" -all -- "$file"
-done
-
-end=`date +%s`
-runtime=$((end-start))
-echo "Generated easy_json files for endpoints +$runtime"
+# TODO : Add new post steps :)
```
stevededalus: mind expanding the comment to list what kinds of post steps?

codegen/method.go
```diff
@@ -77,10 +77,12 @@ type MethodSpec struct {
 	DownstreamService string
 	// The downstream method spec for the endpoint
 	DownstreamMethod *MethodSpec
-	// A map from upstream to downstream field names in the requests.
-	RequestFieldMap map[string]string
-	// A map from upstream to field names to downstream types in the requests.
-	RequestTypeMap map[string]string
+
+	// Statements for converting request types
+	ConvertRequestLines []string
```
stevededalus: maybe change "Lines" to "GolangStatements" or something that makes it clear that these are lines of go code and the go code here is being build ahead of time by the convert function and not by the templates.

stevededalus: also, we need some tests on the type converter.

codegen/runner/post-steps.sh
Raynos: I dont think we have any post steps at the moment. I can delete this file.

codegen/method.go
Raynos: Makes sense.

Raynos: Tests. :sob:

Raynos:
I addressed all comments and wrote tests. cc @stevededalus @ChuntaoLu

codegen/type_converter.go
Raynos: I wrote ALL the unit tests.

PR #99, opened 2017-05-17
Move clients init file generate to module system [breaking change]

ChuntaoLu:
This PR moves the client init file generation into the module system.

**Breaking**: with this change, a `clients-config.json` file is required in the `clients` directory. And to tame the module system, the config file must have following fields as in https://github.com/uber/zanzibar/pull/99/files#diff-4eeffff9fb08e6dff5f0056d5106f422:
```
{
 	"name": "clients",
 	"type": "init",
 	"config": {}
}
```


Raynos:
@ChuntaoLu 

I think @Matt-Esch should be reviewer for this.

I believe he had some plans to generalize the "clients init" without adding a new config file by generating `clients.go` at `build/services/{{service-name}}/clients.go` next to `build/services/{{service-name}}/main.go`

ChuntaoLu:
@Raynos yeah, the config file feels unnecessary, it's currently a hack to tick the module system for a green pass, but might be useful in the future. I requested @Matt-Esch to review the change as well. I want you and @stevededalus to be aware of the breaking change in case this PR goes in.

codegen/module_system.go
```diff
@@ -134,6 +134,24 @@ func NewDefaultModuleSystem(
 		)
 	}
 
+	if err := system.RegisterClass("clients", module.Class{
```
Matt-Esch: This is an interesting way to create a global clients object. You might want to pick a slightly more obvious variant of name here because client vs clients is pretty hard to distinguish at a glance. Nice to see that the SingleModule type actually still works because it isn't used anywhere right now.

codegen/gateway.go
```diff
@@ -156,18 +160,30 @@ func NewTChannelClientSpec(jsonFile string, clientConfig *ClientClassConfig, h *
 
 // NewCustomClientSpec creates a client spec from a json file whose type is custom
 func NewCustomClientSpec(jsonFile string, clientConfig *ClientClassConfig, h *PackageHelper) (*ClientSpec, error) {
-	customImportPath, ok := clientConfig.Config["customImportPath"]
-	if !ok {
-		return nil, errors.Errorf(
-			"client config (%s) must have customImportPath field for type custom", jsonFile,
-		)
+	fields := []string{
```
ChuntaoLu: Good point, done.

codegen/module_system.go
ChuntaoLu: Yeah, I find it neat to be able to leverage the `SingleModule` type for the initialization piece. 👍 

Regarding the naming, I guess we can keep it as is for now since the config is one time only and doesn't really provide much information.

PR #101, opened 2017-05-18
Log unexpected tchannel system error

ChuntaoLu:


runtime/tchannel_server.go
```diff
@@ -177,6 +177,9 @@ func (s *TChannelRouter) handle(ctx context.Context, handler handler, service st
 		if er := reader.Close(); er != nil {
 			return errors.Wrapf(er, "could not close arg3reader for inbound call: %s::%s", service, method)
 		}
+		s.logger.Warn("Unexpected tchannel system error",
+			zap.String("error", err.Error()),
+		)
 		return call.Response().SendSystemError(err)
```
Raynos: we do not want to send the `err` out. We want to send a generic "Server Error" message.

runtime/tchannel_server.go
```diff
@@ -177,6 +177,9 @@ func (s *TChannelRouter) handle(ctx context.Context, handler handler, service st
 		if er := reader.Close(); er != nil {
 			return errors.Wrapf(er, "could not close arg3reader for inbound call: %s::%s", service, method)
 		}
+		s.logger.Warn("Unexpected tchannel system error",
+			zap.String("error", err.Error()),
```
Raynos: we should log the thrift service and the thrift method.

ChuntaoLu: Done

runtime/tchannel_server.go
ChuntaoLu: Done

PR #102, opened 2017-05-19
Add package info to module configuration

Matt-Esch:
The module system will make assumptions about where packages live, what types they export and what their internal naming conventions are. Right now this is just an adhoc process, inconsistent between module types. I've added the package info which we can begin to use as the source of truth. It won't be relevant until we can generate all modules from the module config, instead of the many arbitrary meta configurations and out of band generation processes.


module/module.go
```diff
@@ -277,14 +359,62 @@ func readDeps(jsonDeps map[string][]string) []Dependency {
...
+			"Error computing generated import string for %s",
+			targetGenDir,
+		)
+	}
+
+	return &PackageInfo{
+		PackageName:           defaultAlias,
+		PackageAlias:          defaultAlias + "Static",
+		GeneratedPackageAlias: defaultAlias + "Generated",
+		PackagePath: filepath.Join(
```
Raynos: We dont want to use `filepath.Join()` for go package names as that will use `\\` on windows which is probably invalid ? 

We can use `path.Join()` for go package names

module/module.go
```diff
@@ -464,16 +641,19 @@ type Instance struct {
...
 type JSONClassConfig struct {
-	Name         string              `json:"name"`
-	Config       interface{}         `json:"config"`
-	Dependencies map[string][]string `json:"dependencies"`
-	Type         string              `json:"type"`
+	Name              string              `json:"name"`
+	Config            interface{}         `json:"config"`
+	Dependencies      map[string][]string `json:"dependencies"`
+	Type              string              `json:"type"`
+	IsExportGenerated *bool               `json:"IsExportGenerated"`
```
Raynos: I understand now. Can we have a comment here that this field defaults to `true`.

examples/example-gateway/gateway.json
```diff
@@ -1,5 +1,6 @@
 {
 	"gatewayName": "example-gateway",
+	"packageRoot": "github.com/uber/zanzibar/example-gateway",
```
Matt-Esch: Because I am injecting a new source of truth. Changing all of the code to use this new source of truth is what I'm trying to do but I can't do it without this first. And yes this is the wrong value, I didn't pull that fix from the right commit (and we wouldn't notice because it isn't used to generate anything yet).

module/module.go
```diff
@@ -277,14 +359,62 @@ func readDeps(jsonDeps map[string][]string) []Dependency {
...
+			instanceDirectory,
+		),
+		ExportName: qualifiedInstanceName,
+		ExportType: qualifiedInstanceName + qualifiedClassName,
+		GeneratedPackagePath: filepath.Join(
+			packageRoot,
+			relativeGeneratedPath,
+			instanceDirectory,
+		),
+		IsExportGenerated: jsonConfig.IsExportGenerated == nil ||
```
Raynos: Should we do a `IsNotGenerated` boolean instead of having missing field default to `true`. That way `undefined` and `false` are "generated" and `true` is "generated"

module/module.go
```diff
@@ -277,14 +359,62 @@ func readDeps(jsonDeps map[string][]string) []Dependency {
...
+		return nil, errors.Wrapf(
+			err,
+			"Error computing generated import string for %s",
+			targetGenDir,
+		)
+	}
+
+	return &PackageInfo{
+		PackageName:           defaultAlias,
+		PackageAlias:          defaultAlias + "Static",
```
Raynos: Can we have a comment about why the PackageAlias is different between generated and non-generated.

Is it possible to have a generated Bar client and a custom Bar client with the same name Bar ?

module/module.go
```diff
@@ -277,14 +359,62 @@ func readDeps(jsonDeps map[string][]string) []Dependency {
...
+	instanceDirectory string,
+	jsonConfig *JSONClassConfig,
+	dependencies []Dependency,
+) (*PackageInfo, error) {
+	qualifiedClassName := strings.Title(camelCase(className))
+	qualifiedInstanceName := strings.Title(camelCase(jsonConfig.Name))
+	defaultAlias := camelCase(strings.ToLower(qualifiedInstanceName)) +
+		qualifiedClassName
+
+	relativeGeneratedPath, err := filepath.Rel(baseDirectory, targetGenDir)
```
Raynos: This can be computed once in the module system. If it is then this function doesn't have to return an `error`.

module/module.go
Matt-Esch: Because defaults are not the same, I couldn't find a way to configure the default for missing boolean values, and I think this is better than IsExportCustom (just because generated is clearer in what it means).

module/module.go
Raynos: Maybe the `PackagePath` should be `CustomPackagePath` to avoid confusion ?

module/module.go
```diff
@@ -559,3 +739,18 @@ func writeFile(filePath string, bytes []byte) error {
 func closeFile(file *os.File) {
 	_ = file.Close()
 }
+
+var camelingRegex = regexp.MustCompile("[0-9A-Za-z]+")
+
+func camelCase(src string) string {
```
Matt-Esch: It causes a circular reference and there's no clear place to share the functionality. That's why I copied it. I'm assuming we can resolve this in future  by moving module code into the codegen package at some point. TBD.

Raynos: it doesnt cause cycles if the function stays in module system.

Aka `codegen` can import `CamelCase` from `module` BUT `module` CANNOT import `CamelCase` from `codegen`.

module/module_test.go
```diff
@@ -146,6 +146,7 @@ func TestExampleService(t *testing.T) {
 	// TODO: this doesn't yet generate the build to a dir
 	// TODO: this should return a collection of errors if they occur
 	err = moduleSystem.GenerateBuild(
+		"github.com/uber/zanzibar",
```
Matt-Esch: Not really. It doesn't matter because as the comment says, we're not actually generating a build in this test. It doesn't matter until we're actually using it to generate code. The package name would be the contents of the example service being generated for the test case (which doesn't exist).

module/module.go
Matt-Esch: +1

module/module.go
Matt-Esch: It's simply one package per directory, one unique identifier per package.

PR #103, opened 2017-05-22
Allow multiple thrift services for tchannel client

ChuntaoLu:


codegen/templates/tchannel_client.tmpl
```diff
@@ -47,21 +46,19 @@ func NewClient(gateway *zanzibar.Gateway) *{{$clientName}} {
...
-	thriftService string
 	client        zanzibar.TChannelClient
 }
 
+{{range $svc := .Services}}
 {{range .Methods}}
-	// {{.Name}} ...
-	func (c *{{$clientName}}) {{.Name}}(
+	// {{title $svc.Name}}{{title .Name}} is a client RPC call for method "{{.Name}}" of thrift service "{{$svc.Name}}"
+	func (c *{{$clientName}}) {{title $svc.Name}}{{title .Name}}(
```
Raynos: This method name will get quite long for our generated clients.

Raynos:
It might be nice to change our client generation semantics to only generate the methods we need, with shorter names.

If the client-config.json had a mapping of method names to generate, e.g.

```
{
    config: {
        methods: {
            "Call": "SimpleService::Call"
        }
    }
}
```

This will reduce the amount of code generated and will allow us to give shroter method names instead of the long form.

ChuntaoLu:
@Raynos Agreed, I have thought about it, but ended up with the simplest solution here. I will try to make it happen.

codegen/module_system.go
```diff
@@ -690,6 +691,16 @@ func (g *EndpointGenerator) generateEndpointFile(
 			strings.Title(method.Name) + "Endpoint"
 	}
 
+	// TODO: http client needs to support multiple thrift services
+	var clientMethodName string
+	if method.DownstreamMethod != nil {
+		clientMethodName = method.DownstreamMethod.Name
+		if e.WorkflowType == "tchannelClient" {
+			clientMethodName = method.DownstreamService + strings.Title(clientMethodName)
+
```
zhenghuiwang: extra line?

codegen/module_system.go
```diff
@@ -690,6 +691,16 @@ func (g *EndpointGenerator) generateEndpointFile(
 			strings.Title(method.Name) + "Endpoint"
 	}
 
+	// TODO: http client needs to support multiple thrift services
+	var clientMethodName string
+	if method.DownstreamMethod != nil {
```
zhenghuiwang: Can `method.DownstreamMethod` be `nil` here? (i don't think it can)

If it can, an error should be thrown here. If it can't, this `if` should be removed

ChuntaoLu: For a non-proxy custom endpoint, it can, and it's not an error.

codegen/module_system.go
ChuntaoLu: Interesting, not sure how it gets in here

codegen/module_system.go
zhenghuiwang: If it is nil, `clientMethodName` will be `""` after this if. Shouldn't it be an error to render template with `""` `ClientMethodName`?

ChuntaoLu: The usage of `clientMethodName` is guarded by a downstream existence [check](https://github.com/uber/zanzibar/pull/103/files/b135fde02cc527dfafa160c6810557a624660d88#diff-2426caa827592efd9d55743dd4943e3cR118), it's only used when downstream exists.

Raynos:
lgtm.

PR #104, opened 2017-05-22
Add support for map/list/typedef/enum to converter

Raynos:
This PR adds more features to TypeConverter class.

Map, List, typedef and enum have been added.
Set is NOT implemented yet.

This PR does not add any tests yet, that will come
shortly.

r: @uber/zanzibar-team

### DIFF
--- a/codegen/type_converter.go
+++ b/codegen/type_converter.go
@@ -401,7 +401,7 @@ func (c *TypeConverter) genStructConverter(
 			// 	return nil, err
 			// }
 			// typeName := pkgName + "." + toField.Type.ThriftName()
-			// line := prefix + "(*" + typeName + ")" + postfix
+			// line := toIdentifier + "(*" + typeName + ")" + postfix
 			// c.Lines = append(c.Lines, line)
 		}
 	}
--- a/codegen/type_converter.go
+++ b/codegen/type_converter.go
@@ -179,8 +179,8 @@ func (c *TypeConverter) genStructConverter(
 
 			subFromFields := fromFieldType.Fields
 			err = c.genStructConverter(
-				strings.Title(toField.Name)+".",
-				"	",
+				keyPrefix+strings.Title(toField.Name)+".",
+				indent+"	",
 				subFromFields,
 				subToFields,
 			)
--- a/codegen/type_converter.go
+++ b/codegen/type_converter.go
@@ -49,6 +49,12 @@ func (c *TypeConverter) append(parts ...string) {
 	c.Lines = append(c.Lines, line)
 }
 
+// appendf helper will add a formatted line to TypeConverter
+func (c *TypeConverter) appendf(format string, parts ...interface{}) {
+	line := fmt.Sprintf(format, parts...)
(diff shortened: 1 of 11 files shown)


codegen/type_converter.go
```diff
@@ -42,6 +43,252 @@ type PackageNameResolver interface {
...
+	indent string,
+) error {
+	typeName, err := c.getGoTypeName(toFieldType.ValueSpec)
+	if err != nil {
+		return err
+	}
+
+	valueStruct, isStruct := toFieldType.ValueSpec.(*compile.StructSpec)
+	if isStruct {
+		line := toIdentifier + " = make([]*" +
```
stevededalus: these lines might be easier to read with fmt.Sprintf instead of concatanating.

codegen/type_converter.go
```diff
@@ -42,6 +43,252 @@ type PackageNameResolver interface {
...
+		if err != nil {
+			return err
+		}
+	} else {
+		line = toIdentifier + "[index] = " + typeName + "(value)"
+		c.Lines = append(c.Lines, line)
+	}
+
+	line = "}"
+	c.Lines = append(c.Lines, line)
```
stevededalus: Just inline the single character lines.
c.Lines = append(c.Lines, "}")

codegen/type_converter.go
Raynos: seems reasonable.

codegen/type_converter.go
Raynos: slipper slope. but can move constant strings into one line statements.

codegen/type_converter.go
```diff
@@ -42,6 +43,252 @@ type PackageNameResolver interface {
...
+	toIdentifier string,
+	fromIdentifier string,
+) error {
+	var line string
+	typeName, err := c.getGoTypeName(toField.Type)
+	if err != nil {
+		return err
+	}
+
+	if toField.Required {
```
ChuntaoLu: Feels like I am missing something, why does the conversion differ based on `required` or `optional`?

Raynos: pointer vs not pointer.

codegen/type_converter.go
```diff
@@ -42,6 +43,252 @@ type PackageNameResolver interface {
...
+	fromFieldStruct, ok := fromFieldType.(*compile.StructSpec)
+	if !ok {
+		return errors.Errorf(
+			"could not convert struct fields, "+
+				"incompatible type for %s :",
+			toFieldName,
+		)
+	}
+
+	line := "if " + fromIdentifier + " != nil {"
```
zhenghuiwang: if you have `func (c *TypeConverter) Append(parts ...string)` for adding new line to 'c.Line', you can have more compact/readable generated code block.
```
c.Append("if ", fromIdentifier, " != nil")
c.Append("    ", toIdentifier, " = &", type, "{}")
....
```

Raynos: good idea.

Raynos:
Addressed comments, written tests, added 100% coverage.

PR #105, opened 2017-05-22
Set full type for request and respone

ChuntaoLu:
Currently we assume all request and response types are struct pointer type, and prefix `*` in our templates. This assumption is not always valid.
This PR sets the full request and response type without assuming it is a struct pointer type.

### DIFF
--- a/codegen/method.go
+++ b/codegen/method.go
@@ -198,12 +198,15 @@ func (ms *MethodSpec) setRequestType(curThriftFile string, funcSpec *compile.Fun
 		ms.RequestType, err = packageHelper.TypeFullName(
 			curThriftFile, funcSpec.ArgsSpec[0].Type,
 		)
+		if err == nil && isStructType(funcSpec.ArgsSpec[0].Type) {
+			ms.RequestType = "*" + ms.RequestType
+		}
 	} else {
 		ms.RequestBoxed = false
 
 		goPackageName, err := packageHelper.TypePackageName(curThriftFile)
 		if err == nil {
-			ms.RequestType = goPackageName + "." +
+			ms.RequestType = "*" + goPackageName + "." +
 				ms.ThriftService + "_" + strings.Title(ms.Name) + "_Args"
 		}
 	}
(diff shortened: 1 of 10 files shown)


codegen/method.go
```diff
@@ -198,12 +198,15 @@ func (ms *MethodSpec) setRequestType(curThriftFile string, funcSpec *compile.Fun
 		ms.RequestType, err = packageHelper.TypeFullName(
 			curThriftFile, funcSpec.ArgsSpec[0].Type,
 		)
+		if err == nil && isStructType(funcSpec.ArgsSpec[0].Type) {
```
Raynos: can we add TODO here to check typedef semantics.

codegen/method.go
```diff
@@ -225,6 +228,9 @@ func (ms *MethodSpec) setResponseType(curThriftFile string, respSpec *compile.Re
 		return nil
 	}
 	typeName, err := packageHelper.TypeFullName(curThriftFile, respSpec.ReturnType)
+	if isStructType(respSpec.ReturnType) {
```
Raynos: can we add a TODO here to check typedef semantics.

codegen/templates/endpoint.tmpl
```diff
@@ -250,8 +250,8 @@ func (w {{$workflow}}) Handle(
 }
 
 {{if and (ne .RequestType "") (ne $clientReqType "") -}}
-func convertTo{{title .Name}}ClientRequest(in *{{.RequestType}}) *{{$clientReqType}} {
-	out := &{{$clientReqType}}{}
+func convertTo{{title .Name}}ClientRequest(in {{.RequestType}}) {{$clientReqType}} {
+	out := &{{unref $clientReqType}}{}
```
Raynos: `&string{}` will probably break here ? 

We need some kind of "is primitive" check and not generate converters for non-struct types.

codegen/method.go
ChuntaoLu: typedef should be covered since the type assertion is against the `RootTypeSpec` of given spec. https://github.com/uber/zanzibar/blob/master/codegen/method.go#L217

codegen/templates/endpoint.tmpl
ChuntaoLu: That's good point.

codegen/method.go
ChuntaoLu: same as above

codegen/method.go
Raynos: #coolstory

PR #106, opened 2017-05-22
resolve module dependencies

Matt-Esch:
Populates the resolved dependencies for each module instance in the module system. This will be used by the dependency struct generation code.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -269,6 +269,54 @@ func (system *ModuleSystem) ResolveModules(
 					append(resolvedDependencies, dependencyInstance)
 			}
 		}
+
+		// Resolve the class dependencies
+		for _, classInstance := range classInstances {
+			for _, classDependency := range classInstance.Dependencies {
+				moduleClassInstances, ok :=
+					resolvedModules[classDependency.ClassName]
+
+				if !ok {
+					return nil, errors.Errorf(
+						"Invalid class name \"%q\" in dependencies for %s %s",
+						classDependency.ClassName,
+						classInstance.ClassName,
+						classInstance.InstanceName,
+					)
+				}
+
+				// TODO: We don't want to linear scan here
+				var dependencyInstance *ModuleInstance
+
+				for _, instance := range moduleClassInstances {
+					if instance.InstanceName == classDependency.InstanceName {
+						dependencyInstance = instance
+						break
+					}
+				}
+
+				if dependencyInstance == nil {
+					return nil, errors.Errorf(
+						"Unknown %q class depdendency \"%q\""+
+							"in dependencies for %s %s",
+						classDependency.ClassName,
+						classDependency.InstanceName,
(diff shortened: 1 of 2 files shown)


codegen/module.go
```diff
@@ -269,6 +269,54 @@ func (system *ModuleSystem) ResolveModules(
 					append(resolvedDependencies, dependencyInstance)
 			}
 		}
+
+		// Resolve the class dependencies
+		for _, classInstance := range classInstances {
```
Raynos: looks like merge-conflict. we have this loop twice

codegen/module_test.go
```diff
@@ -169,3 +256,185 @@ func getTestDirName() string {
 	// If dirname is not absolute then its a package name...
 	return filepath.Join(os.Getenv("GOPATH"), "src", dirname)
 }
+
+func compareInstances(
```
Raynos: this method can be replaced with `assert.Equal(t, actual, expected)`

Matt-Esch: It's not fully equal because there is binary content, I'm asserting a subset

codegen/module.go
Matt-Esch: Yes we have this loop twice because the current class instances need to be fully resolved before you can resolve the dependencies (because clients can have peer dependencies, for example). So basically you have to create all possible dependency ModuleInstance objects before you can populate the ResolvedDependencies. I don't think this is incorrect.

codegen/module.go
```diff
@@ -269,6 +269,54 @@ func (system *ModuleSystem) ResolveModules(
...
+			for _, classDependency := range classInstance.Dependencies {
+				moduleClassInstances, ok :=
+					resolvedModules[classDependency.ClassName]
+
+				if !ok {
+					return nil, errors.Errorf(
+						"Invalid class name \"%q\" in dependencies for %s %s",
```
ChuntaoLu: I am guessing `"Invalid class name %q in dependencies for %q %q"` is what you want here?

codegen/module.go
```diff
@@ -269,6 +269,54 @@ func (system *ModuleSystem) ResolveModules(
...
+				for _, instance := range moduleClassInstances {
+					if instance.InstanceName == classDependency.InstanceName {
+						dependencyInstance = instance
+						break
+					}
+				}
+
+				if dependencyInstance == nil {
+					return nil, errors.Errorf(
+						"Unknown %q class depdendency \"%q\""+
```
ChuntaoLu: Same as above, redundant quotes for `%q`

Matt-Esch: I'm not entirely sure that %q is right at all, but I will try to make it consistent at least.

codegen/module.go
Raynos: Can you leave a comment, the loops looked identical at a glance.

PR #107, opened 2017-05-23
add instrumentation to http client dial

srivastavankit:
- first attempt at adding instrumentation only for http dial

### DIFF
--- a/runtime/http_client.go
+++ b/runtime/http_client.go
@@ -21,7 +21,13 @@
 package zanzibar
 
 import "net/http"
-import "go.uber.org/zap"
+import (
+	"net"
+	"time"
+
+	"github.com/uber-go/tally"
+	"go.uber.org/zap"
+)
 
 // HTTPClient defines a http client.
 type HTTPClient struct {
@@ -29,6 +35,7 @@ type HTTPClient struct {
 
 	Client  *http.Client
 	Logger  *zap.Logger
+	Scope   *tally.Scope
 	BaseURL string
 }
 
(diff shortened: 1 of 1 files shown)


runtime/http_client.go
```diff
@@ -21,32 +21,68 @@
...
+	Scope   *tally.Scope
 	BaseURL string
 }
 
 // NewHTTPClient will allocate a http client.
 func NewHTTPClient(
 	gateway *Gateway, baseURL string,
 ) *HTTPClient {
+	scope := gateway.MetricScope.SubScope("") // TODO: get http client name
+	timer := scope.Timer("dial")
```
Raynos: nitpick: a more specific metric name like `http-client.dial` might be nicer.

runtime/http_client.go
```diff
@@ -21,32 +21,68 @@
...
 			},
 		},
 		BaseURL: baseURL,
 	}
 }
+
+func getInstrumentedDial(timer tally.Timer, successCtr, errorCtr tally.Counter) func(string, string) (net.Conn, error) {
+	return func(n, a string) (conn net.Conn, err error) {
+		stat := instrument(timer, successCtr, errorCtr)
+		defer func() {
```
Raynos: nitpick: i find defers hard to read because of unwinding.

```
stat := instrument(timer, successCtr, errorCtr);
conn, err := net.Dial(n, a)
stat(err)
return conn, err
```

is easier to read.

runtime/http_client.go
```diff
@@ -21,32 +21,68 @@
...
 
 	Client  *http.Client
 	Logger  *zap.Logger
+	Scope   *tally.Scope
 	BaseURL string
 }
 
 // NewHTTPClient will allocate a http client.
 func NewHTTPClient(
 	gateway *Gateway, baseURL string,
```
Raynos: we can add an extra argument here for httpClientName.

Then update the template ( https://github.com/uber/zanzibar/blob/master/codegen/templates/http_client.tmpl#L37 ), should be `"{{.Name}}"`

Raynos:
PR is stale. Will close for now. Feel free to re-open if still wanted.

PR #112, opened 2017-05-26
Add support for reading headers into body.

Raynos:
This PR adds support for the header annotation in a
thrift file that can be used to say "this field in the
thrift struct comes from a header"

For example:

```
    BarResponse argWithHeaders (
        1: required string name
        2: optional string userUUID (
            zanzibar.http.ref = "headers.x-uuid"
            go.tag = "json:\"-\""
        )
    )
```

In this implementation we only add support to the
generated code for reading server request headers into
the body. ( not in the client and not in the server response ).

r: @uber/zanzibar-team


stevededalus:
LGTM

PR #113, opened 2017-05-29
Thread generated module specs through dependencies

Matt-Esch:
Some module generators will generate specs. This change supports returning a generated spec from a generator, which mutates a private field on the ModuleInstance for sharing the spec through ResolvedModules. The intended use case is to share the client specs with the endpoint generators so that proxy endpoints can generate code from the client spec.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -498,7 +498,7 @@ func (system *ModuleSystem) GenerateBuild(
 				continue
 			}
 
-			files, err := generator.Generate(classInstance)
+			buildResult, err := generator.Generate(classInstance)
 
 			if err != nil {
 				fmt.Printf(
@@ -511,7 +511,13 @@ func (system *ModuleSystem) GenerateBuild(
 				return err
 			}
 
-			for filePath, content := range files {
+			if buildResult == nil {
+				continue
+			}
+
+			classInstance.genSpec = buildResult.Spec
+
+			for filePath, content := range buildResult.Files {
 				filePath = filepath.Clean(filePath)
 
 				if strings.HasPrefix(filePath, "..") {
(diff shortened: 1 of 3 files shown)


codegen/module.go
```diff
@@ -673,6 +693,12 @@ type ModuleInstance struct {
 	JSONFileRaw []byte
 }
 
+// GetGeneratedSpec returns the last spec result returned for the module
+// instance
+func (instance *ModuleInstance) GetGeneratedSpec() interface{} {
```
ChuntaoLu: nit: per https://golang.org/doc/effective_go.html#Getters, `GenSpec` is the idiomatic Go method name

codegen/module_system.go
```diff
@@ -255,7 +255,7 @@ type HTTPClientGenerator struct {
 // path (relative to the target build directory) to file bytes.
```
ChuntaoLu: The comments of all the `Generate` method need to be updated to reflect the return value change.

codegen/module_system.go
```diff
@@ -639,7 +648,10 @@ func (g *EndpointGenerator) Generate(
 			)
 		}
 	}
-	return ret, nil
+	return &BuildResult{
+		Files: returnFiles,
+		Spec:  nil,
```
ChuntaoLu: Would it be a good idea to return spec (if none-nil) for every instance? This way the clients/endpoint initializer module that depends on all clients/endpoints would have the specs right off the shelve.

Matt-Esch:
> I guess this PR is a first step of many

There are at least 10 prs that are stacked on top of this change. I started with trying to generate structs.

Matt-Esch:
Comments addressed and rebased. The return value for GenerateBuild will be added in a different pr

PR #114, opened 2017-05-31
Use resolved module client specs

Matt-Esch:
Rather than sharing client specs on a globally shared client spec array, we should use the module ResolvedDependencies to retrieve the client specs. The generated specs are shared as part of the generator BuildResult, and will be available to the module generator's Generate function through the instance.ResolvedDependencies argument.


codegen/module.go
```diff
@@ -284,18 +298,6 @@ func (system *ModuleSystem) ResolveModules(
 		if err != nil {
 			return nil, err
 		}
-
-		// Resolve dependencies again in case two instances
```
Raynos: I swear we have to double resolve because of peer deps anyway... This might break peer deps. Needs tests to verify.

codegen/gateway.go
```diff
@@ -685,7 +685,7 @@ func (e *EndpointSpec) EndpointTestConfigPath() string {
 
 // SetDownstream configures the downstream client for this endpoint spec
 func (e *EndpointSpec) SetDownstream(
-	clientModules map[string]*ClientSpec,
```
Raynos: because there is a range this doesnt require any further changes.. ffff---- golang.

codegen/module.go
Matt-Esch: My comment originally was that you need to _read_ the modules before you _resolve_ the modules. I thought the confusion was between the two for loops read and resolve instead of resolving in the read loop, not the fact that resolve was happening twice after read, which is clearly wrong.

Raynos: Resolving twice could in theory be necessary because the first resolve mutates an array that the second resolve reads. Aka you need to keep re-resolving until the array stops growing...

Unless the resolve runs in a deterministic order where it never grows with a second call.

Matt-Esch: No that's not what resolving is. Resolving means taking the Dependencies and turning them into ResolvedDependencies, by pulling out the module instance from the global list of module instance. So Dependencies in a map of Class to InstanceName, something like

client: ["foo", "bar", "baz"]

The ResolveDependencies function is building in reverse DAG order the dependencies, so it reads the clients, resolves the client dependencies, reads the business dir, resovles the dependencies, reads the endpoints, resolves the endpoint dependencies.

Because we do it in order, you should only have to read once and resolve once. It's just replacing the instance name with the instance object that we read. The reason why we can't read and resolve at the same time is because if client Foo depends on client Bar, it might not exist yet. A missing dependency in this resolve function should be an error, because there is no chance of it being read in future. Once we had read all clients, all clients will be resolvable.

PR #115, opened 2017-06-01
Improve workspace support

Raynos:
Improve support for "workspace" and multiple GOPATH

 - Fixed PackageHelper to not rely on "gatewayNamespace" and instead rely surely on packageRoot
 - Fixed code to not generate "package name" from thrift file which is incorrect, instead use module system
 - Fixed makefile to support "workspace"
 - Fixed makefile & code coverage tooling to be GOPATH agnostic.

r: @uber/zanzibar-team

### DIFF
--- a/scripts/easy_json/easy_json.go
+++ b/scripts/easy_json/easy_json.go
@@ -127,6 +127,7 @@ func main() {
 		return
 	}
 
+	// Do not paralleliz, not worth it.
 	for _, file := range files {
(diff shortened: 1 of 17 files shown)


scripts/easy_json/easy_json.go
```diff
@@ -127,6 +127,7 @@ func main() {
 		return
 	}
 
+	// Do not paralleliz, not worth it.
```
stevededalus: sp : parallelize

stevededalus:
LGTM

PR #116, opened 2017-06-01
Add support for Transforms that map fields in the Type Converter

stevededalus:
Generates code that allows fields of the same type to be mapped to one another based on a provided mapping of field names. Performs overwritting if provided flag is set.

Remaining work for future PRs:
- [X] Wire transform configs from a middleware config file in the the type map
- [ ] Add nil checks for deeply nested structs (that can contain required fields nested in optional fields)
- [ ] Add camelCasing and other transforms
- [ ] Ability to remove mapped field
- [x] Support mapping optional -> required using zero-values and pointer dereferencing


codegen/type_converter.go
```diff
@@ -55,6 +55,25 @@ func (c *TypeConverter) appendf(format string, parts ...interface{}) {
...
+	fromIdentifier string,
+	overriddenIdentifier string,
+) {
+	if overriddenIdentifier == "" {
+		c.append(indent, toIdentifier, " = ", typeName, "(", fromIdentifier, ")")
+		return
+	}
+
+	// TODO: Verify all intermediate objects are not nil
+	c.append(indent, toIdentifier, " = ", typeName, "(", overriddenIdentifier, ")")
```
Raynos: Why do we assign from both fields ? 

It looks like a transform should only support the transformed field not the original field.

codegen/type_converter_test.go
```diff
@@ -966,3 +1026,542 @@ func TestConvertWithBadKeyMapOfString(t *testing.T) {
...
+	`), lines)
+}
+
+func TestConverterMapOverrideOptional(t *testing.T) {
+	fieldMap := make(map[string]codegen.FieldMapperEntry)
+	fieldMap["One"] = codegen.FieldMapperEntry{
+		QualifiedName: "Two",
+		Override:      true,
+	} // Map from required filed, unconditional assignment
+	fieldMap["Two"] = codegen.FieldMapperEntry{
```
Raynos: It's confusing to set "override: true" but still have the field come from `in.Two` instead of `in.One`

Maybe we should make both fields required if override is true and fail the code generation if one of them is optional.

codegen/type_converter_test.go
```diff
@@ -966,3 +1026,542 @@ func TestConvertWithBadKeyMapOfString(t *testing.T) {
...
+		QualifiedName: "One",
+		Override:      true,
+	}
+	fieldMap["Three"] = codegen.FieldMapperEntry{
```
Raynos: Override true should preferablly not read from `in.Three`

codegen/type_converter_test.go
```diff
@@ -966,3 +1026,542 @@ func TestConvertWithBadKeyMapOfString(t *testing.T) {
...
+		QualifiedName: "Four",
+		Override:      true,
+	}
+	fieldMap["Four"] = codegen.FieldMapperEntry{
+		QualifiedName: "Three",
+		Override:      false,
```
Raynos: Very sneaky but correct.

codegen/type_converter_test.go
```diff
@@ -966,3 +1026,542 @@ func TestConvertWithBadKeyMapOfString(t *testing.T) {
...
+			2: optional bool two
+		}`,
+		nil,
+		fieldMap,
+	)
+
+	assert.NoError(t, err)
+	assert.Equal(t, trim(`
+		out.One = (*bool)(in.Two)
+		if in.One != nil {
```
Raynos: Might be cleaner to generate an `if ... {} else {}` to reduce confusion.

codegen/type_converter_test.go
```diff
@@ -966,3 +1026,542 @@ func TestConvertWithBadKeyMapOfString(t *testing.T) {
...
+				out.Three.One = string(in.Three.Two)
+			}
+			out.Three.Two = (*string)(in.Three.Two)
+		} else {
+			out.Three = nil
+		}
+		if in.Four != nil {
+			out.Four = &structs.NestedBar{}
+			out.Four.One = string(in.Four.One)
+			out.Four.Two = (*string)(in.Four.One)
```
Raynos: cannot cast string to `*string` generated code does not work.

Need to use `ptr.String(in.Four.One)` to cast from required to optional.

codegen/type_converter_test.go
```diff
@@ -96,16 +98,27 @@ func convertTypes(
...
 	converter := newTypeConverter()
 	program, err := compileProgram(content, otherFiles)
 	if err != nil {
 		return "", err
 	}
+	overrideMap, err = addSpecToMap(overrideMap,
+		program.Types[toStruct].(*compile.StructSpec).Fields,
+		"")
+
+	fmt.Printf("%v", overrideMap)
```
stevededalus: Done

codegen/type_converter_test.go
stevededalus: Add md file to https://github.com/uber/zanzibar/pull/116/files#diff-ab1cd02cf41c42554ab67a9479e64e29

PR #118, opened 2017-06-02
generate client from module instance

Matt-Esch:
First attempt to standardize on using the module instance as the source of truth. I am using the module instance PackageInfo to define the export type, export name and globally unique aliases when importing in generated code.


Matt-Esch:
Worth noting that the default name for Googlenow client is now defined in the module system, and it changes from GoogleNow. We may wish to change the way that this name is computed, but it is standardized via the computed moduleInstance.PackageInfo. Updating how PackageInfo is computed should update all references without breaking. Consistency and single source of truth, can you imagine it.

Raynos:
I rebased it for you.

PR #122, opened 2017-06-03
Support sub directories in module system

ChuntaoLu:
This PR enables the module system to digest module configs located in the sub-directories of a module class.
For example:
```
examples/example-gateway/endpoints/
├── bar
│   ├── arg_not_struct.json
│   ...
│   └── endpoint-config.json
└── tchannel
    └── baz
        ├── baz_call.go
        ├── call.json
        └── endpoint-config.json
```


codegen/package.go
```diff
@@ -52,46 +54,35 @@ type PackageHelper struct {
...
-	relativeGenDir, err := filepath.Rel(absConfigDir, genDir)
-	if err != nil {
-		return nil, errors.Wrapf(
-			err, "Could not compute relative dir for %q", genDir,
+			"%s is not valid path: %s", configRoot, err,
 		)
 	}
 
-	goGatewayNamespace := filepath.Join(packageRoot, relativeGenDir)
+	goGatewayNamespace := filepath.Join(packageRoot, relTargetGenDir)
```
Matt-Esch: technically this should be `path.Join` not `filepath.Join`?

codegen/module.go
```diff
@@ -119,12 +119,16 @@ func (system *ModuleSystem) RegisterClass(
...
+// ModuleClassOption allows customization with the module class.
+type ModuleClassOption func(*ModuleClass)
+
 // RegisterClassType registers a type generator for a specific module class
 // For example, the "http"" type generator for the "Endpoint"" class
 func (system *ModuleSystem) RegisterClassType(
 	className string,
 	classType string,
 	generator BuildGenerator,
+	opts ...ModuleClassOption,
```
Matt-Esch: I generally dislike this approach. We see this sort of code littered around our codebases where he have "preHandleHooks" and stuff. The main issues are that it doesn't document itself and it relies on mutation. The class types should be clearly configurable in one way.

My original suggestion was to make the nested directory a static property of the ClassType definition. So for 
the Class "Endpoint" you can register a ClassType "http", and for that type you can optionally choose the subdirectory for that type.

codegen/module.go
```diff
@@ -303,6 +296,86 @@ func (system *ModuleSystem) ResolveModules(
...
+	if err != nil {
+		return nil, errors.Wrapf(err,
+			"Error relative class directory for %q",
+			className,
+		)
+	}
+
+outer:
+	for _, file := range files {
+		for _, dir := range class.SubDirs {
```
Matt-Esch: This will probably be cleaner if you have an IsSubDir function so you can avoid the nested for loop and the labelled break.

codegen/module_system.go
```diff
@@ -156,55 +156,58 @@ func NewDefaultModuleSystem(
...
-			"Error registering service class",
+			"Error registering HTTP endpoint class type",
 		)
 	}
 
-	if err := system.RegisterClassType("service", "gateway", &GatewayServiceGenerator{
+	if err := system.RegisterClassType("endpoint", "tchannel", &EndpointGenerator{
 		templates:     tmpl,
 		packageHelper: h,
+	}, func(mc *ModuleClass) {
```
Raynos: This pattern of mutation with anon functions is weird.

codegen/package.go
ChuntaoLu: Done

codegen/module.go
ChuntaoLu: Done

codegen/module_system.go
ChuntaoLu: Removed

codegen/module.go
ChuntaoLu: Removed the param, setting explicit subdirs when registering a module class

ChuntaoLu: scratch that, let's have arbitrarily nested sub directories.

PR #123, opened 2017-06-06
Remove example-gateway details from test/lib/bench_gateway

Raynos:
Currently the test/lib/bench_gateway.BenchGateway class
only works with the example-gateway because it's hardcoded
to load the example-gateway config.

This PR removes that and updates our tests to properly configure
the BenchGateway so that it works.

I've updated the tchannel TestBackend to set default timeout
configurations to reduce boilerplate.

Also fixed a bug in StaticConfig in passing where it had
mutable shared state if the seedConfig is re-used.

r: @uber/zanzibar-team

### DIFF
--- a/config/production.json
+++ b/config/production.json
@@ -2,6 +2,7 @@
 	"serviceName": "my-gateway",
 	"port": 4000,
 	"env": "production",
+	"useDatacenter": false,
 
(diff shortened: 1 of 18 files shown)


Raynos:
Oops, `make fast-bench` fails, will fix that up.

PR #129, opened 2017-06-13
Add support for parsing query params in server.

Raynos:
This adds initial support for query params in the server runtime
and the endpoint code generation.

TODO:

 - [x] write tests
 - [x] Ensure we have http.ref header support + query params
 - [x] Fail client generation, HTTP GET + body is for server-only
    If we try HTTP CLIENT GET + body it should have the old semantics
    of saying body is not allowed for GET request.
 - [x] get 100% coverage.
 - [x] enforce required vs optional semantics for query params.
 - [x] Ensure we future-proof for other http.ref by failing query
    param parser for unknown http.ref semantics.

r: @uber/zanzibar-team


codegen/method.go
```diff
@@ -538,16 +551,115 @@ func (ms *MethodSpec) setTypeConverters(
...
+	structType := compile.FieldGroup(funcSpec.ArgsSpec)
+
+	for _, field := range structType {
+		realType := compile.RootTypeSpec(field.Type)
+		fieldName := field.Name
+		identifierName := camelCase(fieldName) + "Query"
+
+		httpRefAnnotation := field.Annotations[antHTTPRef]
+		// If a thrift field has a http.ref annotation then we
+		// should not read this field from query parameters.
```
zhenghuiwang: Can we move this comment at the beginning of this function?

codegen/method.go
```diff
@@ -186,9 +192,16 @@ func NewMethod(
...
-		return nil, errors.Errorf(
-			"invalid annotation: HTTP GET method with body type",
-		)
+		if !method.IsEndpoint {
+			return nil, errors.Errorf(
+				"invalid annotation: HTTP GET method cannot have a body",
+			)
+		}
+
+		err := method.setQueryParamStatements(funcSpec)
```
zhenghuiwang: Nit: better change the signature of `setQueryParamStatements` to 

```
func queryParamStatements(*compile.FunctionSpec) ([]string, error)
```

since the func doesn't have to depend on `MethodSpec` type.

codegen/method.go
```diff
@@ -538,16 +551,115 @@ func (ms *MethodSpec) setTypeConverters(
...
+			statements.append("\treturn")
+			statements.append("}")
+		} else {
+			statements.appendf("%s := req.HasQueryValue(%q)",
+				okIdentifierName, fieldName,
+			)
+			statements.appendf("if %s {", okIdentifierName)
+		}
+
+		var pointerMethod string
```
zhenghuiwang: For readability, if you extract line 601-645 as a function, you can merge two if's at line 587 and line 647. The two cases will be separated clearer.

codegen/method.go
```diff
@@ -538,16 +551,115 @@ func (ms *MethodSpec) setTypeConverters(
...
+			continue
+		}
+
+		okIdentifierName := camelCase(fieldName) + "Ok"
+		if field.Required {
+			statements.appendf("%s := req.CheckQueryValue(%q)",
+				okIdentifierName, fieldName,
+			)
+			statements.appendf("if !%s {", okIdentifierName)
+			statements.append("\treturn")
```
zhenghuiwang: For better error handling, `setQueryParamStatements` should construct a Go function with error return type, instead of lines of code. So that the endpoint code can properly log the error and control the flow of handling requests.

codegen/method.go
```diff
@@ -538,16 +551,115 @@ func (ms *MethodSpec) setTypeConverters(
...
+				identifierName, fieldName,
+			)
+			pointerMethod = "Float64"
+		case *compile.StringSpec:
+			statements.appendf("%s, ok := req.GetQueryValue(%q)",
+				identifierName, fieldName,
+			)
+			pointerMethod = "String"
+		default:
+			panic(fmt.Sprintf("Unknown type (%T) %v", realType, realType))
```
zhenghuiwang: return an error instead of panic? Let the caller handle it.

Raynos: this case should never happen. it should be an exhaustive switch

codegen/method.go
Raynos: good idea

codegen/method.go
Raynos: The methods being called do all the logging and error handling for us.

The endpoint only has to return.

PR #132, opened 2017-06-20
Initialize clients with dependency structs

ChuntaoLu:


codegen/module_system.go
```diff
@@ -563,16 +563,27 @@ type ClientsInitGenerator struct {
...
+	for i, client := range clients {
 		// TODO: there shouldn't be a special thing for custom
 		if client.ClientType == "custom" {
 			includedPkgs = append(includedPkgs, GoPackageImport{
 				PackageName: client.CustomImportPath,
 				AliasName:   client.ImportPackageAlias,
 			})
+			if len(instances[i].ResolvedDependencies["client"]) > 0 {
+				includedPkgs = append(includedPkgs, GoPackageImport{
+					PackageName: instances[i].PackageInfo.GeneratedPackagePath,
```
Matt-Esch: You shouldn't be using the GeneratedPacakgePath, because some packages are not generated. The ImportPackagePath() method will give you the correct thing, likewise, ImportPackageAlias(). I see why you might be confused about the custom type, really there should be no special treatment of the custom type. We're just going to call a custom initializer that returns a custom type wrapper.

codegen/module_system.go
```diff
@@ -563,16 +563,27 @@ type ClientsInitGenerator struct {
...
+					AliasName:   instances[i].PackageInfo.GeneratedPackageAlias,
+				})
+			}
 			continue
```
Matt-Esch: You probably want to avoid continues over use of if/else. Flow control is generally more easily followed with if structures than through continue/return statements.

codegen/module_system.go
```diff
@@ -1098,3 +1116,46 @@ func GenerateDependencyStruct(
 		packageHelper,
 	)
 }
+
+// sort the clients by client height in DAG, lowest node first
```
Matt-Esch: How do you deal with cycles? We need to make sure we clearly surface any cycles in the DAG rather than relying on compilation errors.

codegen/module_system.go
ChuntaoLu: The package we try to include here is for the generated dependencies, isn't it always the generated package?

codegen/module_system.go
ChuntaoLu: That's good point, I will try to have the cycle detection in here

ChuntaoLu: Done

codegen/module_system.go
```diff
@@ -563,16 +563,27 @@ type ClientsInitGenerator struct {
 func (g *ClientsInitGenerator) Generate(
 	instance *ModuleInstance,
 ) (*BuildResult, error) {
-	clients := readClientDependencySpecs(instance)
+	clients := []*ClientSpec{}
+	instances := sortInstances(instance.ResolvedDependencies["client"])
+
+	for _, instance := range instances {
+		clients = append(clients, instance.GeneratedSpec().(*ClientSpec))
```
ChuntaoLu: Done

codegen/module_system.go
ChuntaoLu: Done

PR #133, opened 2017-06-20
Fix client response zero-value

ChuntaoLu:
Previously we assume the zero value of a client response is nil, this is only true for some types. This PR declares a variable of response type, and returns the variable (which has correct zero value) when error is non-nil.


codegen/templates/http_client.tmpl
```diff
@@ -129,16 +133,24 @@ func (c *{{$clientName}}) {{title .Name}}(
...
 			var responseBody {{unref .ResponseType}}
+			{{if isPointerType .ResponseType -}}
 			err = res.ReadAndUnmarshalBody(&responseBody)
+			{{else -}}
+			rawBody, err := res.ReadAll()
 			if err != nil {
-				return nil, respHeaders, err
+				return _res, respHeaders, err
+			}
+			err = json.Unmarshal(rawBody, &responseBody)
```
zhenghuiwang: line 139-146 seems to duplicate `ReadAndUnmarshalBody`.
Because`.ResponseType` is either a struct/basic type or a pointer to it. Is it simpler to write this piece of code like this:
```
var responseBody {{.ResponseType}}
{{if isPointerType .ResponseType -}}
 			err = res.ReadAndUnmarshalBody(responseBody)	
{{else -}}
             err = res.ReadAndUnmarshalBody(&responseBody)	
{{end}}
...
return responseBody, respHeader, nil
```

codegen/templates/http_client.tmpl
```diff
@@ -67,6 +68,9 @@ func (c *{{$clientName}}) {{title .Name}}(
 	r {{.RequestType}},
 ) ({{.ResponseType}}, map[string]string, error) {
 {{end}}
+	{{if .ResponseType -}}
+	var _res  {{.ResponseType}}
```
zhenghuiwang: `_XXX` is not a good name for go variable.

codegen/templates/http_client.tmpl
```diff
@@ -170,35 +182,43 @@ func (c *{{$clientName}}) {{title .Name}}(
...
 		case {{.OKStatusCode.Code}}:
 			var responseBody {{unref .ResponseType}}
+			{{if isPointerType .ResponseType -}}
 			err = res.ReadAndUnmarshalBody(&responseBody)
+			{{else -}}
+			rawBody, err := res.ReadAll()
+			if err != nil {
+				return _res, respHeaders, err
+			}
+			err = json.Unmarshal(rawBody, &responseBody)
```
Raynos: I dont like this naked `json.Unmarshal` without having error logging. Can we add a new method to `res` that's like `ReadAndUnmarshalNonPointerValue()` or something.

examples/example-gateway/idl/clients/baz/baz.thrift
```diff
@@ -39,7 +39,7 @@ service SimpleService {
 }
 
 service SecondService {
-    void Echo(
+    string Echo(
```
Raynos: Can we add a new method to http bar client that returns a string to verify this.

codegen/templates/http_client.tmpl
ChuntaoLu: `ReadAndUnmarshalBody ` expects a `json.Unmarshaler` interface, I don't think simply adding `&` will do it.

codegen/templates/http_client.tmpl
ChuntaoLu: Done

examples/example-gateway/idl/clients/baz/baz.thrift
ChuntaoLu: Done

codegen/templates/http_client.tmpl
ChuntaoLu: Done

codegen/templates/http_client.tmpl
```diff
@@ -170,35 +177,39 @@ func (c *{{$clientName}}) {{title .Name}}(
 	switch res.StatusCode {
 		case {{.OKStatusCode.Code}}:
 			var responseBody {{unref .ResponseType}}
+			{{if isPointerType .ResponseType -}}
 			err = res.ReadAndUnmarshalBody(&responseBody)
+			{{else -}}
+			err = res.ReadAndUnmarshalNonPointerValue(&responseBody)
```
zhenghuiwang: `ReadAndUnmarshalNonPointerValue` takes a pointer but its name means the opposite.

can we use `ReadAndUnmarshalNonPointerValue` to replace `ReadAndUnmarshalBody`?

codegen/templates/http_client.tmpl
```diff
@@ -170,35 +177,39 @@ func (c *{{$clientName}}) {{title .Name}}(
...
-				return nil, respHeaders, err
+				return defaultRes, respHeaders, err
 			}
 
 			{{- if .ResHeaderFields }}
 			// TODO(jakev): read response headers and put them in body
 			{{- end}}
 
-			return &responseBody, respHeaders, nil
+			return {{if isPointerType .ResponseType}}&{{end}}responseBody, respHeaders, nil
```
stevededalus: Is it better to use ` {{if isPointerType .ResponseType}}&{{end}}` or should we just modify the response type with an `&` prefix? Haven't looked through all the code, but seems like in a few of these cases that would be cleaner.

codegen/templates/http_client.tmpl
ChuntaoLu: Changed `ReadAndUnmarshalNonPointerValue ` to `ReadAndUnmarshalNonStructBody`, which I think reads better.

codegen/templates/http_client.tmpl
ChuntaoLu: > or should we just modify the response type with an `&` prefix?

I don't quite get the question, but I will try to clear the confusion: `.ResponseType` is a full type, meaning if it's a pointer type, it has the `*` prefix. `responseBody` has the type of `unref .ResponseType`, `unref` removes the `*` prefix if it is present, which is necessary so that we can pass `&responseBody` to the unmarshal function. And `{{if isPointerType .ResponseType}}&{{end}}` is the opposite of `unref`, we can make it a template function, but I found it not so necessary.

stevededalus: Sorry, what I meant was can't you set `responseBody` to either `responseBody` or `&responseBody` once based on that check either in the generator or the top of the template and remove all the in line checks? Not a big deal either way.

PR #136, opened 2017-06-26
Fix tchannel client method visibility 

stevededalus:
tchannel client methods are name as `methodName` and thus package level visible. They should be `MethodName` so the endpoint code can call them

### DIFF
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -63,7 +63,7 @@ type {{$clientName}} struct {
 {{$methodName := index $exposedMethods $serviceMethod -}}
 {{if $methodName -}}
 	// {{$methodName}} is a client RPC call for method "{{$svc.Name}}::{{.Name}}"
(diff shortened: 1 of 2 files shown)


codegen/templates/tchannel_client.tmpl
```diff
@@ -62,8 +62,8 @@ type {{$clientName}} struct {
 {{$serviceMethod := printf "%s::%s" $svc.Name .Name -}}
 {{$methodName := index $exposedMethods $serviceMethod -}}
 {{if $methodName -}}
-	// {{$methodName}} is a client RPC call for method "{{$svc.Name}}::{{.Name}}"
-	func (c *{{$clientName}}) {{$methodName}}(
+	// {{title $methodName}} is a client RPC call for method "{{$svc.Name}}::{{.Name}}"
```
zhenghuiwang: can we add `title` to line 63 instead?

stevededalus: done.

PR #141, opened 2017-06-28
Add package `repository` to manage gateway configurations.

zhenghuiwang:
Start package `repository` which read/write JSON configs & thrift files for a gateway. It can be later used as UI backend to easily manage gateway.

r: @uber/zanzibar-team


Raynos:
It would be nice to minimize top level folders. We can move `repository` into `backend/` or `backend/repository` .

repository/repository_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/uber/zanzibar/mock/mock_repository"
+)
+
+func TestNewRepository(t *testing.T) {
+	root := os.TempDir()
+	localDir := filepath.Join(root, "repo")
+	remote := "gitolite@domain/repo"
+	fetcher := mock_repository.NewMockFetcher(gomock.NewController(t))
```
Raynos: I don't like this mocking approach to testing. Can we add tests backed by git itself ?

mock/mock_repository/mock_fetcher.go
```diff
@@ -0,0 +1,62 @@
+// Automatically generated by MockGen. DO NOT EDIT!
```
Raynos: Would prefer to avoid using auto generated mocks in the codebase. it encourages this approach to testing when we really need to use integration tests.

repository/repository_test.go
ChuntaoLu: What's the rationale against testing with mocks?

Raynos: This is a platform library which works mostly with IO. It's very hard to know if we broke something without having integration tests.

For the repository and the UI backend we want to verify if zanzibar breaks the generated code via the UI backend. The best way to test that is to get the UI backend to generate real code ( including git interactions ) and to verify that code compiles & passes lint. 

We should favor writing integration tests first and unit tests second.

PR #144, opened 2017-06-29
Generate clients as interfaces instead of structs

ChuntaoLu:
This PR generates the http/tchannel clients as client interfaces instead of structs in order to easily mock clients for testing.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -503,7 +503,7 @@ func readPackageInfo(
 		),
 		ExportName:            "New" + qualifiedClassName,
 		QualifiedInstanceName: qualifiedInstanceName,
-		ExportType:            qualifiedInstanceName + qualifiedClassName,
+		ExportType:            qualifiedClassName,
 		GeneratedPackagePath: filepath.Join(
 			packageRoot,
 			relativeGeneratedPath,
--- a/codegen/templates/dependency_struct.tmpl
+++ b/codegen/templates/dependency_struct.tmpl
@@ -20,6 +20,7 @@ type Dependencies struct {
 // {{$classType | pascal}}Dependencies contains {{$classType}} dependencies
 type {{$classType | pascal}}Dependencies struct {
 	{{ range $idx, $dependency := $moduleInstances -}}
+	{{- /* TODO: the dependency type should cover all types instead of just interface type */ -}}
 	{{$dependency.PackageInfo.QualifiedInstanceName}} {{$dependency.PackageInfo.ImportPackageAlias}}.{{$dependency.PackageInfo.ExportType}}
 	{{end -}}
 }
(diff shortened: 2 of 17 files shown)


ChuntaoLu:
@Raynos here is another example where Travis is not triggered.

codegen/module.go
```diff
@@ -503,7 +503,7 @@ func readPackageInfo(
 		),
 		ExportName:            "New" + qualifiedClassName,
 		QualifiedInstanceName: qualifiedInstanceName,
-		ExportType:            qualifiedInstanceName + qualifiedClassName,
+		ExportType:            qualifiedClassName,
```
Raynos: Why change this part ?

codegen/templates/dependency_struct.tmpl
```diff
@@ -20,7 +20,7 @@ type Dependencies struct {
 // {{$classType | pascal}}Dependencies contains {{$classType}} dependencies
 type {{$classType | pascal}}Dependencies struct {
 	{{ range $idx, $dependency := $moduleInstances -}}
-	{{$dependency.PackageInfo.QualifiedInstanceName}} *{{$dependency.PackageInfo.ImportPackageAlias}}.{{$dependency.PackageInfo.ExportType}}
+	{{$dependency.PackageInfo.QualifiedInstanceName}} {{$dependency.PackageInfo.ImportPackageAlias}}.{{$dependency.PackageInfo.ExportType}}
```
Raynos: not all dependencies are going to be interfaces. We will need something smarter for this.

codegen/templates/http_client.tmpl
```diff
@@ -17,30 +17,55 @@ import (
...
 	baseURL := "http://" + ip + ":" + strconv.Itoa(int(port))
 	return &{{$clientName}}{
-		ClientID: "{{$clientID}}",
-		HTTPClient: zanzibar.NewHTTPClient(gateway, baseURL),
+		clientID: "{{$clientID}}",
+		httpClient: zanzibar.NewHTTPClient(gateway, baseURL),
 	}
 }
 
+// {{$clientName}} is the http client.
```
Raynos: nitpick: move the struct above the newX() func ?

codegen/module.go
ChuntaoLu: The instance name is already (or should be) in the importing package name. Compare `bar.Client` with `bar.BarClient`, I think the former is cleaner and reads better, and linting for avoiding stutter also approves the former.

codegen/templates/dependency_struct.tmpl
ChuntaoLu: Good point, I will drop a TODO here

codegen/templates/http_client.tmpl
ChuntaoLu: Done

Issue #137, opened 2017-06-26
typedef in ThriftSpec not handled correctly in zanzibar


stevededalus:
typedef Thrift aliases are treated as custom types in the compiled.Field golang structs instead of as at the underlying type. This breaks code generation.

A common pattern is alias UUID as strings and returning lists of UUIDs and maps keyed by UUIDs. Code generation in zanzibar breaks because it does not understand the UUID fields are string types.

Raynos:
Another example of this is strings backed by query parameters or headers when using a UUID typedef. We do not generate code that casts the string to the typedef correctly.

Issue #134, opened 2017-06-22
Method request/response type does not cover thrift basic and container types


ChuntaoLu:
The `SetResponseType` and `SetRequestType` methods on `MethodSpec` assumes the Go type is the same as the thrift type, which is only correct for struct type and a couple of basic thrift types. The example thrifts don't cover all thrift types, a thorough thrift with all thrift types is needed to justify the implementation.

PR #146, opened 2017-07-01
Generate correct Go type from client response thrift type

ChuntaoLu:
This PR makes sure that all thrift types are covered for client response.

This PR addresses issue https://github.com/uber/zanzibar/issues/134 https://github.com/uber/zanzibar/issues/137


codegen/package.go
```diff
@@ -197,15 +198,114 @@ func (p PackageHelper) TypeFullName(typeSpec compile.TypeSpec) (string, error) {
...
+func isStructType(spec compile.TypeSpec) bool {
+	spec = compile.RootTypeSpec(spec)
+	_, isStruct := spec.(*compile.StructSpec)
+	return isStruct
+}
+
+// isHashable returns true if the given type is considered hashable by thriftrw.
+//
+// Only primitive types, enums, and typedefs of other hashable types are considered hashable.
+func isHashable(t compile.TypeSpec) bool {
```
zhenghuiwang: The func body says isHashable is the same as is PrimitiveType. Why keep two functions instead of only `isPrimitiveType`?

runtime/client_http_response.go
```diff
@@ -86,39 +86,8 @@ func (res *ClientHTTPResponse) ReadAll() ([]byte, error) {
...
-		return err
-	}
-
-	return res.UnmarshalBody(body, rawBody)
-}
-
-// ReadAndUnmarshalNonStructBody will try to unmarshal non pointer value or fail
-func (res *ClientHTTPResponse) ReadAndUnmarshalNonStructBody(v interface{}) error {
+// ReadAndUnmarshalBody will try to unmarshal non pointer value or fail
+func (res *ClientHTTPResponse) ReadAndUnmarshalBody(v interface{}) error {
```
zhenghuiwang: Great.  These two methods are finally merged :)

examples/example-gateway/idl/clients/baz/baz.thrift
```diff
@@ -45,4 +45,24 @@ service SecondService {
...
+
+    list<base.UUID> EchoUUIDList(
+        1: required list<base.UUID> arg
+    )
+
+    set<string> EchoSet(
+        1: required set<string> arg
+    )
+
+    map<base.UUID, base.BazResponse> EchoMap(
```
zhenghuiwang: Add a test for `map<>` with incomparable map key type?  like `map<base.BazRespoonse, bool>`?

codegen/package.go
ChuntaoLu: This function is primarily for readability, it makes the checks in `goType` function obvious.

examples/example-gateway/idl/clients/baz/baz.thrift
ChuntaoLu: Not sure how that will work, won't we get an error at `make generate` step?

zhenghuiwang: The example I give may not be accurate. I mean you may want to test `map<SOME_STRUCT, bool>` in thrift while SOME_STRUCT is not a comparable type.

codegen/package.go
zhenghuiwang: This violates DRY principle. A comment on the caller is sufficient.

ChuntaoLu: Done

examples/example-gateway/idl/clients/baz/baz.thrift
ChuntaoLu: Done, added unit tests for such cases.

codegen/package.go
```diff
@@ -197,15 +198,112 @@ func (p PackageHelper) TypeFullName(typeSpec compile.TypeSpec) (string, error) {
 	if typeSpec == nil {
 		return "", nil
 	}
-	tfile := typeSpec.ThriftFile()
+	return goType(p, typeSpec)
+}
+
+// goType returns the Go type string representation for the given thrift type.
+func goType(p PackageHelper, spec compile.TypeSpec) (string, error) {
```
Raynos: This function is duplicate with https://github.com/uber/zanzibar/blob/master/codegen/type_converter.go#L68

Raynos: Maybe have a new `thrift.go` file for these types of functions ?

examples/example-gateway/idl/clients/baz/baz.thrift
```diff
@@ -42,7 +49,59 @@ service SimpleService {
 }
 
 service SecondService {
-    string Echo(
+    i8 EchoI8(
```
Raynos: should we write echo tests that all values can go in & out ?

codegen/package.go
ChuntaoLu: Done, extracted type conversions into a separate file and wrote tests for it.

examples/example-gateway/idl/clients/baz/baz.thrift
ChuntaoLu: Done, wrote tests for both http and tchannel clients.

scripts/cover.sh
```diff
@@ -41,16 +41,25 @@ for file in "${FILES_ARR[@]}"; do
...
 	COVERNAME="./coverage/cover-unit-$RAND.out"
 
-	COVER_ON=1 ZANZIBAR_CACHE=1 go test \
-		-cover -coverprofile coverage.tmp $file 2>&1 | \
-		tee test.tmp.out >>test.out && \
-		mv coverage.tmp "$COVERNAME" 2>/dev/null || true
+	relativeName=$(echo $file | sed s#github.com/uber/zanzibar#.#)
+
+    # TODO: need better solution for coverage from different package
+    if [[ "$relativeName" == *"test/clients"* ]] || [[ "$relativeName" == *"test/endpoints"* ]]; then
```
Raynos: endpoints use the child process test gateway so they do not have this problem.

scripts/cover.sh
```diff
@@ -41,16 +41,25 @@ for file in "${FILES_ARR[@]}"; do
...
+        coverpkg=$(echo $file | sed s#zanzibar/test/#zanzibar/examples/example-gateway/build/#)
+        COVER_ON=1 ZANZIBAR_CACHE=1 go test \
```
Raynos: This approach for clients is good enough but probably will not work long term.

PR #150, opened 2017-07-06
Update readme regarding updating golden files

ChuntaoLu:


Raynos:
codegen still requires -update. Maybe we should just do `go test ./codegen ./backend/...`

ChuntaoLu:
@Raynos that's right, the command in the pr is just an example as noted by `e.g.`

PR #152, opened 2017-07-07
Make `repository` be able to update its client configs.

zhenghuiwang:
This PR makes a `repository` be able to update its own client configurations:

- Change the `ClientConfig` definition, which defines the request for updating a client (also the response for querying a client configs).

- Client update logic in `Repository.UpdateClientConfigs`

- On testing setup: 1. Copy `example-gateway` to a local repository. 2. Update a client. 3. Compare the configuration with hand-written one in `example-gateway`.

r: @uber/zanzibar-team


backend/repository/client_update.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return cfgJSON
+}
+
+// UpdateClientConfigs updates JSON configuration files for a client update request.
+func (r *Repository) UpdateClientConfigs(req *ClientConfig, clientCfgDir, thriftFileSha string) error {
+	if err := validateClientUpdateRequest(req); err != nil {
+		return err
+	}
+	cfgJSON := NewClientConfigJSON(*req)
```
ChuntaoLu: why not `NewClientConfigJSON` take a pointer as well?

backend/repository/client_update.go
```diff
@@ -0,0 +1,179 @@
...
+	if req.Type == "tchannel" && req.Timeout == 0 {
+		req.Timeout = 10000
+	}
+	if req.Type == "tchannel" && req.TimeoutPerAttempt == 0 {
+		req.TimeoutPerAttempt = 10000
+	}
+	return nil
+}
+
+type clientModuleJSONConfig struct {
```
ChuntaoLu: overlap with https://github.com/uber/zanzibar/blob/master/codegen/gateway.go#L112, we should establish single SOT to ensure consistency.

backend/repository/config.go
```diff
@@ -78,10 +78,15 @@ type EndptMidConfig struct {
...
 // ClientConfig stores configuration for an client.
 type ClientConfig struct {
-	ID             string
-	Type           ProtocolType
-	ThriftFile     string
-	ExposedMethods map[string]string
+	Name              string            `json:"name"`
+	Type              ProtocolType      `json:"type"`
+	ThriftFile        string            `json:"thriftFile,omitempty"`
+	MuttleyName       string            `json:"muttleyName,omitempty"`
```
ChuntaoLu: This name is not straightforward without certain context.

backend/repository/config.go
```diff
@@ -117,8 +122,13 @@ type ThriftMeta struct {
 type ProtocolType string
 
 const (
-	gatewayConfigFile = "gateway.json"
+	gatewayConfigFile     = "gateway.json"
+	productionCfgJSONPath = "config/production.json"
+	clientConfigFileName  = "client-config.json"
+	clientModuleFileName  = "clients-config.json"
```
ChuntaoLu: `clientsConfigFileName` is arguably a better name from module system point of view, despite it's awfully close to `clientConifgFileName`

backend/repository/data/client/baz_client_update.json
```diff
@@ -0,0 +1,33 @@
+{
+    "_comment_": "JSON for updating a tchannel client.",
```
ChuntaoLu: The JSON schema way is to name this field "description", http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.2

backend/repository/client_update.go
zhenghuiwang: will change it.

backend/repository/client_update.go
zhenghuiwang: You're right. ( wired: /codegen/gateway.go@`master`#L112 doesn't have `dependencies`. `dependencies` in the hand-written config seems not used at all!?)

backend/repository/config.go
zhenghuiwang: `ServiceName` is overload for http. what could be a better one? `TchannelServiceName`, `TchannelRoutingName`?

backend/repository/config.go
zhenghuiwang: that's why i use `clientModule`, meaning for all clients but also distinctive?

backend/repository/data/client/baz_client_update.json
zhenghuiwang: Thought about it: JSON schema reserves string "description", which could be a key and is not good for the general purpose of a JSON. `_description_` or `_comment_` `__description__` or `__comment__` all seem good to me (they should be the first one lexicographically).

backend/repository/client_update.go
ChuntaoLu: Dependencies are took care of at module level, we can add it to the clientConfig struct

backend/repository/config.go
ChuntaoLu: Not sure what you mean by overload. How about we use `ServiceName` to denote service discovery/routing name, and `ThriftServiceName` for service defined in thrift file?

backend/repository/config.go
ChuntaoLu: maybe `clientsModule`?

backend/repository/data/client/baz_client_update.json
ChuntaoLu: makes sense

Issue #131, opened 2017-06-19
Overlap between ExportType and CustomClientType for custom client


ChuntaoLu:
For custom clients, the `CustomClientType` and `ExportType` in client spec struct seem to mean the same thing, the client's type when it's referred in other packages. Having two different fields for the same thing is bad, and the worse part is the discrepancy between how the field is set: `CustomClientType` is read from client config  while `ExportType` is always set as `{Name}Client`.

PR #153, opened 2017-07-07
Remove unused fileds in custom client config

ChuntaoLu:
This PR addresses issue https://github.com/uber/zanzibar/issues/131

### DIFF
--- a/codegen/gateway.go
+++ b/codegen/gateway.go
@@ -45,8 +45,6 @@ var mandatoryClientFields = []string{
 }
 var mandatoryCustomClientFields = []string{
 	"customImportPath",
-	"customClientType",
-	"customPackageName",
 }
 var mandatoryEndpointFields = []string{
 	"endpointType",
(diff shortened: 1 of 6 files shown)


codegen/gateway.go
```diff
@@ -125,7 +119,7 @@ func NewClientSpec(
 	if err := json.Unmarshal(instance.JSONFileRaw, &clientConfig); err != nil {
 		return nil, errors.Wrapf(
 			err,
-			"Could not parse class config json file %s: ",
+			"Could not parse class config json file: %s",
```
zhenghuiwang: Nit: better use `%q` instead of `: %s`, because the `:` is a separator for wrapped errors.

ChuntaoLu: I prefer `%q` most of the time to make the interesting part stand out, but here the json file path will be long, and wrapping it in quotes doesn't serve that purpose well.

PR #155, opened 2017-07-10
Update logging to assist in debugging downstream integration issues

stevededalus:
- record downstream tchannel request 
- record downstream tchannel response 
- record tchannel headers 
- record incoming http request body


codegen/templates/tchannel_client.tmpl
```diff
@@ -70,12 +74,18 @@ func {{$exportName}}(gateway *zanzibar.Gateway) Client {
 
 	return &{{$clientName}}{
 		client: client,
+		{{if $logDownstream -}}
```
Raynos: we should always pass the logger here.

codegen/templates/tchannel_client.tmpl
```diff
@@ -70,12 +74,18 @@ func {{$exportName}}(gateway *zanzibar.Gateway) Client {
...
+		logger: gateway.Logger,
+		{{end -}}
 	}
 }
 
 // {{$clientName}} is the TChannel client for downstream service.
 type {{$clientName}} struct {
 	client        zanzibar.TChannelClient
+	{{if $logDownstream -}}
```
Raynos: We should use `.client.Logger` instead.

codegen/templates/tchannel_client.tmpl
```diff
@@ -99,10 +109,33 @@ type {{$clientName}} struct {
...
+
+		{{if $logDownstream -}}
+			var fields []zapcore.Field
+			fields = append(fields, zap.String("Downstream-Client", "{{$clientName}}"))
+			fields = append(fields, zap.String("Downstream-Method", "{{$methodName}}"))
+			fields = append(fields, zap.Time("timestamp", time.Now().UTC()))
+			for k, v := range reqHeaders {
+				fields = append(fields, zap.String("Downstream-Request-Header-"+k, v))
+			}
+			fields = append(fields, zap.String("Downstream-Request-Body", args.String()))
```
Raynos: Instead of `String()` do we want to do json encoding ? `zap.Any(key, args)` ?

codegen/templates/tchannel_client.tmpl
```diff
@@ -99,10 +109,33 @@ type {{$clientName}} struct {
...
+
 		success, respHeaders, err := c.client.Call(
 			ctx, "{{$svc.Name}}", "{{.Name}}", reqHeaders, args, &result,
 		)
 
+		{{if $logDownstream -}}
+			for k, v := range respHeaders {
+				fields = append(fields, zap.String("Downstream-Response-Header-"+k, v))
+			}
+			fields = append(fields, zap.String("Downstream-Response-Body", result.String()))
```
Raynos: again maybe `zap.Any(str, result)` instead.

runtime/router.go
```diff
@@ -295,14 +297,16 @@ func logRequestFields(r *http.Request) []zapcore.Field {
...
 	// TODO: Add a gateway level configurable body unmarshaller
 	// to extract only non-PII info.
 
-	// body, err := ioutil.ReadAll(r.Body)
-	// r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
-	// fields = append(fields, zap.String("body", string(body)))
-
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		fields = append(fields, zap.String("Request-Body", string(body)))
```
Raynos: `zap.Any(str, body)`

codegen/templates/tchannel_client.tmpl
Raynos: actually logger is already on `client.Logger` so need to pass.

codegen/templates/tchannel_client.tmpl
stevededalus: zanzibar tchannel_interface doesn't have a logger.

codegen/templates/tchannel_client.tmpl
stevededalus: Done

test/endpoints/googlenow/googlenow_test.go
```diff
@@ -160,75 +160,6 @@ func (conn *failingConn) Write(b []byte) (int, error) {
 	return n, err
 }
 
-func TestGoogleNowFailReadAllCall(t *testing.T) {
```
Raynos: This test is not old, we need it for code coverage.

runtime/router.go
```diff
@@ -295,14 +297,16 @@ func logRequestFields(r *http.Request) []zapcore.Field {
...
+	// TODO: Do not log body by default because PII and bandwidth.
+	// Temporarily log during the developement cycle
 	// TODO: Add a gateway level configurable body unmarshaller
 	// to extract only non-PII info.
 
-	// body, err := ioutil.ReadAll(r.Body)
-	// r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
-	// fields = append(fields, zap.String("body", string(body)))
-
+	body, err := ioutil.ReadAll(r.Body)
```
Raynos: we are swallowing an err here.

We need to create a new reader that remembers this error when we re-assign `r.Body`

runtime/router.go
```diff
@@ -295,14 +297,16 @@ func logRequestFields(r *http.Request) []zapcore.Field {
...
+	if err != nil {
+		fields = append(fields, zap.Any("Request-Body", body))
+	}
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
```
Raynos: We should log the response body in `logResponseFields`

runtime/router.go
Raynos: instead of consuming the body here we should leave the body as is, touching it is a side-effect.

We can store the raw incoming body on `ServerHTTPRequest` and then read that in `logResponseFields` to log the raw incoming body.

Raynos: Actually, we shouldn't create a new reader here. this mutation of the request seems backwards.

stevededalus: Sounds good, went with that approach

PR #156, opened 2017-07-10
Setup tchannel stats reporter

ChuntaoLu:


runtime/gateway.go
```diff
@@ -440,20 +440,20 @@ func (gateway *Gateway) setupTChannel(config *StaticConfig) error {
 	serviceName := config.MustGetString("tchannel.serviceName")
 	processName := config.MustGetString("tchannel.processName")
 
+	subScope := gateway.MetricScope.SubScope("tchannel")
```
Raynos: I suspect tchannel already prefixes its own metrics with tchannel.

Raynos: Can we write a test for http endpoint + tchannel client and write assertions on what metrics are emitted ?

We should also re-run the benchmarks to see what the performance implications are and make some flamegraphs to double check if there are any low hanging fruit.

runtime/tchannel_metrics.go
```diff
@@ -0,0 +1,71 @@
...
+// key that is used when reporting to M3.
+type MetricKeyFunc func(name string, tags map[string]string) string
+
+// NewDefaultTChannelStatsReporter returns a StatsReporter that reports stats
+// using the given tally.Scope. It only reports a metric key that is generated
+// from the given list of tags.
+func NewDefaultTChannelStatsReporter(s tally.Scope) tchannel.StatsReporter {
+	return &statsReporter{
+		scope: s,
+		metricKey: func(name string, tags map[string]string) string {
```
Raynos: we do not use statsd, only m3 so we should not need this function.

runtime/tchannel_metrics.go
```diff
@@ -0,0 +1,71 @@
...
+
+// NewTChannelStatsReporter returns a StatsReporter that reports stats
+// using the given tally.Scope, using the given metric key function.
+func NewTChannelStatsReporter(s tally.Scope, m MetricKeyFunc) tchannel.StatsReporter {
+	return &statsReporter{s, m}
+}
+
+// IncCounter ...
+func (s statsReporter) IncCounter(name string, tags map[string]string, value int64) {
+	s.scope.Counter(s.metricKey(name, tags)).Inc(value)
```
Raynos: We do not want to convert tags into the metric key.

`s.scope.Tagged(tags).Counter(name).Inc(value)` instead.

Raynos: We really are going to need some form of caching for this :( allocating a new sub scope and counter for every request is going to be expensive.

This will either require a patch to tchannel-go upstream or extensive benchmarking and tuning.

runtime/tchannel_metrics.go
ChuntaoLu: makes sense, will get rid of it

runtime/tchannel_metrics.go
ChuntaoLu: the same set of tags will resolve to the same subscope, tally itself caches subscopes/counters/gauges/timers

runtime/gateway.go
ChuntaoLu: I will double check with the prefix, and work on integration tests and perf.

runtime/tchannel_metrics.go
ChuntaoLu: Done.

runtime/gateway.go
ChuntaoLu: Verified that tchannel metrics doesn't come with the "tchannel" prefix, also added integration tests.

ChuntaoLu:
@Raynos below are the benchmark results, seems like the tchannel metrics add about 2% latency overhead on average, which is pretty small.
- 16B payload on master:
```
➜  zanzibar git:(master) ✗ wrk -t12 -c400 -d30s -s ./benchmarks/baz_call_16B.lua http://localhost:8093/baz/call
Running 30s test @ http://localhost:8093/baz/call
    Latency    24.93ms   14.97ms 154.57ms   68.41%
    Req/Sec     1.33k   224.11     2.63k    72.58%
  478695 requests in 30.05s, 52.50MB read
  Socket errors: connect 0, read 225, write 0, timeout 0
Requests/sec:  15927.53
Transfer/sec:      1.75MB
```
- 16B payload on PR branch:
```
➜  zanzibar git:(lu.metrics) wrk -t12 -c400 -d30s -s ./benchmarks/baz_call_16B.lua http://localhost:8093/baz/call
Running 30s test @ http://localhost:8093/baz/call
    Latency    25.32ms   14.81ms 152.08ms   68.56%
    Req/Sec     1.32k   184.55     2.29k    73.17%
  472162 requests in 30.06s, 51.78MB read
  Socket errors: connect 0, read 201, write 0, timeout 0
Requests/sec:  15706.42
Transfer/sec:      1.72MB
```

- 4KB payload on master: 
```
➜  zanzibar git:(master) wrk -t12 -c400 -d30s -s ./benchmarks/baz_call_4KB.lua http://localhost:8093/baz/call
Running 30s test @ http://localhost:8093/baz/call
    Latency    28.09ms   16.40ms 141.10ms   68.05%
    Req/Sec     1.18k   177.21     2.15k    74.56%
  424751 requests in 30.05s, 46.58MB read
  Socket errors: connect 0, read 207, write 0, timeout 0
Requests/sec:  14135.21
Transfer/sec:      1.55MB
```
- 4KB payload on PR branch:
```
➜  zanzibar git:(lu.metrics) wrk -t12 -c400 -d30s -s ./benchmarks/baz_call_4KB.lua http://localhost:8093/baz/call
Running 30s test @ http://localhost:8093/baz/call
    Latency    28.88ms   17.19ms 156.88ms   67.44%
    Req/Sec     1.16k   179.58     2.18k    72.36%
  414965 requests in 30.06s, 45.51MB read
  Socket errors: connect 0, read 240, write 0, timeout 0
Requests/sec:  13806.42
Transfer/sec:      1.51MB
```

ChuntaoLu:
@Raynos searching "tally" in the flamegraphs shows that we do have more cycles spent on processing the metrics.
 
screen shot 2017-07-11 at 3 28 40 pm
screen shot 2017-07-11 at 3 27 58 pm

Makefile
```diff
@@ -152,10 +152,12 @@ fast-bench:
 bench:
 	time -p sh -c "go test -run _NONE_ -bench . -benchmem -benchtime 7s -cpu 2 ./test/... | grep -v '^ok ' | grep -v '\[no test files\]' | grep -v '^PASS'"
 
+.PHONY: $(PROGS)
```
Raynos: why are these phony ?

ChuntaoLu: I marked them as `.PHONY` with the same reason as the `$(EXAMPLE_SERVICES)` target, should they not be?

Raynos: i dont think these are PHONY they are actual files that exist.

Makefile
```diff
@@ -156,6 +156,10 @@ $(PROGS): $(GO_FILES)
 	@echo Building $@
 	go build -o $@ $(dir ./$@)
 
+# These dirs are generated by `make generate`, if `make generate` is ran after any $(GO_FILES) is modified,
```
ChuntaoLu: @Raynos fyi

PR #157, opened 2017-07-10
Setup tracer for tchannel

ChuntaoLu:


runtime/gateway.go
```diff
@@ -413,6 +424,25 @@ func (gateway *Gateway) setupLogger(config *StaticConfig) error {
...
 }
 
+func (gateway *Gateway) setupTracer(config *StaticConfig) error {
+	opts := []jaegerConfig.Option{
+		// TChannel logger implements jaeger logger interface
+		jaegerConfig.Logger(NewTChannelLogger(gateway.Logger)),
+		jaegerConfig.Metrics(NewJaegerMetricsFactory(gateway.MetricScope)),
+	}
+	// TODO: using default tracing configurations here, we may want to set it from static config
+	jc := jaegerConfig.Configuration{}
```
Raynos: In our test suite we should set the sampler to always sample.

test/endpoints/baz/baz_metrics_test.go
```diff
@@ -146,7 +147,77 @@ func TestCallMetrics(t *testing.T) {
 		assert.Equal(t, expectedEndpoitTags[tag.GetTagName()], tag.GetTagValue())
 	}
 
-	tchannelOutboundSuccessMetric := metrics[3]
+	expectedJaegerSpanMetricName := "test-gateway.production.all-workers.jaeger.spans"
```
Raynos: It would be really good to be able to write tests on the spans emitted themselves instead of just the metrics jaeger emits.

Raynos: We can use `MockAgent` which will run a fake agent on a UDP port to be able to write end-to-end tests.

https://github.com/uber/jaeger-client-go/blob/master/testutils/mock_agent.go#L43-L65

It has a `agent.SpanServerAddr()` method so we can configure the address of the reporter with `ReporterConfig.LocalAgentHostPort`

Raynos: https://github.com/uber/jaeger-client-go/blob/master/config/config.go#L75

^ can set the host port which defaults to `"localhost:6831"` which should just put the host port in the `config/production.json` and overwrite it in the TestGateway / BenchGateway.

Raynos: We setup a metrics backend in `TestGateway` ( https://github.com/uber/zanzibar/blob/master/test/lib/test_gateway/test_gateway.go#L185 ). This is a good place to setup a jaeger reporter backend ( using `MockAgent` ).

runtime/gateway.go
```diff
@@ -413,6 +424,25 @@ func (gateway *Gateway) setupLogger(config *StaticConfig) error {
...
+	serviceName := config.MustGetString("serviceName")
+	closer, err := jc.InitGlobalTracer(serviceName, opts...)
```
ChuntaoLu: We can, but this is actually the [recommended](https://godoc.org/github.com/uber/jaeger-client-go/config#example-Configuration-InitGlobalTracer-Production) way to initialize the tracer per jaeger. The benefit is that the tracer ownership is managed by the opentracing package, so one can start a span without the need to figure out where to get the tracer.

runtime/jaeger_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/uber-go/tally"
+	jaeger_metrics "github.com/uber/jaeger-lib/metrics"
+)
+
+type jaegerGauge struct {
+	gauge tally.Gauge
+}
+
+func (j *jaegerGauge) Update(value int64) {
```
ChuntaoLu: Good to know. It seems that jaeger-lib creates a wrapper struct each time we call Counter/Gauge/Timer, and those wrapper structs are not cached as they are in tally. The only thing that really needs a wrapper struct is Gauge as the interface is different, and I am caching it here. But I agree that we don't need reimplement this and those wrapper structs are probably not a big deal. Will reuse jaeger-lib.

runtime/gateway.go
Raynos: It's best to pass dependencies explicitly. It makes understanding the code easier and reduces complexity in writing tests.

Relying on singletons can create code that is near impossible to refactor without breaking everything :(

runtime/jaeger_metrics.go
Raynos: Ok, cool we can contribute back any optimizations we make to jaeger-lib.

ChuntaoLu:
Closing as dup of https://github.com/uber/zanzibar/pull/193

PR #162, opened 2017-07-14
Initialize dependencies per service

Matt-Esch:
Putting this up for review. It's not complete yet. The intention is to initialize dependencies per service binary to ensure that the binaries only contain the clients/endpoints they need. The test/bench code needs to be fixed to complete this PR.


codegen/templates/dependency_struct.tmpl
```diff
@@ -1,5 +1,5 @@
 {{$instance := . -}}
-package {{$instance.PackageInfo.PackageName}}
+package module
```
Raynos: since you fixed everything else to use `$instance.PackageInfo.PackageName` can this use `PackageInfo` again ?

codegen/templates/main.tmpl
```diff
@@ -9,20 +10,20 @@ import (
...
 	_, file, _, _ := runtime.Caller(0)
 	return zanzibar.GetDirnameFromRuntimeCaller(file)
 }
 
+// TODO: remove this
 func getConfigDirName() string {
 	return filepath.Join(
 		getDirName(),
-		"{{.RelativePathToAppConfig}}",
+		"../../../",
```
Raynos: hard coding this seems bad. It assumes `targetGenDir` is always `./build/`

codegen/templates/endpoint.tmpl
```diff
@@ -34,22 +35,56 @@ import (
...
 
 // {{$handlerName}} is the handler for "{{.HTTPPath}}"
 type {{$handlerName}} struct {
-	Clients *clients.Clients
+	Clients *module.ClientDependencies
 }
 
-// New{{title .Name}}Endpoint creates a handler
-func New{{title .Name}}Endpoint(
+// New{{$handlerName}}} creates a handler
```
Raynos: typo, triple `}}}`

runtime/gateway.go
```diff
@@ -139,10 +135,7 @@ func CreateGateway(
 type RegisterFn func(gateway *Gateway)
```
Raynos: RegisterFn is now unused.

runtime/gateway.go
```diff
@@ -131,6 +125,8 @@ func CreateGateway(
 		return nil, err
 	}
 
+	gateway.registerPredefined()
```
Raynos: does this function return an err ?

Matt-Esch:
Getting this ready for a full review

runtime/gateway.go
Matt-Esch: If it did the linter would catch it (I've seen it catch other function calls that return error but are not assigned/handled). This line was moved.

runtime/gateway.go
Matt-Esch: Removed

codegen/templates/endpoint.tmpl
Matt-Esch: Fixed

codegen/templates/main.tmpl
Matt-Esch: Yes it seems bad, but it was hard coded inside the execution of the template without configuration anyway. I'm punting on this entire form of config being removed because it doesn't work when we purge source and run a binary in its own directory.

codegen/templates/dependency_struct.tmpl
Matt-Esch: It is solving the cycle issue right now, so I'm keeping it in the module subpackage

test/lib/test_gateway/test_gateway_cover.go
```diff
@@ -166,6 +166,8 @@ func createTestBinaryFile(
 
 	mainTestPath := strings.Replace(mainPath, "main.go", "main_test.go", -1)
 
+	fmt.Println(mainTestPath)
```
Raynos: debug println; please remove.

Matt-Esch: done

codegen/module.go
```diff
@@ -218,6 +235,145 @@ func appendUniqueModule(
...
+		insertIndex := i
+		for j := 0; j < i; j++ {
+			if insertIndex == i {
+				if peerDepends(sorted[j], instance) {
+					insertIndex = j
+				}
+			}
+
+			if insertIndex != i {
+				if peerDepends(instance, sorted[j]) {
```
ChuntaoLu: I don't think this will prevent an indirect dependency cycle like `a->b->c->a`.

codegen/module.go
```diff
@@ -218,6 +235,145 @@ func appendUniqueModule(
...
+
+func (s sortableDependencyList) Less(i int, j int) bool {
+	return s[i].InstanceName < s[j].InstanceName
+}
+
+func (s sortableDependencyList) Swap(i int, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func sortDependencyList(
```
ChuntaoLu: This function is O(n^2), I guess it's probably fine as long as we have sparse dependencies.

ChuntaoLu: Can we have some tests coverage on this function?

codegen/module.go
Matt-Esch: It will be prevented as long as the peerDepends function looks at the recursively resolved dependencies (i.e. that is all dependencies in the dependency tree of the module) and that I am computing the recursive dependencies accurately (which might not be the case right now). Will add some tests for sure.

codegen/module.go
Matt-Esch: Yes. It's worst case n^2, best case n, and n memory. Given that there is a weak partial order ( A can come before B  or B can come before A if they do not depend on each other) and given the use case (we consider peer dependencies to not be the norm), it's generally speaking going to be closer to n comparisons than n^2

codegen/module.go
Matt-Esch: I added a test specifically for the dependency cycle

ChuntaoLu: I see what it's doing now. In this case, I think it is better if we error early when the dependencies are resolved recursively, i.e., if the `instance` itself is in any of its `classDeps` https://github.com/uber/zanzibar/pull/162/files#diff-c367c203be38c31c4c65e2a02e2470f8R287, then we have a cycle.

Matt-Esch: I think it makes more sense to enforce it at sort time as it gives you a better sense for where the cycle actually occurs, and sorting is necessary either way. If instance A depends on A, the error message cannot direct you at the next dependency in the cycle, whereas right now you're told A cannot be initialized before B.

codegen/module.go
Matt-Esch: I've added tests for the full permutation of 7 instances, the function is fully covered.

codegen/module.go
ChuntaoLu: Sorting would be unnecessary when there is a cycle, also the full cycle can be constructed from the `classDeps` map. But again, basing on our use case, it is probably not a big deal.

codegen/module.go
ChuntaoLu: 👍

codegen/templates/endpoint.tmpl
```diff
@@ -1,6 +1,20 @@
...
+{{- $instance := .Instance }}
+package {{$instance.PackageInfo.PackageName}}
+
+{{- $workflow := .WorkflowName }}
+{{- $reqHeaderMap := .ReqHeaderMap }}
+{{- $reqHeaderMapKeys := .ReqHeaderMapKeys }}
+{{- $resHeaderMap := .ResHeaderMap }}
+{{- $resHeaderMapKeys := .ResHeaderMapKeys }}
+{{- $clientName := title .ClientName }}
+{{- $serviceMethod := printf "%s%s" .Method.ThriftService .Method.Name }}
```
Matt-Esch: Yeah I picked the endpoint (not tchannel endpoint) convention just because I don't like computing stuff in the templates. I've updated it per your suggestion.

PR #163, opened 2017-07-15
Add support for nested query parameters.

Raynos:
This PR adds support for GET requests whose body has nested
fields.

 - Parse query parameters like `/foo?bar.baz.x=4`
 - Populated nested structs
 - Parse headers into nested fields.

r: @uber/zanzibar-team


codegen/method.go
```diff
@@ -407,25 +410,77 @@ func findParamsAnnotation(
...
 func (ms *MethodSpec) setRequestHeaderFields(
-	funcSpec *compile.FunctionSpec,
-) {
+	funcSpec *compile.FunctionSpec, packageHelper *PackageHelper,
+) error {
 	fields := compile.FieldGroup(funcSpec.ArgsSpec)
-	ms.ReqHeaderFields = map[string]HeaderFieldInfo{}
+	// ms.ReqHeaderFields = map[string]HeaderFieldInfo{}
+
+	statements := LineBuilder{}
```
zhenghuiwang: statements := `&`LineBuilder{}  to avoid a copy?

codegen/method.go
```diff
@@ -622,73 +654,113 @@ func getQueryMethodForType(typeSpec compile.TypeSpec) string {
...
+
+			return false
+		}
+
+		var longQueryName string
+		if thriftPrefix == "" {
+			longQueryName = field.Name
+		} else {
+			longQueryName = thriftPrefix + "." + field.Name
+			if longQueryName[0] == '.' {
```
zhenghuiwang: Is this nested `if` necessary? can it be treated as a case for the outer `if` so that we don't need nested if?

codegen/method.go
```diff
@@ -622,73 +654,113 @@ func getQueryMethodForType(typeSpec compile.TypeSpec) string {
...
+				longFieldName, typeName,
+			)
+			statements.append("}")
+
+			return false
+		}
+
+		var longQueryName string
+		if thriftPrefix == "" {
+			longQueryName = field.Name
```
zhenghuiwang: extra space for `. Name`?

codegen/method.go
Raynos: how does it avoid a copy ? should i use `var statements LineBuilder` instead ?

codegen/method.go
```diff
@@ -63,10 +63,8 @@ type MethodSpec struct {
...
 	QueryParamGoStatements []string
 
-	// ReqHeaderFields is a map of "header name" to
-	// a golang field accessor expression like ".Foo.Bar"
-	// Use to place request headers in the body
-	ReqHeaderFields map[string]HeaderFieldInfo
+	// Statements for reading request headers
+	ReqHeaderGoStatements []string
 
 	// ResHeaderFields is a map of header name to a golang
```
stevededalus: Should we also convert ResHeaderField to GoStatements to be consistent?

codegen/test_data/bar.json
```diff
@@ -157,7 +157,7 @@
 					],
 					"IsEndpoint": false,
 					"QueryParamGoStatements": null,
-					"ReqHeaderFields": {},
+					"ReqHeaderGoStatements": null,
```
stevededalus: Let's be careful when updating edge-gateway, this will break a lot of existing configs

codegen/thrift.go
```diff
@@ -133,3 +133,93 @@ func isHashable(spec compile.TypeSpec) bool {
...
+				goPrefix+"."+pascalCase(field.Name),
+				thriftPrefix+"."+field.Name,
+				t.Fields,
+				visitField,
+			)
+			if bail {
+				return true
+			}
+		case *compile.SetSpec:
+			// TODO: implement
```
stevededalus: Should we panic on unimplemented types so that they are rejects in the UI continuous integration?

codegen/test_data/bar.json
Raynos: this is just part of tests, not the config file format.

codegen/thrift.go
Raynos: Makes sense.

codegen/method.go
Raynos: The `ResHeaderField` currently does not generate any code. So there are no go statements to generate.

codegen/method.go
Raynos: That doesnt pass gofmt; gofmt doesnt want any spaces.

codegen/method.go
Raynos: will fix.

PR #166, opened 2017-07-20
Resolve class order regardless of registration order

ChuntaoLu:
This PR resolves the module class order based on class dependency instead of the registration order.


ChuntaoLu:
@Matt-Esch PR is updated to cover all cases, sorting code is fully covered.

The build is failing because the thriftrw branch we are on is deleted, https://github.com/thriftrw/thriftrw-go/pull/305#event-1172995186, upgrading thriftrw doesn't work as the "required" json tag feature is missing in thriftrw upstream.

PR #168, opened 2017-07-21
Adding runtime metrics

ravirajj:
Currently collection of CPU, Mem or GC metrics is configurable. Should we allow configuration of individual runtime metrics?


runtime/runtime_metrics.go
```diff
@@ -0,0 +1,267 @@
...
+	rm.Start()
+	return rm
+}
+
+// NewRuntimeMetricsCollector creates a new runtime metrics collector.
+func NewRuntimeMetricsCollector(
+	config RuntimeMetricsConfiguration,
+	scope tally.Scope,
+) RuntimeMetricsCollector {
+	return &runtimeCollector{
```
Raynos: We need to tag the host name or application identifier for these metrics to be useful.

ravirajj: The host, env and service tags are added by the m3.reporter from the m3.options and commonTags.

Raynos: I'm pretty sure we dont ( https://github.com/uber/zanzibar/blob/master/test/health_test.go#L159-L160 ). There are only two tags by default, need to write a test.

ravirajj:
Updated commit with tests.

runtime/gateway.go
```diff
@@ -276,19 +281,25 @@ func (gateway *Gateway) setupConfig(config *StaticConfig) {
...
 		}
 
-		env := config.MustGetString("env")
-
-		commonTags := map[string]string{"env": env}
+		// TODO: Why aren't common tags emitted?
+		// m3.Options.Env and m3.Options.Service adds 'env' and 'service' common tags.
+		// Providing 'host' common tag to prevent m3.Options.IncludeHost from returning an error on 'unknown'host.
+		commonTags := map[string]string{
+			"host": hostname,
```
Raynos: we do not want the host tag at all, please pass the 'env' tag explicitly, it's less confusing, we can drop the `Env` field in `m3.Options`.

It's also verify concerning that the `env` common tag does not show up in the health test with metrics.

runtime/gateway.go
```diff
@@ -311,18 +322,31 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) error {
...
+	gateway.runtimeMetrics = StartRuntimeMetricsCollector(
+		RuntimeMetricsOptions{
+			EnableCPUMetrics: config.MustGetBoolean("metrics.runtime.enableCPUMetrics"),
+			EnableMemMetrics: config.MustGetBoolean("metrics.runtime.enableMemMetrics"),
+			EnableGCMetrics:  config.MustGetBoolean("metrics.runtime.enableGCMetrics"),
+			CollectInterval:  time.Duration(config.MustGetInt("metrics.runtime.collectIntervalMs")) * time.Millisecond,
+		},
+		gateway.MetricScope.SubScope("runtime").Tagged(map[string]string{
+			"host":    hostname,
+			"service": service,
```
Raynos: service tag should be defaulted in `m3.Reporter` itself, do not duplicate.

runtime/gateway.go
```diff
@@ -311,18 +322,31 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) error {
...
+			"env":     env,
```
Raynos: env tag is already defined in commonTags, do not duplicate.

runtime/gateway.go
```diff
@@ -311,18 +322,31 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) error {
...
+	// start collecting runtime metrics
+	gateway.runtimeMetrics = StartRuntimeMetricsCollector(
+		RuntimeMetricsOptions{
+			EnableCPUMetrics: config.MustGetBoolean("metrics.runtime.enableCPUMetrics"),
+			EnableMemMetrics: config.MustGetBoolean("metrics.runtime.enableMemMetrics"),
+			EnableGCMetrics:  config.MustGetBoolean("metrics.runtime.enableGCMetrics"),
+			CollectInterval:  time.Duration(config.MustGetInt("metrics.runtime.collectIntervalMs")) * time.Millisecond,
+		},
+		gateway.MetricScope.SubScope("runtime").Tagged(map[string]string{
+			"host":    hostname,
```
Raynos: we actually want a new root scope here instead of a sub scope.

The current `MetricScope` is called `all-workers`.

Let's create a new rootscope with the a prefix like `service + ".production.per-worker"` and have that root scope have default tags including the host.

We should store a `PerWorkerMetricScope` on the gateway as well so others can use it.

runtime/gateway.go
ravirajj: NewReporter will return an error if `env` is defined.

ravirajj: Done.  Yes, don't show up with the endpoint metrics too.

runtime/gateway.go
ravirajj: As discussed, creating one RootScope for `service + ".production"` and subscopes for `all-workers` and `per-workers`.

runtime/gateway.go
ravirajj: Done. Removed service. The service tag should come from the commonTags.

runtime/gateway.go
ravirajj: Done. Removed env tag.

runtime/gateway.go
ravirajj: Why don't you want host in the backend?

runtime/gateway.go
```diff
@@ -62,20 +62,22 @@ type Gateway struct {
 	WaitGroup        *sync.WaitGroup
 	Channel          *tchannel.Channel
 	Logger           *zap.Logger
-	MetricScope      tally.Scope
+	RootScope        tally.Scope
```
Raynos: this is technically a breaking change.

Let's do `RootScope` , `MetricsScope` , `PerWorkerMetricsScope`

runtime/gateway.go
```diff
@@ -279,74 +281,57 @@ func (gateway *Gateway) setupConfig(config *StaticConfig) {
...
+	rootScopeOpts := tally.ScopeOptions{
+		Tags:           defaultTags,
+		Prefix:         service + ".production",
+		CachedReporter: gateway.metricsBackend,
+		Separator:      tally.DefaultSeparator,
+	}
+	reportInterval := time.Duration(config.MustGetInt("metrics.tally.flushInterval")) * time.Millisecond
+	gateway.RootScope, gateway.scopeCloser = tally.NewRootScope(rootScopeOpts, reportInterval)
+	gateway.AllWorkersScope = gateway.RootScope.SubScope("all-workers")
+	gateway.PerWorkerScope = gateway.RootScope.SubScope("per-worker")
```
Raynos: The PerWorker scope needs the host tag in it.

test/health_test.go
```diff
@@ -249,63 +247,53 @@ func TestRuntimeMetrics(t *testing.T) {
...
-			assert.Equal(t, expectedTags[tag.GetTagName()], tag.GetTagValue(), "expected tag value to be correct")
-		}
-	}
-}
-
-func getHostname() string {
-	host, err := os.Hostname()
-	if err != nil {
-		return "unknown"
+		// TODO: Why are common tags not emitted?
```
Raynos: lets figure out the common tags thing in this PR.

test/endpoints/baz/baz_metrics_test.go
```diff
@@ -88,7 +88,7 @@ func TestCallMetrics(t *testing.T) {
 		return
 	}
 
-	cg.MetricsWaitGroup.Wait()
+	<-cg.MetricsWaitGroup.Wait
```
Raynos: channels should not be part of the public interface, they are an implementation detail.

Let's add a `Wait()` method that calls `<-cg.MetricsWaiter.waitCh`

test/health_test.go
```diff
@@ -131,7 +131,7 @@ func TestHealthMetrics(t *testing.T) {
 
 	assert.Equal(t, res.Status, "200 OK", "got http 200")
 
-	cgateway.MetricsWaitGroup.Wait()
+	<-cgateway.MetricsWaitGroup.Wait
```
Raynos: What if the channel is already done ?

test/lib/wait_at_least.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Done subtracts one from the wait count.
+func (w *WaitAtLeast) Done() {
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+
+	w.count--
+	if w.count == 0 {
+		w.Wait <- true
```
Raynos: If we read from the channel after the counter is zero then it will block forever.

This really needs a Wait() method which will return immediately if the count is zero or negative and otherwise waits on the channel.

test/lib/wait_at_least.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"sync"
+)
+
+// WaitAtLeast signals the wait channel when at least the added count are done.
+// This avoids sync.WaitGroup panic when count is negative.
+//
+// Usage:
+//     w := WaitAtLeast{ Wait: make(chan bool) }
```
Raynos: Channel is an implementation detail. Make it a private field and add a `NewWaitAtLeast` constructor function.

test/endpoints/baz/baz_metrics_test.go
ravirajj: Actually, I had that way initially, but decided to expose the Channel similar to how sync.Cond exposes the Locker.  I'll change it back!

runtime/gateway.go
ravirajj: Done.

test/health_test.go
ravirajj: As discussed, wrapping sync.WaitGroup with WaitAtLeast.

test/endpoints/baz/baz_metrics_test.go
ravirajj: Done. Replaced with Wait().

test/lib/wait_at_least.go
ravirajj: Done. As discussed, wrapping sync.WaitGroup with WaitAtLeast.

test/lib/wait_at_least.go
ravirajj: Done.

test/health_test.go
ravirajj: I'll need your help with this one.

runtime/gateway.go
ravirajj: Done.

test/health_test.go
Raynos: if you dont think common tags work then lets remove all common tag related configuration and create our own default tags.

runtime/gateway.go
```diff
@@ -274,55 +281,57 @@ func (gateway *Gateway) setupConfig(config *StaticConfig) {
...
-		m3Backend, err := m3.NewReporter(m3.Options{
+		// TODO: Why aren't common tags emitted?
+		// NewReporter adds 'env' and 'service' common tags; and no 'host' tag.
+		commonTags := map[string]string{
+			"env": config.MustGetString("env"),
+		}
+		opts := m3.Options{
 			HostPorts:          []string{config.MustGetString("metrics.m3.hostPort")},
-			Service:            config.MustGetString("metrics.tally.service"),
+			Service:            service,
```
Raynos: Since `commonTags` does not work we should just not pass `Service` and `CommonTags`

runtime/gateway.go
```diff
@@ -274,55 +281,57 @@ func (gateway *Gateway) setupConfig(config *StaticConfig) {
...
+		if gateway.metricsBackend, err = m3.NewReporter(opts); err != nil {
+			return err
 		}
-		metricsBackend = gateway.metricsBackend
+	} else if gateway.metricsBackend == nil {
+		panic("expected gateway to have MetricsBackend in opts")
 	}
 
 	// TODO: decide what default tags we want...
 	defaultTags := map[string]string{}
```
Raynos: We should put env and service here in default tags.

runtime/gateway.go
```diff
@@ -274,55 +281,57 @@ func (gateway *Gateway) setupConfig(config *StaticConfig) {
...
+		CachedReporter: gateway.metricsBackend,
 		Separator:      tally.DefaultSeparator,
-	}, time.Duration(flushIntervalConfig)*time.Millisecond)
-	gateway.MetricScope = scope
-	gateway.metricScopeCloser = closer
+	}
+	reportInterval := time.Duration(config.MustGetInt("metrics.tally.flushInterval")) * time.Millisecond
+	gateway.RootScope, gateway.scopeCloser = tally.NewRootScope(rootScopeOpts, reportInterval)
+	gateway.MetricsScope = gateway.RootScope.SubScope("all-workers")
+	gateway.PerWorkerMetricsScope = gateway.RootScope.SubScope("per-worker")
```
Raynos: still need the hostname tag here.

runtime/gateway.go
ravirajj: Done. Moving tags to default tags since commonTags is not working.

runtime/gateway.go
ravirajj: Done. Moving commonTags to defaultTags.

runtime/gateway.go
ravirajj: I need to set Service and Env otherwise m3.NewReporter() returns error.

runtime/gateway.go
ravirajj: Done.

test/lib/util.go
```diff
@@ -0,0 +1,32 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package lib
+
+import "os"
+
+// GetHostname returns os host name; otherwise 'unknown'
+func GetHostname() string {
```
Raynos: Just make `zanzibar.getHostName` public so the tests can re-use `zanzibar.GetHostName` instead of copying this function.

ravirajj: Done.

PR #169, opened 2017-07-24
Add support for HTTP Client with GET + params method

Raynos:
This loosens the limitation on the HTTP client.

Now a GET method can have either "no body" or a "body that only has params fields".

Query parameters for HTTP client is still not implemented yet.

r: @uber/zanzibar-team


codegen/method.go
```diff
@@ -219,6 +223,34 @@ func NewMethod(
...
+	visitor := func(
+		goPrefix string, thriftPrefix string, field *compile.FieldSpec,
+	) bool {
+		realType := compile.RootTypeSpec(field.Type)
+		// ignore nested structs
+		if _, ok := realType.(*compile.StructSpec); ok {
+			return false
+		}
+
+		if param, ok := field.Annotations[antHTTPRef]; ok {
```
zhenghuiwang: Nit: this if can be simplified as 

```
if param, ok  := field.Annotations[antHTTPRef]; !ok || param[0:6]!="params" {
				hasNonParams = true
				return true
}
```

Raynos: fixed.

PR #170, opened 2017-07-25
Read/Write Thrift file meta in a repository

zhenghuiwang:
This PR adds "idls.json" file to store the meta data of thrift files in a repository.

In next PR, I will add a thrift registration repository to update the thrift files and the meta data in "idls.json".

### DIFF
--- a/examples/example-gateway/idls.json
+++ b/examples/example-gateway/idls.json
@@ -1,0 +1,50 @@
+{
+	"clients/bar/bar.thrift": {
+		"path": "clients/bar/bar.thrift",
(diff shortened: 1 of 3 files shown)


examples/example-gateway/idls.json
```diff
@@ -0,0 +1,50 @@
+{
```
Raynos: What generates this file ?

zhenghuiwang: The function that generates this file is not included in this PR. I just ran the func locally once.

PR #174, opened 2017-07-27
Expose dependencies tree on BenchGateway for testing

Raynos:
By exposing the dependency tree on the BenchGateway class,
developers can write tests with all the components and
test backends pre-configured and then grab a task/client
"module instance" and use it for integration tests.

r: @uber/zanzibar-team

### DIFF
--- a/codegen/templates/service.tmpl
+++ b/codegen/templates/service.tmpl
@@ -14,24 +14,27 @@ import (
 	module "{{$instance.PackageInfo.ModulePackagePath}}"
 )
 
+// DependenciesTree re-exported for convenience.
+type DependenciesTree module.DependenciesTree
+
 // CreateGateway creates a new instances of the {{$instance.InstanceName}}
 // service with the specified config
 func CreateGateway(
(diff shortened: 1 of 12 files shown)


codegen/templates/service.tmpl
```diff
@@ -14,24 +14,27 @@ import (
...
+// DependenciesTree re-exported for convenience.
+type DependenciesTree module.DependenciesTree
+
 // CreateGateway creates a new instances of the {{$instance.InstanceName}}
 // service with the specified config
 func CreateGateway(
 	config *zanzibar.StaticConfig,
 	opts *zanzibar.Options,
-) (*zanzibar.Gateway, error) {
+) (*zanzibar.Gateway, interface{}, error) {
```
srivastavankit: can we change this to named variable?

Raynos: Does that help ? i find named return variables confusing.

srivastavankit: for something like interface{}, it at least documents what that interface{} contextually means.

PR #175, opened 2017-07-31
Expose more error information in http client.

Raynos:
This changes the http client to return an error struct
with the status code and raw body.

r: @uber/zanzibar-team

### DIFF
--- a/runtime/http_client.go
+++ b/runtime/http_client.go
@@ -20,8 +20,12 @@
 
 package zanzibar
 
-import "net/http"
-import "go.uber.org/zap"
+import (
+	"net/http"
+	"strconv"
+
+	"go.uber.org/zap"
+)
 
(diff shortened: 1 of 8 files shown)


runtime/http_client.go
```diff
@@ -32,6 +36,17 @@ type HTTPClient struct {
...
 
+// RawHTTPError defines an error for HTTP
+type RawHTTPError struct {
+	StatusCode int
+	RawBody    []byte
+}
+
+func (rawErr *RawHTTPError) Error() string {
+	return "Unexpected http client response (" +
+		strconv.Itoa(rawErr.StatusCode) + ")"
```
Raynos: Sure can rename it.

Rather not put the entire body in the in the error message as it's just raw bytes, too easy to return the error back out which leads to double encoded JSON.

PR #177, opened 2017-07-31
Add http handlers to modify a gateway repository

zhenghuiwang:
This PR adds HTTP handlers for viewing/updating gateway configurations.

- Add first two handlers: one returns the gateway configs; the other modifies configurations and creates a diff for the changes. More handlers will be in the following PRs.

- Abstract two interfaces for `Logger` and `DiffCreater`. Logger is based on `zap.Logger` but other loggers can be easily plugged in.

- Set up local HTTP server to test handlers against test cases that are saved on disk as a JSON file.

r: @uber/zanzibar-team


backend/repository/handler.go
```diff
@@ -0,0 +1,223 @@
...
+	return &Handler{
+		Manager:       m,
+		DiffCreater:   dc,
+		gatewayHeader: gatewayHeader,
+		logger:        logger,
+	}
+}
+
+// NewHTTPRouter constructs the endpoints for http server.
+func (h *Handler) NewHTTPRouter() *httprouter.Router {
```
zhenghuiwang: I'd like to use Zanzibar `HTTPRouter` here, but the router is coupled with `Gateway`. `Gateway` was created for performance so it chooses specific logger and metrics. The static init config for Gateway is also complex to a HTTP server.

I aim to provide a bundle of handlers that can be used immediately but don't want to confine the way of organizing them. The routing function here is not essential. I can change the handlers to be standard `http.HandleFunc` and move routings here into tests. So zanzibar users can pick up whatever handler they need to fit in their HTTP routing/server framework.

Raynos: is it possible to create a `HandlerServer` instead of a `Handler`

We create a function `NewRepositoryHandler` server that takes some options ( like a port / ip ) and starts a gateway and initializes all of the things internally.

zanzibar has evolved away from gateway to service baseline, we should look into using it for edge-gateway-orchestrator.

I think it's out of scope for this PR as we might need extra refactoring in zanzibar/runtime/ to make it compatible with edge gateway orchestrator

zhenghuiwang: Yes, I can work on making `HandlerServer` in future PRs and use it in orchestrator.

PR #178, opened 2017-07-31
Generalize template helper

Matt-Esch:
I've generalized the template helper over a generic asset collection so that we can use this inside other codebases with different collections of templates.

### DIFF
--- a/codegen/template.go
+++ b/codegen/template.go
@@ -33,6 +33,24 @@ import (
 	"github.com/uber/zanzibar/codegen/template_bundle"
 )
 
+// AssetProvider provides access to template assets
+type AssetProvider interface {
+	// AssetNames returns a list of named assets available
+	AssetNames() []string
+	// Asset returns the bytes for a provided asset name
+	Asset(string) ([]byte, error)
+}
+
+type defaultAssetCollection struct{}
+
+func (*defaultAssetCollection) AssetNames() []string {
+	return templates.AssetNames()
+}
+
+func (*defaultAssetCollection) Asset(assetName string) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


codegen/template.go
```diff
@@ -81,9 +99,17 @@ type Template struct {
...
 func NewTemplate() (*Template, error) {
+	return NewTemplateFromAssetProvider(&defaultAssetCollection{})
+}
+
+// NewTemplateFromAssetProvider returns a template helper for the
+// provided asset collection
+func NewTemplateFromAssetProvider(
+	assetProvider AssetProvider,
+) (*Template, error) {
 	t := tmpl.New("main").Funcs(funcMap)
```
zhenghuiwang: `funcMap` is not exposed here. If we want to use customized template, `funcMap` should also be part of the params of constructor `NewTemplateFromAssetProvider`.

codegen/template.go
```diff
@@ -81,9 +99,17 @@ type Template struct {
 
 // NewTemplate creates a bundle of templates.
 func NewTemplate() (*Template, error) {
```
zhenghuiwang: Nit: I would rename
`NewTemplate`-> `NewDefaultTemplate`
`NewTemplateFromAssetProvider` -> `NewTemplate`, because the arguments type says `FromAssetProvider`

Matt-Esch: Yeah I did think about this, was trying to avoid the breaking change.

zhenghuiwang: Will find/replace `NewTemplate` break any user of zanzibar? I think users don't need to call `NewTemplate` directly but call module system.

PR #183, opened 2017-08-09
Initial query parameter support

Raynos:
This is ready for initial review.

 - Add support for query parameters in generated HTTP client
 - Encode types properly into query params
 - Fix server-side header parsing logic.

There are some future TODOs still pending:

 - [x] Ensure we add nil checks for nested structs

r: @uber/zanzibar-team


ravirajj:
Can you add a test case for thrift with recursive nested structs:

```
struct Strategy {
  1: optional Strategy subStrategy
}

service MyService {
  void method(
    1: optional Strategy strategy
  )
}
```

Raynos:
recursive data structures are out of scope for this PR.

Also recursive data structures for query parameters are going to be really annoying and require a different approach to populating values. I'd rather avoid it until there is a use case.

I suspect there is a use case for recursive data structures in HTTP+JSON using POST and json bodies ( not query parameters ).

PR #185, opened 2017-08-11
Add config flag to specify runtime config files

Matt-Esch:
I have removed the hard-coded configurations for each binary. Now the defaults are bundled with the runtime, and extra configs are passed with a --config flag to the binary.


config/production.json.go
```diff
@@ -0,0 +1,252 @@
...
+	"metrics.runtime.enableMemMetrics": true,
+	"metrics.runtime.enableGCMetrics": true,
+	"metrics.runtime.collectInterval": 1000,
+
+	"tchannel.serviceName": "my-gateway",
+	"tchannel.processName": "my-gateway"
+}
+`)
+
+func productionJsonBytes() ([]byte, error) {
```
ChuntaoLu: Is this function necessary? Seems like it is only used in the function below and doesn't do much.

config/production.json.go
```diff
@@ -0,0 +1,252 @@
...
+	return fi.modTime
+}
+func (fi bindataFileInfo) IsDir() bool {
+	return false
+}
+func (fi bindataFileInfo) Sys() interface{} {
+	return nil
+}
+
+var _productionJson = []byte(`{
```
ChuntaoLu: How is this going to be updated? And how do we make sure it's valid json?

Matt-Esch: It's generated the same way that the template bundle is generated, it's added to the generate makefile. Staleness is caught in tests

config/production.json.go
Matt-Esch: It's a generated function. It returns the bytes for productionJson, and it's a private function

PR #186, opened 2017-08-15
Support module class to map to multiple directories

ChuntaoLu:
Currently the module system has 1:1 mapping from module class to the top level hosting directory, e.g., all endpoints configs reside in `example-gateway/endpoints`. This can be cumbersome for applications that, for whatever reason, prefer to have multiple top level directories for the same module class. 

This PR adds API to register additional config directories for a known module class.
@uber/zanzibar-team

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -70,32 +70,38 @@ func (system *ModuleSystem) RegisterClass(class ModuleClass) error {
 	}
 
 	if system.classes[name] != nil {
-		return errors.Errorf(
-			"The module class %q is already defined",
-			name,
-		)
+		return errors.Errorf("Module class %q is already defined", name)
 	}
 
-	class.Directory = filepath.Clean(class.Directory)
+	for i, dir := range class.Directories {
+		dir = filepath.Clean(dir)
 
-	if strings.HasPrefix(class.Directory, "..") {
-		return errors.Errorf(
-			"The module class %q must map to an internal directory but was %q",
-			name,
-			class.Directory,
-		)
-	}
-
-	// Validate the module class directory name is unique
-	for moduleClassName, moduleClass := range system.classes {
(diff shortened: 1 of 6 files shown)


codegen/module.go
```diff
@@ -104,6 +90,79 @@ func (system *ModuleSystem) RegisterClass(class ModuleClass) error {
...
+		unique[i] = key
+		i++
+	}
+	sort.Strings(unique)
+	return unique
+}
+
+// validateDir checks if the module class can map to the given dir
+func (system *ModuleSystem) validateClassDir(class *ModuleClass, dir string) error {
+	dir = filepath.Clean(dir)
```
Matt-Esch: You have cleaned the dirs twice

codegen/module.go
```diff
@@ -104,6 +90,79 @@ func (system *ModuleSystem) RegisterClass(class ModuleClass) error {
...
+
+	if strings.HasPrefix(dir, "..") {
+		return errors.Errorf(
+			"Module class %q must map to internal directories but found %q",
+			class.Name,
+			dir,
+		)
+	}
+
+	// Validate the module class directories are unique
```
Matt-Esch: It might not be strictly necessary to enforce this requirement. It would be possible for two class types to live in the same directory, as the instances are distinguishable by their config name.

codegen/module.go
ChuntaoLu: This clean step is part of the validation which is used at multiple places and does not assume the given dir is clean. 

The other clean is before dedup, which is before validation. So yes, we cleaned twice in `RegisterClass`, but it doesn't hurt.

codegen/module.go
ChuntaoLu: Makes sense, this logic is now removed and I have added a test case for it as well.

PR #187, opened 2017-08-16
Expose all client methods by default

zhenghuiwang:
When updating a client config, then request needs to provide exposed methods for the client.

This PR adds `exposedMethods` to be all methods defined in client thrift service if no exposed methods provided.

r: @uber/zanzibar-team

### DIFF
--- a/backend/repository/data/client/contacts_client_update.json
+++ b/backend/repository/data/client/contacts_client_update.json
@@ -4,7 +4,7 @@
     "type": "http",
     "thriftFile": "clients/contacts/contacts.thrift",
     "exposedMethods": {
-        "SaveContacts": "Contacts::saveContacts"
+        "saveContacts": "Contacts::saveContacts"
     },
     "ip": "127.0.0.1",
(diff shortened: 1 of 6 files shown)


backend/repository/data/client/contacts_client_update.json
```diff
@@ -4,7 +4,7 @@
     "type": "http",
     "thriftFile": "clients/contacts/contacts.thrift",
     "exposedMethods": {
-        "SaveContacts": "Contacts::saveContacts"
+        "saveContacts": "Contacts::saveContacts"
```
ChuntaoLu: what's the rationale making the method name lowercase? The generated method will always have first letter uppercase.

zhenghuiwang: the method name in the thrift service is lower case so i make it the same as the service name. anyway I will make it the upper case then.

ChuntaoLu: 👍

PR #188, opened 2017-08-17
Add a http handler for landing diff

zhenghuiwang:
plus a check to read only thrift files for parsing thrift services.

r: @uber/zanzibar-team

### DIFF
--- a/backend/repository/handler.go
+++ b/backend/repository/handler.go
@@ -90,8 +90,9 @@ func (h *Handler) NewHTTPRouter() *httprouter.Router {
 	r.GET("/thrift-service/*path", h.ThriftServicesByPath)
 	r.GET("/thrift-list", h.ThriftList)
 	r.GET("/thrift-file/*path", h.ThriftFile)
-	r.POST("/create-diff", h.CreateDiff)
 	r.POST("/validate-updates", h.ValidateUpdates)
+	r.POST("/create-diff", h.CreateDiff)
+	r.POST("/land-diff", h.LandDiff)
 	return r
 }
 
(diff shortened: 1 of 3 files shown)


backend/repository/handler.go
```diff
@@ -347,6 +348,45 @@ func (h *Handler) CreateDiff(w http.ResponseWriter, r *http.Request, ps httprout
 	h.WriteJSON(w, http.StatusOK, &createDiffResponse{DiffURI: resp})
 }
 
+// LandDiff lands a diff.
+func (h *Handler) LandDiff(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	type LandDiffRequest struct {
```
ChuntaoLu: Shouldn't this be defined outside the function?

backend/repository/handler.go
```diff
@@ -347,6 +348,45 @@ func (h *Handler) CreateDiff(w http.ResponseWriter, r *http.Request, ps httprout
...
+	if err != nil {
+		h.WriteErrorResponse(w, http.StatusNotFound, err)
+		return
+	}
+	repo, err := h.Manager.NewRuntimeRepository(gatewayConfig.ID)
+	if err != nil {
+		h.WriteErrorResponse(w, http.StatusInternalServerError, errors.Wrap(err, "failed to create temp runtime dirs"))
+		return
+	}
+	h.logger.Info("Landing diff...",
```
ChuntaoLu: seems like only happy path is logged? do we not need log the errors?

backend/repository/handler.go
zhenghuiwang: I prefer defining type inside a func if the struct type is only to be used within the func, similar to a func scope variable. Parsing the request is a one-time thing. If we have multiple handlers like this, defining a type inside a func seems to be more readable.

backend/repository/handler.go
zhenghuiwang: Errors are logged inside the `h.WriteErrorResponse` func

ChuntaoLu: 👍

backend/repository/handler.go
ChuntaoLu: parsing is one-time, but per request. This particular handler is probably of low volume. I wonder if this practice cause overhead for high volume requests, in which case the same type declaration repeats more often.

Regarding readability, I think it makes sense for small structs, and gets awkward for non trivial ones.

PR #190, opened 2017-08-17
Adding recursive request struct fields

ravirajj:


ChuntaoLu:
Is the conclusion that we don't allow annotation inside a recursive field?

ravirajj:
@ChuntaoLu Yes, annotation won't be allowed inside recursive fields.  It may capture the top level recursive struct annotations, but won't see any nested recursive struct annotations.

PR #192, opened 2017-08-22
Adding service environment config

ravirajj:
User can add `service.env.config` to override existing config values with environment values. For example:

```
	"service.env.config": {
		"UBER_PORT_HTTP": {
			"key": "http.port",
			"dataType": "number"
		},
		"UBER_PORT_TCHANNEL": {
			"key": "tchannel.port",
			"dataType": "number"
		}
	}
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -21,27 +21,39 @@
 package config
 
 import (
-	"github.com/pkg/errors"
-	zanzibar "github.com/uber/zanzibar/runtime"
+	"os"
+
+	"github.com/uber/zanzibar/runtime"
 )
 
-// NewRuntimeConfigOrDie returns a static config struct that is pre-set with
-// the service configuration defaults
+// EnvConfig map from environment variable to config key and data type
+type EnvConfig map[string]struct {
+	Key      string `json:"key"`
+	DataType string `json:"dataType"`
+}
+
+// NewRuntimeConfigOrDie returns a static config struct
+// that is pre-set with the service configuration defaults
+// and overridden by service.env.config
 func NewRuntimeConfigOrDie(
 	files []string,
 	seedConfig map[string]interface{},
 ) *zanzibar.StaticConfig {
 	defaultConfig, err := defaultConfig()
 	if err != nil {
-		panic(errors.Wrap)
-	}
-	serviceConfig := make([]*zanzibar.ConfigOption, len(files)+1)
-	serviceConfig[0] = defaultConfig
-	for i, configFilePath := range files {
-		serviceConfig[i+1] = zanzibar.ConfigFilePath(configFilePath)
+		panic("error getting default config")
 	}
 
-	return zanzibar.NewStaticConfigOrDie(serviceConfig, seedConfig)
+	var serviceConfig []*zanzibar.ConfigOption
+	serviceConfig = append(serviceConfig, defaultConfig)
+	for _, configFilePath := range files {
+		serviceConfig = append(serviceConfig, zanzibar.ConfigFilePath(configFilePath))
+	}
+
+	staticConfig := zanzibar.NewStaticConfigOrDie(serviceConfig, seedConfig)
+	getEnvConfig(staticConfig)
+
+	return staticConfig
 }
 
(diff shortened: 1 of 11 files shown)


ravirajj:
Code coverage seems to be failing because of:
`# TODO: need better solution for coverage from different package`

config/config.go
```diff
@@ -21,34 +21,55 @@
...
 	if err != nil {
-		panic(errors.Wrap)
+		panic("error getting default config")
 	}
-	serviceConfig := make([]*zanzibar.ConfigOption, len(files)+1)
-	serviceConfig[0] = defaultConfig
-	for i, configFilePath := range files {
-		serviceConfig[i+1] = zanzibar.ConfigFilePath(configFilePath)
+
+	var serviceConfig []*zanzibar.ConfigOption
```
Matt-Esch: The previous allocation should remain in place, we know it will be at least this length

config/config.go
```diff
@@ -21,34 +21,55 @@
...
 func defaultConfig() (*zanzibar.ConfigOption, error) {
 	bytes, err := Asset("production.json")
 	if err != nil {
 		return nil, err
 	}
-
 	return zanzibar.ConfigFileContents(bytes), nil
 }
+
+func getEnvConfig(cfg *zanzibar.StaticConfig) {
```
Matt-Esch: the function shouldn't be called "get" if it doesn't return anything, "setEnvConfig" would be better

runtime/static_config.go
```diff
@@ -267,8 +265,34 @@ func (conf *StaticConfig) MustGetStruct(key string, ptr interface{}) {
 	panic(errors.Errorf("Key (%s) not available", key))
 }
 
+// SetConfigValue sets the static config value.
+// dataType can be a boolean, number or string.
+// SetConfigValue will panic if the config is frozen.
+func (conf *StaticConfig) SetConfigValue(key string, bytes []byte, dataType string) {
```
Matt-Esch: "OrDie" if you panic

Matt-Esch: This isn't clearly named because it doesn't distinguish itself clearly from `SetOrDie`. Might be better named as `SetRawOrDie`

Matt-Esch:
Code coverage fails because you reduced the coverage of static_config.go (It was 100% before and should remain 100%)

ravirajj:
@Matt-Esch Yes, and it reduced code coverage because it didn't run/include my config_test.go in my code coverage results, because it only includes test/clients and test/endpoints in the code coverage results. There is a `# TODO: need better solution for coverage from different package` comment acknowledging this in the `cover.sh` script.

Matt-Esch:
static_config_test.go should cover static_config.go 100%. The missing lines are the lines in SetConfigValue

ravirajj:
Ok I'll move my test from /test/config/config.go to static_config.go.  Thanks!

config/config.go
ravirajj: Setting capacity to known files+1 length.  append() will not copy the array if there is capacity to append an item.

config/config.go
ravirajj: Done.

runtime/static_config.go
ravirajj: Done.

ravirajj: Yeah, I was thinking that too. How about changing SetOrDie to SetSeedOrDie(), so both are clear.

Matt-Esch:
> Ok I'll move my test from /test/config/config.go to static_config.go. Thanks! I was trying to test the logic in config/config.go, which implicitly tests the static_config.go SetConfigValue method.

That wasn't the suggestion. The suggestion is that you should write tests that only touch static_config.go from static_config_test.go. You should not use implicit tests to cover the feature in static_config.go, you should be able to run static_config_test.go and it covers static_config.go 100% and **only** static_config.go

ravirajj:
@Matt-Esch Not sure if you saw the latest diff, but as you had suggested, I added a test to static_config_test.go so it covers static_config.go 100%.

PR #194, opened 2017-08-28
Updated glide.yaml

ravirajj:


glide.yaml
```diff
@@ -29,7 +29,7 @@ import:
   subpackages:
   - lib/go/thrift
 - package: github.com/uber-go/tally
-  version: v2.1.0
+  version: ^3
```
Matt-Esch: So it looks to me like the fixed versions were on purpose, but I don't have the full context as to why we're managing dependencies like this. What I do find interesting is bumping a major version without any code changes. Do you understand what this changes/impacts?

ravirajj: Yes, I had a look at the differences between v2.1.0 and the latest v3:
- https://github.com/uber-go/tally/releases
- https://github.com/uber-go/tally/compare/v2.1.0...master

The only breaking change for zanzibar was in the how the metrics key names were set for TestScope snapshots, so they are unique by including tags.  

v3 also adds the ability now sanitizes M3 metric names, keys, values at the source; and provides some useful bug fixes for Close().

PR #195, opened 2017-08-29
Fixing metric names, tags, config and tests

ravirajj:
- The metrics are identified/key'ed by their name and tags combination.
- Removing redundant `env` and `service` from metric name, since every metric has the `env` and `service` tags.
- Removing redundant `per-workers` and `all-workers` from metric name, and as per M3 best practices, creating separate all-host and per-host metrics to reduce metric cardinality when querying metrics for all hosts.
- Removing redundant `metrics.m3.service` and `metrics.m3.flushInterval` config.
- Setting defaults in config/production.json instead of hardcoding it in the implementation.
- Using combined name and tag keys to check if metrics are emitted correctly in tests.


test/lib/test_m3_server/test_m3_server.go
```diff
@@ -141,18 +143,23 @@ func (m *FakeM3Service) GetMetrics() []*m3.Metric {
 // EmitMetricBatch is called by thrift message processor
 func (m *FakeM3Service) EmitMetricBatch(batch *m3.MetricBatch) (err error) {
 	m.lock.Lock()
+	defer func() { m.lock.Unlock() }()
```
ChuntaoLu: why not just `defer m.lock.Unlock()`?

test/lib/test_m3_server/test_m3_server.go
```diff
@@ -141,18 +143,23 @@ func (m *FakeM3Service) GetMetrics() []*m3.Metric {
...
+
 	m.batches = append(m.batches, batch)
 	if m.countBatches {
 		m.wg.Done()
 	}
 
 	for _, metric := range batch.Metrics {
-		m.metrics = append(m.metrics, metric)
+		tags := make(map[string]string)
```
ChuntaoLu: nit: preallocate map with `len(tags)`

test/lib/test_m3_server/test_m3_server.go
ravirajj: Done.

test/lib/test_m3_server/test_m3_server.go
ravirajj: Done.

runtime/gateway.go
```diff
@@ -57,23 +52,23 @@ type Options struct {
...
+	TChannelPort     int32
+	RealHTTPPort     int32
+	RealHTTPAddr     string
+	RealTChannelPort int32
+	RealTChannelAddr string
+	WaitGroup        *sync.WaitGroup
+	Channel          *tchannel.Channel
+	Logger           *zap.Logger
+	RootScope        tally.Scope
+	AllHostScope     tally.Scope
```
Raynos: breaking changes are expensive. we don't want to land breaking changes.

runtime/gateway.go
```diff
@@ -322,24 +317,23 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
...
+		time.Duration(config.MustGetInt("metrics.flushInterval"))*time.Millisecond,
 	)
-	gateway.MetricsScope = gateway.RootScope.SubScope("all-workers")
-	gateway.PerWorkerMetricsScope = gateway.RootScope.SubScope("per-worker").Tagged(map[string]string{
-		"host": GetHostname(),
-	})
+	// As per M3 best practices, creating separate all-host and per-host metrics
+	// to reduce metric cardinality when querying metrics for all hosts.
+	gateway.AllHostScope = gateway.RootScope.Tagged(map[string]string{"host": "all"})
+	gateway.PerHostScope = gateway.RootScope.Tagged(map[string]string{"host": GetHostname()})
```
Raynos: We want the per host scope to be prefixed with some unique prefix to make grafana querying easier.

Raynos:
See https://github.com/uber/zanzibar/pull/204

PR #196, opened 2017-08-31
Adding metrics and logs to client http request/response

ravirajj:
HTTP client emits the following metrics:

Names:
- outbound.calls.sent
- outbound.calls.success
- outbound.calls.errors
- outbound.calls.latency
- outbound.calls.status.XXX

Tags:
- host - “all”
- client - module instance name, e.g. “bar”, etc.
- method - module instance exposed method, e.g. “echostring”

### DIFF
--- a/codegen/test_data/clients/bar.gogen
+++ b/codegen/test_data/clients/bar.gogen
@@ -25,8 +25,10 @@ package barClient
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"strconv"
+	"time"
 
 	"github.com/pkg/errors"
 	clientsBarBar "github.com/uber/zanzibar/examples/example-gateway/build/gen-code/clients/bar/bar"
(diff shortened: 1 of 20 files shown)


runtime/client_http_request.go
```diff
@@ -129,8 +113,37 @@ func (req *ClientHTTPRequest) WriteJSON(
 // Do will send the request out.
 func (req *ClientHTTPRequest) Do(
 	ctx context.Context,
-) (*ClientHTTPResponse, error) {
-	res, err := req.client.Client.Do(
+) (resp *ClientHTTPResponse, err error) {
+	defer func() {
+		scope := req.client.scope.Tagged(map[string]string{
```
ChuntaoLu: It makes more sense to subscope the method scope at where the method is defined, not in the lower layer where request is made.

codegen/test_data/clients/bar.gogen
```diff
@@ -209,9 +220,7 @@ func (c *barClient) ArgNotStruct(
 	headers map[string]string,
 	r *clientsBarBar.Bar_ArgNotStruct_Args,
 ) (map[string]string, error) {
-	req := zanzibar.NewClientHTTPRequest(
-		c.clientID, "argNotStruct", c.httpClient,
-	)
+	req := c.httpClient.NewClientHTTPRequest("ArgNotStruct", "Bar", "argNotStruct")
```
ChuntaoLu: If you define/subscope the method scope at the method definition layer, there would be no need to pass around the client method name, which crosses the abstraction boundaries.

ravirajj: I was trying to follow the tchannel client style which has the metrics embedded in the client.

ravirajj:
@ChuntaoLu  I can create the sub scoped/tagged metrics/logger clients in the generated client method, and just pass them to `NewClientHTTPRequest(logger, scope)`, instead of passing the method names.

runtime/client_http_request.go
ravirajj: Creating pre-allocated metrics where the method created.

codegen/test_data/clients/bar.gogen
ravirajj: Done.

examples/example-gateway/config/production.json
```diff
@@ -1,15 +1,18 @@
 {
 	"clients.bar.ip": "127.0.0.1",
 	"clients.bar.port": 4001,
+	"clients.bar.timeout": 1000,
```
Raynos: nitpick: 10000; 10 seconds is a better default thats less aggressive.

ravirajj: Done

PR #198, opened 2017-08-31
Adding metrics and logs to tchannel client

ravirajj:
TChannel client emits the following metrics:

Names:
- outbound.calls.sent
- outbound.calls.success
- outbound.calls.app-errors
- outbound.calls.system-errors
- outbound.calls.latency

Tags:
- client - module instance name, e.g. “bar”, etc.
- method - module instance exposed method, e.g. “echostring”
- target-service - thrift service, e.g. “bar”, etc.
- target-endpoint - thrift method, e.g. “Bar::echostring”

### DIFF
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/uber-go/tally"
 	"github.com/uber/tchannel-go"
 	"go.uber.org/thriftrw/protocol"
 	"go.uber.org/zap"
@@ -53,6 +54,7 @@ type tchannelClient struct {
 	timeoutPerAttempt time.Duration
 	routingKey        *string
 	loggers           map[string]*zap.Logger
+	metrics           map[string]*OutboundTChannelMetrics
 }
 
 type tchannelCall struct {
@@ -68,22 +70,31 @@ type tchannelCall struct {
 	reqHeaders    map[string]string
 	resHeaders    map[string]string
 	logger        *zap.Logger
+	metrics       *OutboundTChannelMetrics
 }
 
 // NewTChannelClient returns a tchannelClient that makes calls over the given tchannel to the given thrift service.
 func NewTChannelClient(
 	ch *tchannel.Channel,
 	logger *zap.Logger,
+	scope tally.Scope,
 	opt *TChannelClientOption,
 ) TChannelClient {
 	loggers := make(map[string]*zap.Logger, len(opt.MethodNames))
+	metrics := make(map[string]*OutboundTChannelMetrics, len(opt.MethodNames))
 	for serviceMethod, methodName := range opt.MethodNames {
 		loggers[serviceMethod] = logger.With(
 			zap.String("clientID", opt.ClientID),
 			zap.String("methodName", methodName),
-			zap.String("target-service", opt.ServiceName),
-			zap.String("target-endpoint", serviceMethod),
+			zap.String("serviceName", opt.ServiceName),
+			zap.String("serviceMethod", serviceMethod),
 		)
+		metrics[serviceMethod] = NewOutboundTChannelMetrics(scope.Tagged(map[string]string{
+			"client":          opt.ClientID,
+			"method":          methodName,
+			"target-service":  opt.ServiceName,
+			"target-endpoint": serviceMethod,
+		}))
 	}
 
 	return &tchannelClient{
@@ -96,6 +107,7 @@ func NewTChannelClient(
 		timeoutPerAttempt: opt.TimeoutPerAttempt,
 		routingKey:        opt.RoutingKey,
 		loggers:           loggers,
+		metrics:           metrics,
 	}
 }
 
@@ -126,8 +138,9 @@ func (c *tchannelClient) Call(
 		serviceMethod: serviceMethod,
 		reqHeaders:    reqHeaders,
 		logger:        c.loggers[serviceMethod],
+		metrics:       c.metrics[serviceMethod],
 	}
-	defer call.finish()
+	defer func() { call.finish(err) }()
 	call.start()
 
 	err = c.ch.RunWithRetry(ctx, func(ctx netContext.Context, rs *tchannel.RequestState) (cerr error) {
@@ -183,10 +196,21 @@ func (c *tchannelClient) Call(
 
 func (c *tchannelCall) start() {
 	c.startTime = time.Now()
+	c.metrics.Sent.Inc(1)
 }
 
-func (c *tchannelCall) finish() {
+func (c *tchannelCall) finish(err error) {
 	c.finishTime = time.Now()
+
+	// emit metrics
+	if err != nil {
+		c.metrics.SystemErrors.Inc(1)
+	} else if !c.success {
+		c.metrics.AppErrors.Inc(1)
+	} else {
+		c.metrics.Success.Inc(1)
+	}
+	c.metrics.Latency.Record(c.finishTime.Sub(c.startTime))
 
 	// write logs
 	c.logger.Info("Finished an outgoing client TChannel request", c.logFields()...)
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/uber/tchannel-go"
-
 	"go.uber.org/thriftrw/protocol"
 	netContext "golang.org/x/net/context"
 )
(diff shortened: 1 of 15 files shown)


runtime/tchannel_interfaces.go
```diff
@@ -36,7 +36,7 @@ type RWTStruct interface {
 type TChannelClient interface {
 	// Call should be passed the method to call, headers and the request/response thriftrw structs.
 	// The arguments returned are (whether there was an application error, unexpected error)
-	Call(ctx context.Context, service, method string, reqHeaders map[string]string, req, resp RWTStruct) (success bool, respHeaders map[string]string, err error)
+	Call(ctx context.Context, clientMethod, thriftService, thriftMethod string, reqHeaders map[string]string, req, resp RWTStruct) (success bool, respHeaders map[string]string, err error)
```
ChuntaoLu: We should not add a `clientMethod` argument, `clientMethod` is a concept of upper level abstraction, and has nothing to do with the `Call` interface.

runtime/tchannel_client.go
```diff
@@ -97,59 +106,80 @@ func (c *tchannelClient) writeArgs(call *tchannel.OutboundCall, headers map[stri
...
-	var isOK bool
+func (c *tchannelClient) Call(ctx context.Context,
+	clientMethod, thriftService, thriftMethod string,
+	reqHeaders map[string]string,
+	req, resp RWTStruct,
+) (success bool, respHeaders map[string]string, err error) {
+	serviceMethod := fmt.Sprintf("%s::%s", thriftService, thriftMethod)
+	// TODO: Instead of creating method scope, add method call tags to subchannel stats reporter.
+	// Pending uber-go/tchannel PR to add custom tags to subchannel calls.
+	methodScope := c.scope.Tagged(map[string]string{
```
ChuntaoLu: This `methodScope` should be defined/subscoped in the outer layer, i.e, the client method definition, and accessed here, this way there is no need to pass around the `clientMethod` param.

runtime/tchannel_client.go
```diff
@@ -97,59 +106,80 @@ func (c *tchannelClient) writeArgs(call *tchannel.OutboundCall, headers map[stri
...
-
 		return false, nil, errors.Wrapf(err, "could not ensure arg2reader is empty for outbound call response: %s", c.serviceName)
 	}
 
-	if err := treader.Close(); err != nil {
+	if err = treader.Close(); err != nil {
 		return false, nil, errors.Wrapf(err, "could not close arg2reader for outbound call response: %s", c.serviceName)
 	}
 
-	success := !response.ApplicationError()
```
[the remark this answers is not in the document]
ravirajj: Done.

runtime/tchannel_client.go
```diff
@@ -97,59 +106,80 @@ func (c *tchannelClient) writeArgs(call *tchannel.OutboundCall, headers map[stri
...
-	treader, err = response.Arg3Reader()
-	if err != nil {
-		return success, headers, errors.Wrapf(err, "could not create arg3Reader for outbound call response: %s", c.serviceName)
+	if treader, err = response.Arg3Reader(); err != nil {
+		return false, respHeaders, errors.Wrapf(err, "could not create arg3Reader for outbound call response: %s", c.serviceName)
```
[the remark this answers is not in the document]
ravirajj: Fixing.

runtime/tchannel_client.go
ravirajj: If I create the tagged 'methodScope' in the client method definition, then I'll need to pass the `methodScope` to the `Call()` method instead of the `clientMethod` string.

runtime/tchannel_interfaces.go
ravirajj: See previous comment.

runtime/tchannel_client.go
ravirajj: Actually, on second thought, I'm going to collect the metrics in the generated client method, so we collect the complete client method latency, and no need to pass scope to TChannelClient.

runtime/tchannel_client.go
```diff
@@ -43,150 +47,352 @@ type tchannelClient struct {
...
+		)
 	}
 
-	twriter, err := call.Arg2Writer()
+	return call.success, call.resHeaders, err
+}
+
+func (c *tchannelCall) start() {
+	c.startTime = time.Now()
+	c.metrics.Sent.Inc(1)
```
Raynos: we have not sent it yet here. Need to move this elsewhere.

runtime/tchannel_client.go
```diff
@@ -43,150 +47,352 @@ type tchannelClient struct {
...
+	structWireValue, err := req.ToWire()
 	if err != nil {
-		return errors.Wrapf(err, "could not create arg3writer for outbound call %s: ", c.serviceName)
+		c.logger.Error("Could not write request for outbound request", zap.Error(err))
+		return errors.Wrapf(
+			err, "Could not write request for outbound %s.%s (%s %s) request",
+			c.client.clientID, c.methodName, c.client.serviceName, c.serviceMethod,
+		)
 	}
+	c.reqBody = structWireValue.GetBinary()
```
Raynos: `GetBinary()` does not do what you think it does. This `reqBody` is not useful here.

runtime/tchannel_client.go
```diff
@@ -43,150 +47,352 @@ type tchannelClient struct {
...
-		return success, headers, errors.Wrapf(err, "could not read outbound call response: %s", c.serviceName)
+	structWireValue, err := resp.ToWire()
+	if err != nil {
+		c.logger.Error("Could not serialize response for outbound response", zap.Error(err))
+		return errors.Wrapf(
+			err, "Could not serialize response for outbound %s.%s (%s %s) response",
+			c.client.clientID, c.methodName, c.client.serviceName, c.serviceMethod,
+		)
 	}
+	c.resBody = structWireValue.GetBinary()
```
Raynos: again `GetBinary()` does not do what you think it does. We can just not log the reqBody and resBody for now.

runtime/tchannel_client.go
ravirajj: Removing `reqBody`.

runtime/tchannel_client.go
ravirajj: Removing `resBody`.

test/endpoints/tchannel/baz/baz_simpleservice_method_call_test.go
```diff
@@ -42,6 +42,9 @@ func TestCallTChannelSuccessfulRequestOKResponse(t *testing.T) {
 		KnownTChannelBackends: []string{"baz"},
 		TestBinary:            util.DefaultMainFile("example-gateway"),
 		ConfigFiles:           util.DefaultConfigFiles("example-gateway"),
+		TChannelClientMethods: map[string]string{
```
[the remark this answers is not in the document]
ravirajj: It was needed to avoid nil pointer exceptions for client method loggers in tests. I can default to a client logger if the method logger is not present.

runtime/tchannel_client.go
ravirajj: Will do.

test/endpoints/tchannel/baz/baz_simpleservice_method_call_test.go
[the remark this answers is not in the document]
ravirajj: I can't avoid passing the tchannel client methods as part of the test gateway options, since the tchannel client is created/called on the test gateway (not actual gateway), so the test gateway needs know the client methods when setting up the method logger and metrics for the tchannel client.

ravirajj:
Benchmark results:
```
#
#   benchmark output rj.tchannelclient
#

Running 30s test @ http://localhost:8093/baz/call
    Latency    29.98ms   18.16ms 197.10ms   67.22%
    Req/Sec     1.10k   180.29     2.26k    79.03%
  396214 requests in 30.07s, 43.45MB read
  Socket errors: connect 0, read 244, write 0, timeout 0
Requests/sec:  13177.03
Transfer/sec:      1.45MB

#
#   benchmark output master
#

Running 30s test @ http://localhost:8093/baz/call
    Latency    31.51ms   19.55ms 189.58ms   67.53%
    Req/Sec     1.07k   162.83     1.76k    74.11%
  382576 requests in 30.07s, 41.96MB read
  Socket errors: connect 0, read 254, write 0, timeout 0
Requests/sec:  12723.21
Transfer/sec:      1.40MB
```

Raynos:
Can you explain why its faster.

ravirajj:
Produced flamegraphs for `master` and `rj.tchannelclient`, but didn't any significant difference.

PR #199, opened 2017-09-05
Adding metrics and logs to tchannel server

ravirajj:
TChannel endpoint emits the following metrics:

Names:
- inbound.calls.sent
- inbound.calls.success
- inbound.calls.app-errors
- inbound.calls.system-errors
- inbound.calls.latency

Tags:
- host - “all”
- endpoint - endpointId, e.g. “bar”
- handler - handleId, e.g. “normal”


runtime/tchannel_server.go
```diff
@@ -74,16 +76,29 @@ func NewTChannelRouter(registrar tchannel.Registrar, logger *zap.Logger) *TChann
 
 // Register registers the given TChannelHandler to be called on an incoming call for its method.
 // "service" is the thrift service name as in the thrift definition.
-func (s *TChannelRouter) Register(service string, method string, h TChannelHandler) {
-	handler := &handler{tchannelHandler: h}
+func (s *TChannelRouter) Register(
+	service, method string,
+	h TChannelHandler,
+	scope tally.Scope,
```
ChuntaoLu: The semantics here is messy, with `Register(service string, method string, h TChannelHandler)` it's clear that I register a handler for the given service and method. Scope should made accessible from the `TChannelHandler`, and you could do that by adding a `Scope()` method to the `TChannelHandler` interface.

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -45,6 +46,14 @@ func (h *{{$handlerName}}) Register(g *zanzibar.Gateway) error {
 		"{{.ThriftService}}",
 		"{{.Name}}",
 		h,
+		g.AllHostScope.Tagged(map[string]string{
```
ChuntaoLu: Instead of passing around the scope, it's better to define scope on the [handler](https://github.com/uber/zanzibar/pull/199/files#diff-7c8f7ec39bf077182fa0b404b3d49d76R38) and initialize it when a new handler is created.

runtime/tchannel_server.go
ravirajj: Yes, I agree. Adding `Scope()` method to `TChannelHandler` interface would be better.

runtime/tchannel_server.go
```diff
@@ -95,26 +110,47 @@ func (s *TChannelRouter) register(service string, method string, h *handler) {
...
 	s.RLock()
-	handler, ok := s.handlers[op]
+	handler, ok = s.handlers[op]
 	s.RUnlock()
 	if !ok {
 		s.logger.Error(fmt.Sprintf("Handle got call for %s which is not registered", op))
 	}
 
-	if err := s.handle(ctx, handler, service, method, call); err != nil {
+	if success, err = s.handle(ctx, handler, service, method, call, startTime); err != nil {
```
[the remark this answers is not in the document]
ravirajj: Good catch, `startTime` should have been removed, but `!success` is used to emit the `app-errors` metrics.

runtime/tchannel_server.go
```diff
@@ -131,49 +167,48 @@ func (s *TChannelRouter) onError(err error) {
...
 }
 
-func (s *TChannelRouter) handle(ctx context.Context, handler handler, service string, method string, call *tchannel.InboundCall) error {
+func (s *TChannelRouter) handle(
+	ctx context.Context,
+	handler handler,
+	service, method string,
+	call *tchannel.InboundCall,
+	startTime time.Time,
+) (success bool, err error) {
```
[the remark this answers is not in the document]
ravirajj: See previous comment.

runtime/tchannel_server.go
```diff
@@ -182,72 +217,60 @@ func (s *TChannelRouter) handle(ctx context.Context, handler handler, service st
...
 	}
-
 	if err := twriter.Close(); err != nil {
-		return errors.Wrapf(err, "could not close arg3writer for inbound call response: %s::%s", service, method)
+		return false, errors.Wrapf(err, "could not close arg3writer for inbound call response: %s::%s", service, method)
 	}
 
-	return nil
+	if !success {
+		err = call.Response().SetApplicationError()
```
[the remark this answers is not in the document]
ravirajj: Moving it back.

codegen/templates/tchannel_endpoint.tmpl
ravirajj: Yes, will do.

ravirajj:
Also thinking we could move the metrics collection to be within the `TChannelHandler.Handle()` method, so we don't have to expose the `Scope()` and `Logger()` methods to `TChannelRouter.handle()`.  But then we will miss collecting the latency overhead of `TChannelRouter.handle()`, or use a handler scoped logger in it.

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -100,6 +116,7 @@ func (h *{{$handlerName}}) Handle(
 
 	{{if eq (len .Exceptions) 0 -}}
 		if err != nil {
+			h.Logger.Error("Handler returned error", zap.Error(err))
```
Raynos: not an `Error()` ; do we log about this anywhere else ?

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -109,15 +126,24 @@ func (h *{{$handlerName}}) Handle(
 			{{$method := .Name -}}
 			{{range .Exceptions -}}
 				case *{{.Type}}:
+					h.endpoint.Logger.Error(
```
Raynos: this log needs to be inside the if statement.

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -109,15 +126,24 @@ func (h *{{$handlerName}}) Handle(
...
+						return false, nil, resHeaders, errors.Errorf(
+							"%s.%s (%s) handler returned non-nil error type *{{.Type}} but nil value",
+							h.endpoint.EndpointID, h.endpoint.HandlerID, h.endpoint.Method,
 						)
 					}
 					res.{{title .Name}} = v
 			{{end -}}
 				default:
-					return false, nil, resHeaders, err
+					h.endpoint.Logger.Error("Handler returned error", zap.Error(err))
```
Raynos: warning.

runtime/router.go
```diff
@@ -53,21 +53,21 @@ type RouterEndpoint struct {
...
-	handlerName string,
+	endpointID string,
+	handlerID string,
 	handler HandlerFn,
 ) *RouterEndpoint {
 	logger := gateway.Logger.With(
-		zap.String("endpointID", endpointName),
-		zap.String("handlerID", handlerName),
+		zap.String("endpointID", endpointID),
+		zap.String("HandlerID", handlerID),
```
Raynos: please stop renaming stuff when it does not need to change thanks. `handlerID` not `HandlerID`

runtime/tchannel_server.go
```diff
@@ -64,197 +83,339 @@ func (ncr netContextRouter) Handle(ctx netContext.Context, call *tchannel.Inboun
...
+func (s *TChannelRouter) handle(
+	ctx context.Context,
+	e *TChannelEndpoint,
+	call *tchannel.InboundCall,
+) (err error) {
+	c := tchannelInboundCall{
+		endpoint: e,
+		call:     call,
 	}
+	defer func() { c.finish(err) }()
```
Raynos: this is not correct really. we do not always write to the body in this method.

For example we can `return err` and have the caller write an error frame on our behalf

So `finish(err)` happesn in this method only if we call `c.writeResBody()` otherwise `finish(err)` should be called upstream in the caller of this method.

runtime/tchannel_server.go
```diff
@@ -64,197 +83,339 @@ func (ncr netContextRouter) Handle(ctx netContext.Context, call *tchannel.Inboun
...
-	tracer := tchannel.TracerFromRegistrar(s.registrar)
-	ctx = tchannel.ExtractInboundSpan(ctx, call, headers, tracer)
+func (c *tchannelInboundCall) finish(err error) {
+	c.finishTime = time.Now()
 
-	wireValue, err := protocol.Binary.Decode(bytes.NewReader(buf.Bytes()), wire.TStruct)
+	// emit metrics
 	if err != nil {
-		return errors.Wrapf(err, "could not decode arg3 for inbound call: %s::%s", service, method)
+		c.endpoint.metrics.SystemErrors.Inc(1)
```
Raynos: There are other system errors. We are going to really lose metrics here.

For example tchannel handles timeouts internally. I strongly suspects that if the method times out we do not capture this properly in `finish()`

runtime/tchannel_server.go
```diff
@@ -64,197 +83,339 @@ func (ncr netContextRouter) Handle(ctx netContext.Context, call *tchannel.Inboun
...
+		_ = treader.Close()
+		c.endpoint.Logger.Error("Could not read from arg3reader for inbound request", zap.Error(err))
+		err = errors.Wrapf(err, "Could not read from arg3reader for inbound %s.%s (%s) request",
+			c.endpoint.EndpointID, c.endpoint.HandlerID, c.endpoint.Method,
+		)
+		return
+	}
+	wireValue, err = protocol.Binary.Decode(bytes.NewReader(buf.Bytes()), wire.TStruct)
+	if err != nil {
+		c.endpoint.Logger.Error("Could not decode arg3 for inbound request", zap.Error(err))
```
Raynos: there are so many `Error()` here that need to be downgraded to `Warn()`

Rule of thumb : 
 - If we are dealing with network input or any input foreign to this process then it's a `Warn()`
 - If we are dealing with the application developer passing the wrong arguments or doing the wrong thing ( aka its a bug in my own application ) then it's an `Error()` 
 - If we are dealing with writing out of the process ; aka writes to the TCP socket that can fail then again its something foreign to the process or application developer and its a `Warn()`

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -59,7 +58,14 @@ func (h *{{$handlerName}}) Handle(
 	wfReqHeaders := zanzibar.ServerTChannelHeader(reqHeaders)
 	{{if .ReqHeaders -}}
 	if err := wfReqHeaders.Ensure({{.ReqHeaders | printf "%#v" }}); err != nil {
-		return false, nil, nil, err
+		h.endpoint.Logger.Error("Request missing request headers",
```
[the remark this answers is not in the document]
ravirajj: Moved log to `ServerTChannelHeader.Ensure()`.

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -68,13 +74,17 @@ func (h *{{$handlerName}}) Handle(
 	{{if ne .RequestType "" -}}
 	var req {{unref .RequestType}}
 	if err := req.FromWire(*wireValue); err != nil {
-		return false, nil, nil, err
+		h.endpoint.Logger.Error("Error converting request from wire", zap.Error(err))
```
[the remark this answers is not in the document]
ravirajj: Changed to Warning.  

req.FromWire is generated by thriftrw. I don't think I can inject the logger into the generated req struct.

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -89,7 +99,13 @@ func (h *{{$handlerName}}) Handle(
 
 	{{- if .ResHeaders}}
 	if err := wfResHeaders.Ensure({{.ResHeaders | printf "%#v" }}); err != nil {
-		return false, nil, nil, err
+		h.endpoint.Logger.Error("Request missing response headers", zap.Error(err),
```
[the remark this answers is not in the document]
ravirajj: Done.

codegen/templates/tchannel_endpoint.tmpl
ravirajj: Changed to warning.  This is the only place where we log errors returned by `workflow.Handle()`.  The `workflow.Handle()` function itself is provided by the developer, so may or may not log the error it returns.

codegen/templates/tchannel_endpoint.tmpl
ravirajj: This should be under the `case` since it logs the appropriate `.Type` error message.

codegen/templates/tchannel_endpoint.tmpl
ravirajj: Done.

runtime/router.go
ravirajj: Done.

runtime/tchannel_server.go
ravirajj: To me it makes sense that reading or writing from/to the network should be an application error (It's the server's fault it couldn't read or write to the network).  If the wrong thing (i.e. missing headers, wrong body fields/format, etc.) was sent by the foreign caller then that's a client error or application warning.

runtime/tchannel_server.go
ravirajj: I'm not sure what you mean?  The `defer` function is always called where ever the `handle()` function returns, on any of the return statements. So `start()` and `finish()` are always called at the beginning and end of the `handle()` function, regardless of `c.writeResBody()` succeeding or not.

Raynos:
So after talking to @akshayjshah we can listen for `ctx.Done` and call `finish()` to mark the request as failed when the context is done ( aka assume tchannel-go timed it out )

cc @ravirajj

runtime/tchannel_server.go
ravirajj: Looking into `ctx.Done()` to capture timeout errors.

ravirajj:
@akshayjshah How do I distinguish between the ctx being canceled due to SystemError or when the response completes?

Where is what I am doing:
```
func (s *TChannelRouter) Handle(ctx context.Context, call *tchannel.InboundCall) {
	var err error
	errc := make(chan error, 1)
	c := tchannelInboundCall{
		endpoint: e,
		call:     call,
	}
	c.start()
	go func() { errc <- s.handle(ctx, e, &c) }()
	select {
	case <-ctx.Done():
		err = ctx.Err()
	case err = <-errc:
	}
	c.finish(err)
}

func (c *tchannelInboundCall) finish(err error) {
	c.finishTime = time.Now()

	// emit metrics
	// *** How to distinguish cancel error from  SystemError (error) and response complete (success)
	if err != nil {
		c.endpoint.metrics.SystemErrors.Inc(1)
	} else if !c.success {
		c.endpoint.metrics.AppErrors.Inc(1)
	} else {
		c.endpoint.metrics.Success.Inc(1)
	}
	c.endpoint.metrics.Latency.Record(c.finishTime.Sub(c.startTime))

	// write logs
	if err != nil {
		if err == context.Canceled || err == context.DeadlineExceeded ||
			tchannel.GetSystemErrorCode(err) == tchannel.ErrCodeTimeout {
			c.endpoint.Logger.Warn("Thrift server timeout", zap.Error(err))
		} else {
			c.endpoint.Logger.Error("Thrift server error", zap.Error(err))
		}
	}
	c.endpoint.Logger.Info("Finished an incoming server TChannel request", c.logFields()...)

}
```

runtime/tchannel_logger.go
```diff
@@ -119,3 +121,20 @@ func (l TChannelLogger) WithFields(fields ...tchannel.LogField) tchannel.Logger
...
 	return TChannelLogger{logger: l.logger.With(zfields...)}
 }
+
+// LogError logs warnings for timeout errors
+func LogError(logger *zap.Logger, err error, msg string) {
+	OnError(logger, err, msg, msg)
+}
+
+// OnError logs warnings for timeout errors
+func OnError(logger *zap.Logger, err error, warnMsg, errMsg string) {
```
Raynos: Add a `TODO` here. We want to improve the classification of errors to be something like : 

https://github.com/uber/tchannel-node/blob/master/errors.js#L907-L930

examples/example-gateway/endpoints/tchannel/baz/baz_call.go
```diff
@@ -63,9 +61,10 @@ func (w CallEndpoint) Handle(
 		case *clientBaz.AuthErr:
 			return respHeaders, (*endpointBaz.AuthErr)(v)
 		default:
-			w.Logger.Error(
+			zanzibar.OnError(
```
Raynos: nitpick: `zanzibar.LogError()`

Raynos: Oh I see ; maybe call this method `LogErrorWarn()` ?

runtime/tchannel_server.go
```diff
@@ -64,197 +84,364 @@ func (ncr netContextRouter) Handle(ctx netContext.Context, call *tchannel.Inboun
...
+		c.endpoint.metrics.Success.Inc(1)
 	}
+	c.endpoint.metrics.Latency.Record(c.finishTime.Sub(c.startTime))
 
-	buf := GetBuffer()
-	defer PutBuffer(buf)
-	if _, err := buf.ReadFrom(treader); err != nil {
-		_ = treader.Close()
+	// write logs
+	OnError(c.endpoint.Logger, err, "Thrift server timeout", "Thrift server error")
```
Raynos: nitpick: longer names

`Got an error whilst finishing a server TChannel request `

We can use the same message for both warn / error and then use the fields on the log to filter out timeouts.

runtime/tchannel_server.go
```diff
@@ -64,197 +84,364 @@ func (ncr netContextRouter) Handle(ctx netContext.Context, call *tchannel.Inboun
...
+		LogError(c.endpoint.Logger, err, "Could not close arg3reader for inbound request")
+		err = errors.Wrapf(err, "Could not close arg3reader for inbound %s.%s (%s) request",
+			c.endpoint.EndpointID, c.endpoint.HandlerID, c.endpoint.Method,
+		)
+		return
+	}
 
-		_ = call.Response().SendSystemError(errors.New("Server Error"))
+	// request received when arg3reader is closed
+	c.endpoint.metrics.Recvd.Inc(1)
```
Raynos: the request is recvd earlier. We return an error above this which should return an `err` via `handle()` and thus report an error in `finish()`

I think we want to emit recvd in `start()`

ravirajj: Moving to back to where it was originally.

runtime/tchannel_server.go
ravirajj: I wasn't sure if you wanted to maintain parity with https://github.com/uber/tchannel-go/blob/dev/thrift/server.go#L97, but I also prefer having the same msg for both warn and error.  We don't need the `onError()` function then, `logErrors` will suffice.

examples/example-gateway/endpoints/tchannel/baz/baz_call.go
ravirajj: Sure calling it `LogErrorWarn`.

runtime/tchannel_logger.go
ravirajj: Done.

PR #200, opened 2017-09-05
write testFixtures from endpoint config to `_test.json` file

cl1337:
first write testfixtures to _test.json file

### DIFF
--- a/backend/repository/config.go
+++ b/backend/repository/config.go
@@ -65,10 +65,41 @@ type EndpointConfig struct {
 	WorkflowType     string                    `json:"workflowType"`
 	ClientID         string                    `json:"clientID"`
 	ClientMethod     string                    `json:"clientMethod"`
-	TestFixtures     []string                  `json:"testFixtures"`
+	TestFixtures     []*TestStub                  `json:"testFixtures"`
 	Middlewares      []*codegen.MiddlewareSpec `json:"middlewares"`
 	ReqHeaderMap     map[string]string         `json:"reqHeaderMap"`
 	ResHeaderMap     map[string]string         `json:"resHeaderMap"`
+}
+
+// TestStub saves stubbed requests/responses for an endpoint test.
+type TestStub struct {
+	TestName               string `json:"testName"`
+	EndpointID             string `json:"endpointId"`
+	HandlerID              string `json:"handlerId"`
+	EndpointRequest        map[string]interface{} `json:"endpointRequest"`
+	EndpointRequestString  string `json:"endpointRequestString"`
+	EndpointReqHeaders     map[string]string      `json:"endpointReqHeaders"`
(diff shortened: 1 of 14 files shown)


backend/repository/config.go
```diff
@@ -65,12 +65,43 @@ type EndpointConfig struct {
...
 	ClientMethod     string                    `json:"clientMethod"`
-	TestFixtures     []string                  `json:"testFixtures"`
+	TestFixtures     []*TestStub                  `json:"testFixtures"`
 	Middlewares      []*codegen.MiddlewareSpec `json:"middlewares"`
 	ReqHeaderMap     map[string]string         `json:"reqHeaderMap"`
 	ResHeaderMap     map[string]string         `json:"resHeaderMap"`
 }
 
+// TestStub saves stubbed requests/responses for an endpoint test.
+type TestStub struct {
```
ChuntaoLu: Is this the same as https://github.com/uber/zanzibar/blob/master/codegen/module_system.go#L75 except the json tags? If so, why duplicate?

cl1337: yes and duplicated on purpose
1, there should be one testFixture struct in the code repo, a user passed in test case, either structured JSON (file) or go struct should have ubiquitous form
2, I'm not convinced testFixture/testStub struct we use in backend repo & runtime makes sense
-> refactor all testFixtures used in this repo with a redesigned struct that:
1, use maps instead of string list, to facilitate user edit from UI
2, use a better request / response struct, like https://github.com/uber/zanzibar/blob/master/backend/repository/handler_test.go#L44

this will be captured in another change to limit the scope

backend/repository/config.go
```diff
@@ -65,12 +65,43 @@ type EndpointConfig struct {
...
+	TestName               string `json:"testName"`
+	EndpointID             string `json:"endpointId"`
+	HandlerID              string `json:"handlerId"`
+	EndpointRequest        map[string]interface{} `json:"endpointRequest"`
```
zhenghuiwang: Seems that you can define `Request` type for both Endpoint and client, including the json body/ string body and header.

backend/repository/config.go
```diff
@@ -65,12 +65,43 @@ type EndpointConfig struct {
...
+	EndpointRequestString  string `json:"endpointRequestString"`
+	EndpointReqHeaders     map[string]string      `json:"endpointReqHeaders"`
+	EndpointReqHeaderKeys  []string               `json:"endpointReqHeaderKeys"`
```
zhenghuiwang: Why headerKeys are needed as you already have header map?

backend/repository/config.go
cl1337: put this struct here not because it makes completely sense but since it's used in module system, we might(will have to) need to refactor how test fixtures are used & stored

backend/repository/config.go
cl1337: same above

PR #201, opened 2017-09-05
Adding metrics and logs to server http request/response

ravirajj:
HTTP endpoint emits the following metrics:

Names:
- inbound.calls.latency
- inbound.calls.recvd
- inbound.calls.success
- inbound.calls.errors
- inbound.calls.status.XXX

Tags:
- host - “all”
- endpoint - endpointId, e.g. “bar”
- handler - handleId, e.g. “normal”

### DIFF
--- a/runtime/call_metrics.go
+++ b/runtime/call_metrics.go
@@ -1,0 +1,252 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zanzibar
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/uber-go/tally"
+)
+
+const (
+	inboundCallsRecvd        = "inbound.calls.recvd"
+	inboundCallsLatency      = "inbound.calls.latency"
+	inboundCallsSuccess      = "inbound.calls.success"
+	inboundCallsAppErrors    = "inbound.calls.app-errors"
+	inboundCallsSystemErrors = "inbound.calls.system-errors"
+	inboundCallsErrors       = "inbound.calls.errors"
+	inboundCallsStatus       = "inbound.calls.status"
+
+	outboundCallsSent         = "outbound.calls.sent"
+	outboundCallsLatency      = "outbound.calls.latency"
+	outboundCallsSuccess      = "outbound.calls.success"
+	outboundCallsAppErrors    = "outbound.calls.app-errors"
+	outboundCallsSystemErrors = "outbound.calls.system-errors"
+	outboundCallsErrors       = "outbound.calls.errors"
+	outboundCallsStatus       = "outbound.calls.status"
+)
+
+var knownStatusCodes = map[int]bool{
+	http.StatusContinue:                      true, // 100
+	http.StatusSwitchingProtocols:            true, // 101
+	http.StatusProcessing:                    true, // 102
+	http.StatusOK:                            true, // 200
+	http.StatusCreated:                       true, // 201
+	http.StatusAccepted:                      true, // 202
+	http.StatusNonAuthoritativeInfo:          true, // 203
+	http.StatusNoContent:                     true, // 204
+	http.StatusResetContent:                  true, // 205
+	http.StatusPartialContent:                true, // 206
+	http.StatusMultiStatus:                   true, // 207
+	http.StatusAlreadyReported:               true, // 208
+	http.StatusIMUsed:                        true, // 226
+	http.StatusMultipleChoices:               true, // 300
+	http.StatusMovedPermanently:              true, // 301
+	http.StatusFound:                         true, // 302
+	http.StatusSeeOther:                      true, // 303
+	http.StatusNotModified:                   true, // 304
+	http.StatusUseProxy:                      true, // 305
+	http.StatusTemporaryRedirect:             true, // 307
+	http.StatusPermanentRedirect:             true, // 308
+	http.StatusBadRequest:                    true, // 400
+	http.StatusUnauthorized:                  true, // 401
+	http.StatusPaymentRequired:               true, // 402
+	http.StatusForbidden:                     true, // 403
+	http.StatusNotFound:                      true, // 404
+	http.StatusMethodNotAllowed:              true, // 405
+	http.StatusNotAcceptable:                 true, // 406
+	http.StatusProxyAuthRequired:             true, // 407
+	http.StatusRequestTimeout:                true, // 408
+	http.StatusConflict:                      true, // 409
+	http.StatusGone:                          true, // 410
+	http.StatusLengthRequired:                true, // 411
+	http.StatusPreconditionFailed:            true, // 412
+	http.StatusRequestEntityTooLarge:         true, // 413
+	http.StatusRequestURITooLong:             true, // 414
+	http.StatusUnsupportedMediaType:          true, // 415
+	http.StatusRequestedRangeNotSatisfiable:  true, // 416
+	http.StatusExpectationFailed:             true, // 417
+	http.StatusTeapot:                        true, // 418
+	http.StatusUnprocessableEntity:           true, // 422
+	http.StatusLocked:                        true, // 423
+	http.StatusFailedDependency:              true, // 424
+	http.StatusUpgradeRequired:               true, // 426
+	http.StatusPreconditionRequired:          true, // 428
+	http.StatusTooManyRequests:               true, // 429
+	http.StatusRequestHeaderFieldsTooLarge:   true, // 431
+	http.StatusUnavailableForLegalReasons:    true, // 451
+	http.StatusInternalServerError:           true, // 500
+	http.StatusNotImplemented:                true, // 501
+	http.StatusBadGateway:                    true, // 502
+	http.StatusServiceUnavailable:            true, // 503
+	http.StatusGatewayTimeout:                true, // 504
+	http.StatusHTTPVersionNotSupported:       true, // 505
+	http.StatusVariantAlsoNegotiates:         true, // 506
+	http.StatusInsufficientStorage:           true, // 507
+	http.StatusLoopDetected:                  true, // 508
+	http.StatusNotExtended:                   true, // 510
+	http.StatusNetworkAuthenticationRequired: true, // 511
+}
+
+type inboundMetrics struct {
+	Recvd tally.Counter // inbound.calls.recvd
+}
+
+type outboundMetrics struct {
+	Sent tally.Counter // outbound.calls.sent
+}
+
+type commonMetrics struct {
+	Latency tally.Timer   // [inbound|outbound].calls.latency
+	Success tally.Counter // [inbound|outbound].calls.success
+}
+
+type tchannelMetrics struct {
+	commonMetrics
+	AppErrors    tally.Counter // [inbound|outbound].calls.app-errors
+	SystemErrors tally.Counter // [inbound|outbound].calls.system-errors
+}
+
+type httpMetrics struct {
+	commonMetrics
+	Errors tally.Counter         // [inbound|outbound].calls.errors
+	Status map[int]tally.Counter // [inbound|outbound].calls.status.XXX
+}
+
+// InboundHTTPMetrics ...
+type InboundHTTPMetrics struct {
+	inboundMetrics
+	httpMetrics
+}
+
+// InboundTChannelMetrics ...
+type InboundTChannelMetrics struct {
+	inboundMetrics
+	tchannelMetrics
+}
+
+// OutboundHTTPMetrics ...
+type OutboundHTTPMetrics struct {
+	outboundMetrics
+	httpMetrics
+}
+
+// OutboundTChannelMetrics ...
+type OutboundTChannelMetrics struct {
+	outboundMetrics
+	tchannelMetrics
+}
+
+// NewInboundHTTPMetrics returns inbound HTTP metrics
+func NewInboundHTTPMetrics(scope tally.Scope) *InboundHTTPMetrics {
+	metrics := InboundHTTPMetrics{}
+	metrics.Recvd = scope.Counter(inboundCallsRecvd)
+	metrics.Latency = scope.Timer(inboundCallsLatency)
+	metrics.Success = scope.Counter(inboundCallsSuccess)
+	metrics.Errors = scope.Counter(inboundCallsErrors)
+	metrics.Status = make(map[int]tally.Counter, len(knownStatusCodes))
+	for statusCode := range knownStatusCodes {
+		metrics.Status[statusCode] = scope.Counter(fmt.Sprintf("%s.%d", inboundCallsStatus, statusCode))
+	}
+	return &metrics
+}
+
+// TODO: Add remaining call metrics
+
+// NewInboundTChannelMetrics returns inbound TChannel metrics
+//func NewInboundTChannelMetrics(scope tally.Scope) *InboundTChannelMetrics {
+//	metrics := InboundTChannelMetrics{}
+//	metrics.Recvd = scope.Counter(inboundCallsRecvd)
+//	metrics.Latency = scope.Timer(inboundCallsLatency)
+//	metrics.Success = scope.Counter(inboundCallsSuccess)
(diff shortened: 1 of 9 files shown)


runtime/router.go
```diff
@@ -133,47 +120,60 @@ func NewRouterEndpoint(
...
+
+		scope := endpoint.scope.Tagged(map[string]string{
+			"status-code": strconv.Itoa(req.res.StatusCode),
+			//"calling-service": "unknown", // TODO: Add 'calling-service' tag to scope
+		})
+		scope.Counter("inbound.calls.recvd").Inc(1)
+		if req.res.StatusCode >= 400 && req.res.StatusCode < 600 {
+			scope.Counter("inbound.calls.errors").Inc(1)
+			if req.res.StatusCode >= 400 && req.res.StatusCode < 500 {
+				scope.Counter("inbound.calls.status.4XX").Inc(1)
```
ChuntaoLu: nit: assign const vars for strings like "inbound.calls.status.4XX"

codegen/templates/endpoint.tmpl
```diff
@@ -87,6 +88,14 @@ func (handler *{{$handlerName}}) Register(g *zanzibar.Gateway) error {
 			{{- else -}}
 			handler.HandleRequest,
 			{{- end}}
+			g.AllHostScope.Tagged(map[string]string{
```
ChuntaoLu: Not a fan of creating and passing a subscope here, it's subscoped again at https://github.com/uber/zanzibar/pull/201/files#diff-7aed2fb9afc6ff1780de1ca94cd0a938R155. Why not pass down the tags info and create proper subscope once?

ravirajj: It's subscoped with the `status-code` tag, which is only known when handling the request.  Instead, I'll need to pre-allocate scopes and counters for each status-code.

runtime/router.go
```diff
@@ -133,47 +120,60 @@ func NewRouterEndpoint(
...
+				zap.Int("UnexpectedStatusCode", req.res.StatusCode),
+			)
+		}
+
+		scope := endpoint.scope.Tagged(map[string]string{
+			"status-code": strconv.Itoa(req.res.StatusCode),
+			//"calling-service": "unknown", // TODO: Add 'calling-service' tag to scope
+		})
+		scope.Counter("inbound.calls.recvd").Inc(1)
+		if req.res.StatusCode >= 400 && req.res.StatusCode < 600 {
```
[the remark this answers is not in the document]
ravirajj: Done.

runtime/router.go
ravirajj: Done.

ravirajj:
@Raynos I won't be able to add a `calling-service` tag to the pre-allocated metrics, since it is only know at the handler call time, not when the handler is created.  We could get the calling-service value from jaeger tracing.  Any suggestions on how to add calling-service to the metrics?

codegen/templates/endpoint.tmpl
```diff
@@ -60,6 +64,22 @@ func New{{$handlerName}}(
 ) *{{$handlerName}} {
 	return &{{$handlerName}}{
 		Clients: deps.Client,
+		Logger: gateway.Logger.With(
```
Raynos: Let's avoid doing this at code generation. Let's do this at `runtime.NewRouterEndpoint()` instead.

runtime/server_http_request.go
```diff
@@ -376,13 +338,9 @@ func (req *ServerHTTPRequest) CheckQueryValue(key string) bool {
 
 	values := req.queryValues[key]
 	if len(values) == 0 {
-		req.Logger.Warn("Got request with missing query string value",
-			zap.String("expectedKey", key),
-		)
+		req.Logger.Warn("Got request with missing query string value", zap.String("expectedKey", key))
```
Raynos: stop changing whitespace. stick to 80 characters or less.

runtime/router.go
```diff
@@ -108,72 +40,50 @@ type HandlerFn func(
...
-	reqFields := logRequestFields(r)
+	startTime := time.Now()
+	var req *ServerHTTPRequest
 	var resFields []zapcore.Field
+	reqFields := logRequestFields(r)
 
 	defer func() {
-		writeLogs(endpoint.gateway.Logger, reqFields, resFields)
+		writeLogs(req.Logger, reqFields, resFields)
+		endpoint.metrics.CollectMetrics(startTime, req.res.StatusCode)
```
Raynos: We do not want to do metrics like this. Makes the code too difficult to follow. please revert to emitting metrics inline.

codegen/templates/endpoint.tmpl
```diff
@@ -51,6 +53,8 @@ import (
 // {{$handlerName}} is the handler for "{{.HTTPPath}}"
 type {{$handlerName}} struct {
 	Clients *module.ClientDependencies
+	Logger  *zap.Logger
```
Raynos: Remove this logger change from this PR. This needs to be a seperate PR.

codegen/templates/endpoint.tmpl
```diff
@@ -60,6 +60,14 @@ func New{{$handlerName}}(
 ) *{{$handlerName}} {
 	return &{{$handlerName}}{
 		Clients: deps.Client,
+		Scope: gateway.AllHostScope.Tagged(map[string]string{
```
Raynos: Can we not allocate a scope in the template ? 

We can allocate one in runtime/ instead. we don't really need `thrift-method` or `workflow-type`

runtime/router.go
```diff
@@ -108,103 +44,70 @@ type HandlerFn func(
...
 ) {
-	reqFields := logRequestFields(r)
-	var resFields []zapcore.Field
-
+	var req *ServerHTTPRequest
 	defer func() {
-		writeLogs(endpoint.gateway.Logger, reqFields, resFields)
+		// Post-process logs & metrics
+		writeLogs(endpoint.gateway.Logger, logRequestFields(r), logResponseFields(req.Res))
+		endpoint.metrics.CollectMetrics(req.StartTime, req.Res.FinishTime, req.Res.StatusCode)
```
Raynos: Deferring the collection of metrics loses information.

We really want to emit the metrics in the right place at the start and finish methods of the request and response class.

ravirajj: We don't loose the metrics in the defer.  Defer is also run at the end of the function call.

codegen/templates/endpoint.tmpl
ravirajj: Sure I can remove the extra tags: `module`, `module-type`, `thrift-method`, `workflow-type`.  They were there for completeness, the I can remove if not necessary.

runtime/server_http_response.go
```diff
@@ -33,81 +32,61 @@ import (
 // ServerHTTPResponse struct manages request
 type ServerHTTPResponse struct {
```
Raynos: please do not make these private fields public. It allows consumers to mess with too much of the state.

runtime/server_http_response.go
```diff
@@ -33,81 +32,61 @@ import (
...
 	responseWriter    http.ResponseWriter
-	Request           *ServerHTTPRequest
-	gateway           *Gateway
-	finishTime        time.Time
-	finished          bool
-	metrics           *EndpointMetrics
+	Req               *ServerHTTPRequest
```
Raynos: Renaming `Request` to `Req` is breaking change. It's not worth it.

runtime/server_http_request.go
```diff
@@ -30,31 +30,24 @@ import (
...
 )
 
 // ServerHTTPRequest struct manages request
 type ServerHTTPRequest struct {
 	httpRequest *http.Request
-	res         *ServerHTTPResponse
-	gateway     *Gateway
+	Res         *ServerHTTPResponse
 	started     bool
-	startTime   time.Time
```
Raynos: please keep these fields private.

ravirajj: Sure.  I can keep them private.  I made them public since they were accessed by other runtime classes, but they are all under the same package, so should be fine to make them private.

runtime/server_http_response.go
ravirajj: Trying to be consistent in the naming.  Why is this a breaking change?

runtime/server_http_response.go
ravirajj: Sure, no problem.

runtime/router.go
```diff
@@ -134,77 +58,60 @@ func NewRouterEndpoint(
...
-	statusCodes := make(map[int]tally.Counter, len(knownStatusCodes))
-
-	for _, statusCode := range knownStatusCodes {
-		metricName := "inbound.calls.status." + strconv.Itoa(statusCode)
-		statusCodes[statusCode] = endpointScope.Counter(metricName)
-	}
-
+	})
 	return &RouterEndpoint{
-		EndpointName: endpointName,
```
srivastavankit: is the endpoint name already added to the logger?

runtime/router.go
```diff
@@ -250,30 +158,48 @@ func (router *HTTPRouter) Register(
...
+		writeLogs(router.logger, logRequestFields(r), resFields)
+		router.metrics[notFound].CollectMetrics(startTime, finishTime, http.StatusNotFound)
+	}()
+
+	startTime = time.Now()
 	// TODO custom NotFound
-	// A NotFound request is not started...
-	// TODO: inc.finish()
 	http.NotFound(w, r)
+	finishTime = time.Now()
```
srivastavankit: so this latency is only monitoring how much time it takes to write a response?

runtime/router.go
ravirajj: That will be done in a separate PR for logger changes.

runtime/router.go
ravirajj: Yes, amount of time between read/write to/from wire.

runtime/router.go
srivastavankit: Then wondering if you want to remove endpointName from the struct.

runtime/server_http_response.go
```diff
@@ -81,23 +75,8 @@ func (res *ServerHTTPResponse) finish() {
...
-	if counter == nil {
-		res.Request.Logger.Error(
-			"Could not emit statusCode metric",
-			zap.Int("UnexpectedStatusCode", res.StatusCode),
-		)
-	} else {
-		counter.Inc(1)
-	}
-
-	res.metrics.requestLatency.Record(
```
Raynos: Let's keep the metric emission here in the `finish()` method.

runtime/server_http_request.go
```diff
@@ -96,13 +83,8 @@ func (req *ServerHTTPRequest) start(endpoint string, handler string) {
 		/* coverage ignore next line */
 		return
 	}
-
-	req.EndpointName = endpoint
-	req.HandlerName = handler
 	req.started = true
 	req.startTime = time.Now()
-
-	req.metrics.requestRecvd.Inc(1)
```
Raynos: Let's keep the recvd metric here in the `start()` method.

runtime/router.go
```diff
@@ -250,30 +158,48 @@ func (router *HTTPRouter) Register(
...
 func (router *HTTPRouter) handleMethodNotAllowed(
-	w http.ResponseWriter, r *http.Request,
+	w http.ResponseWriter,
+	r *http.Request,
 ) {
-	resFields := []zapcore.Field{
-		zap.Int(statusCodeZapName, 405),
-	}
-	writeLogs(router.gateway.Logger, logRequestFields(r), resFields)
+	var startTime, finishTime time.Time
```
Raynos: would prefer if we just allocated an `NewServerHTTPRequest()` here and started it. Then we can finish the res associated with it.

runtime/router.go
```diff
@@ -250,30 +158,48 @@ func (router *HTTPRouter) Register(
...
-func (router *HTTPRouter) handleNotFound(w http.ResponseWriter, r *http.Request) {
-	resFields := []zapcore.Field{
-		zap.Int(statusCodeZapName, 404),
-	}
-	writeLogs(router.gateway.Logger, logRequestFields(r), resFields)
+func (router *HTTPRouter) handleNotFound(
+	w http.ResponseWriter,
+	r *http.Request,
+) {
+	var startTime, finishTime time.Time
```
Raynos: would prefer if we just allocated an NewServerHTTPRequest() here and started it. Then we can finish the res associated with it.

runtime/router.go
```diff
@@ -134,77 +58,60 @@ func NewRouterEndpoint(
...
 ) {
-	reqFields := logRequestFields(r)
-	var resFields []zapcore.Field
-
+	var req *ServerHTTPRequest
 	defer func() {
-		writeLogs(endpoint.gateway.Logger, reqFields, resFields)
+		// Post-process logs & metrics
+		writeLogs(endpoint.gateway.Logger, logRequestFields(r), logResponseFields(req.res))
+		endpoint.metrics.CollectMetrics(req.startTime, req.res.finishTime, req.res.StatusCode)
```
Raynos: let's not defer metrics collection and do it inline in `finish()` and `start()`

runtime/router.go
ravirajj: This particular EndpointName field was not used by the logger.

runtime/call_metrics.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	return &metrics
+}
+
+// CollectMetrics for inbound HTTP calls
+func (m *InboundHTTPMetrics) CollectMetrics(startTime, finishTime time.Time, statusCode int) {
+	m.Recvd.Inc(1)
+	_, known := knownStatusCodes[statusCode]
+	if !known {
+		m.logger.Error(unknownStatusCodeMsg, zap.Int(unknownStatusCode, statusCode))
```
Raynos: please put the error string inline here so it's easier to find.

Raynos: please put the `unknownStatusCode` inline here as well so it's easier to read.

runtime/router.go
srivastavankit: The question is specifically to ensure that we are able to tag the logger with endpoint context and if it would be used in future, so wondering why do we need to remove it now.

ravirajj: The plan is to pass in `map[string]string` to `NewRouterEndpoint()` and use that to create tagged scope and add fields to logger.

runtime/router.go
ravirajj: I don't think it's worth allocating the req/res objects here just to record the start and finish times.

runtime/router.go
ravirajj: We have a cleaner separation of concerns by not embedding the metrics collection in the res/req.  The req/res also doesn't need to spend time collecting metrics before it writes a response to the wire.  We write to the wire quickly, then record logs/metrics extra.  The defer delay to emit metrics or logs will be negligible.

runtime/server_http_response.go
ravirajj: Ok.  As we discussed offline.

runtime/server_http_request.go
ravirajj: Yes, as we discussed offline.

runtime/call_metrics.go
ravirajj: Done.

runtime/router.go
```diff
@@ -134,77 +55,55 @@ func NewRouterEndpoint(
 	handlerName string,
 	handler HandlerFn,
 ) *RouterEndpoint {
-	endpointTags := map[string]string{
+	logger := gateway.Logger.With(
+		zap.String("endpoint", endpointName),
```
Raynos: can we make this `endpointID` and `handlerID`

runtime/router.go
```diff
@@ -250,74 +150,28 @@ func (router *HTTPRouter) Register(
...
+func (router *HTTPRouter) handleNotFound(
+	w http.ResponseWriter,
+	r *http.Request,
+) {
+	req := NewServerHTTPRequest(w, r, nil, router.notFoundEndpoint)
 	// TODO custom NotFound
-	// A NotFound request is not started...
-	// TODO: inc.finish()
 	http.NotFound(w, r)
+	req.res.writeHeader(http.StatusNotFound)
```
Raynos: why do we have to call `writeHeader` ? and can we do so before `http.NotFound()` ?

runtime/router.go
```diff
@@ -250,74 +150,28 @@ func (router *HTTPRouter) Register(
...
-func writeLogs(l *zap.Logger, reqFlds []zapcore.Field, resFlds []zapcore.Field) {
-	fields := reqFlds
-	if resFlds != nil {
-		fields = append(reqFlds, resFlds...)
-	}
-	l.Info(
-		"Finished an incoming server HTTP request",
-		fields...,
-	)
+	req.res.writeHeader(http.StatusMethodNotAllowed)
```
Raynos: why do we have to call `writeHeader` ? and can we do so before `http.Error()` ?

runtime/server_http_request.go
```diff
@@ -30,31 +30,25 @@ import (
...
 	startTime   time.Time
-	metrics     *EndpointMetrics
+	metrics     *InboundHTTPMetrics
 	queryValues url.Values
 	parseFailed bool
-
-	Logger *zap.Logger
-	Scope  tally.Scope
-
-	EndpointName string
```
Raynos: it's useful to keep the `EndpointName` and `HandlerName` for future debugging.

runtime/server_http_request.go
```diff
@@ -65,28 +59,22 @@ func NewServerHTTPRequest(
...
-
-		Logger: endpoint.gateway.Logger,
-		Scope:  endpoint.gateway.AllHostScope,
-
-		URL:     r.URL,
-		Method:  r.Method,
-		Params:  params,
-		Header:  NewServerHTTPHeader(r.Header),
-		metrics: &endpoint.metrics,
+		Logger:      endpoint.logger.With(logRequestFields(r)...),
```
Raynos: we do not want to put all of these fields in the logger as it's just too much data and/or PII. These log fields are only used for the request finished log line.

runtime/router.go
ravirajj: Done.

runtime/router.go
ravirajj: Actually, I just need to set the `req.res.StatusCode`, so that it gets picked up by `finish()`.  Removing call to `writeHeader()`

runtime/router.go
ravirajj: Just need to set `req.res.StatusCode`, not call `writeHeader`

runtime/server_http_request.go
ravirajj: Done.

runtime/server_http_request.go
ravirajj: Well, we mainly call `logger.Error()`, so I thought it would be good to get as much context as possible.

PR #202, opened 2017-09-06
Upgrade thriftrw to v1.6.0

ChuntaoLu:
Thriftrw v1.6.0 supports getter helper generation for optional fields, feature parity with Apache Thrift.


glide.yaml
```diff
@@ -35,9 +35,7 @@ import:
 - package: github.com/kardianos/osext
   version: c2c54e542fb797ad986b31721e1baedf214ca413
 - package: go.uber.org/thriftrw
-  version: bbc73fe4e851cac67043d021242eb03920c2f1dd
-  repo:    https://github.com/raynos/thriftrw-go.git
-  vcs:     git
+  version: v1.6.0
```
ravirajj: https://github.com/thriftrw/thriftrw-go recommends setting the version to `^1`

glide.yaml
```diff
@@ -35,9 +35,7 @@ import:
 - package: github.com/kardianos/osext
   version: c2c54e542fb797ad986b31721e1baedf214ca413
 - package: go.uber.org/thriftrw
-  version: bbc73fe4e851cac67043d021242eb03920c2f1dd
-  repo:    https://github.com/raynos/thriftrw-go.git
```
ravirajj: We were using this thriftrw fork since it added `json:required` tags to struct fields, as they are used by easyjson for validation.

ChuntaoLu: v1.6.0 has that feature merged

glide.yaml
ChuntaoLu: that sets the minimum version requirement to be 1.0.0, which is not sufficient, we can do ^1.6.0, but I prefer to have it pinned to a specific version so that `make generate` is consistent and changes from generate code won't happen when I upgrade other packages.

ravirajj: Yes, let's do `^1.6.0`.

srivastavankit: Yes, Can we pin to a specific version? There are small nuances between versions and if developers have different versions on their machines, then it results in generated code to be written on each diff. (even unrelated changes)

ChuntaoLu: Yep, that's why I am pinning it so that `glide up && make generate` won't produce unexpected diff

Matt-Esch: I prefer v1.6.0. Consumers of the package should not be depending on thriftrw manaully, so all other references are nested libraries. Those libraries need to consider the version we've picked as compatible with them. If they do not, then that's an issue that needs to be resolved. We should avoid upgrading of modules that cause churn in project generated code. I'm not against using the latest minor version for modules that are runtime/test only and that we trust to be semantically versioned.

PR #204, opened 2017-09-07
runtime: ensure per-worker scope has prefix

Raynos:
This means that we can always query for per-worker
metrics without having the "all" tag in the host
dimension.

r: @uber/zanzibar-team


runtime/gateway.go
```diff
@@ -332,8 +332,12 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 	)
 	// As per M3 best practices, creating separate all-host and per-host metrics
 	// to reduce metric cardinality when querying metrics for all hosts.
-	gateway.AllHostScope = gateway.RootScope.Tagged(map[string]string{"host": "all"})
-	gateway.PerHostScope = gateway.RootScope.Tagged(map[string]string{"host": GetHostname()})
+	gateway.AllHostScope = gateway.RootScope.Tagged(
+		map[string]string{"host": "all"},
```
ravirajj: You can also remove the `host: all` tag altogether if you like.

PR #206, opened 2017-09-13
add remote config

cl1337:
implement remote config in addition to static config. What diverges remote config from static config is remote config is read only, remote config is been periodically reloaded from a config json file that could be updated by other services.

- periodically polling routine to refreshing config
- pub/sub with a callback function upon config change

### DIFF
--- a/runtime/remote_config.go
+++ b/runtime/remote_config.go
@@ -23,7 +23,6 @@ package zanzibar
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"sync"
@@ -43,7 +42,7 @@ type RemoteConfig interface {
 	GetInt(string, int64) int64
 	GetString(string, string) string
 	GetStruct(string, interface{}) bool
-	Subscribe(string, string, *func())
+	Subscribe(string, string, *Callback)
 	Unsubscribe(string)
 	Refresh() error
 	Close()
@@ -55,9 +54,12 @@ type RemoteConfigOptions struct {
 	PollingInterval time.Duration
 }
 
-// Subscriber is used to track subscribers' callbacks and keys they subscribe
+// Callback is registered to be triggered upon config change
+type Callback func(map[string]bool)
+
+// Subscriber is used to track subscribers' callbacks and keys subscribed
 type Subscriber struct {
-	Callback *func()
+	Callback *Callback
 	Key      string
 }
 
@@ -95,17 +97,16 @@ type remoteConfig struct {
 //		remoteConfig also provides pub/sub to allow caller act on specific key changes
 func NewRemoteConfig(cfg *RemoteConfigOptions, logger *zap.Logger, scope tally.Scope) (RemoteConfig, error) {
 	if logger == nil {
-		logger = zap.NewNop()
+		return nil, errors.Errorf("no logger is passed in")
+	}
+	if scope == nil {
+		return nil, errors.Errorf("no scope is passed in")
 	}
 	if len(cfg.FilePath) < 1 || cfg.PollingInterval == 0 {
 		return nil, errors.Errorf("invalid remote config options")
 	}
 	if _, err := os.Stat(cfg.FilePath); os.IsNotExist(err) {
 		return nil, errors.Errorf("invalid remote config file path")
-	}
-	if scope == nil {
-		logger.Warn("no valid metrics scope")
-		scope = tally.NoopScope
 	}
 	rc := &remoteConfig{
 		config:      cfg,
@@ -131,11 +132,15 @@ func (rc *remoteConfig) loadConfig() RemoteConfigMap {
 func (rc *remoteConfig) getValidatedValue(key string, vt jsonparser.ValueType) (*RemoteConfigValue, bool) {
 	ret, ok := rc.loadConfig()[key]
 	if !ok {
-		rc.logger.Warn(fmt.Sprintf("Key <%s> missing", key))
+		rc.logger.With(zap.String("key", key)).Warn("Key is missing")
 		return nil, false
 	}
 	if ret.dataType != vt {
-		rc.logger.Warn(fmt.Sprintf("Key <%s> type mismatch (expected %s actual %s)", key, vt, ret.dataType))
+		rc.logger.With(
+			zap.String("key", key),
+			zap.Any("expected_type", vt),
+			zap.Any("actual_type", ret.dataType),
+		).Warn("Key type mismatch")
 		return nil, false
 	}
 	return ret, true
@@ -168,7 +173,7 @@ func (rc *remoteConfig) GetInt(key string, fallback int64) int64 {
 	if v, err := jsonparser.ParseInt(ret.bytes); err == nil {
 		return v
 	}
-	rc.logger.Warn(fmt.Sprintf("Key <%s> is not int64", key))
+	rc.logger.With(zap.String("key", key)).Warn("key is not int64")
 	return fallback
 }
 
@@ -188,6 +193,10 @@ func (rc *remoteConfig) GetStruct(key string, ptr interface{}) bool {
 		return false
 	}
 	if err := json.Unmarshal(ret.bytes, ptr); err != nil {
+		rc.logger.With(
+			zap.ByteString("bytes", ret.bytes),
+			zap.String("error", err.Error()),
+		).Error("GetStruct unmarshal error")
 		return false
 	}
 	return true
@@ -237,7 +246,9 @@ func (rc *remoteConfig) Refresh() error {
 func (rc *remoteConfig) checkAndReturnStat() (os.FileInfo, bool) {
 	stat, err := os.Stat(rc.config.FilePath)
 	if err != nil {
-		rc.logger.Error("Error stat remote config file " + rc.config.FilePath)
+		rc.logger.With(
+			zap.String("configPath", rc.config.FilePath),
+		).Error("Error stat remote config file")
 		return nil, true
 	}
 	if rc.currStat != nil && rc.currStat.ModTime().Equal(stat.ModTime()) && rc.currStat.Size() == stat.Size() {
@@ -283,13 +294,18 @@ func (rc *remoteConfig) Close() {
 }
 
 // Subscribe adds a callback function to be executed when config refresh is complete
-// Subscribe takes a string identifier and a function pointer to be called after
-// a completed config refresh
-func (rc *remoteConfig) Subscribe(identifier, key string, fn *func()) {
+// Subscribe takes a string identifier, a key and a function pointer to be called after
+// a config change for a certain key
+// Params:
+// 		callback: callback function should be lightweight, ideally just reload the
+// 				  key subscribed.
+// 		key: if no specific key passed in (empty string), callback will be triggered on
+//			 *every* config update
+func (rc *remoteConfig) Subscribe(identifier, key string, callback *Callback) {
 	rc.mutex.Lock()
 	defer rc.mutex.Unlock()
 	rc.subscribers[identifier] = &Subscriber{
-		Callback: fn,
+		Callback: callback,
 		Key:      key,
 	}
 }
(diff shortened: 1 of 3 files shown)


runtime/remote_config.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	err = rc.reloadConfig()
+	if stat != nil && err == nil {
+		rc.currStat = stat
+	}
+	rc.execCallbacks()
+	return err
+}
+
+// _checkAndReturnStat returns file stat and boolean of `whether we should refresh from config file`
```
cl1337: note -> fix name in comments

runtime/remote_config_test.go
```diff
@@ -0,0 +1,344 @@
...
+	zanzibar "github.com/uber/zanzibar/runtime"
+	"io/ioutil"
+	"os"
+	"sync"
+	"time"
+)
+
+const tmpCfgPath = "/tmp/zanzibar_remote_cfg.json"
+const refreshInterval = time.Second * 10
+const defaultCfgJsonStr = `{
```
cl1337: what's the suggested way to manage fixtures in tests for this repo

runtime/remote_config.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+// RemoteConfig allow accessing values from a periodically refreshed json file
+type RemoteConfig interface {
+	GetBoolean(string) (bool, error)
+	GetFloat(string, float64) (float64, error)
+	GetInt(string, int64) (int64, error)
+	GetString(string, string) (string, error)
+	GetStruct(string, interface{}) error
+	Subscribe(string, *func())
```
Raynos: in `Subscribe()` we need to know which keys changed and which did not.

runtime/remote_config.go
```diff
@@ -0,0 +1,300 @@
...
+	"go.uber.org/zap"
+)
+
+// RemoteConfig allow accessing values from a periodically refreshed json file
+type RemoteConfig interface {
+	GetBoolean(string) (bool, error)
+	GetFloat(string, float64) (float64, error)
+	GetInt(string, int64) (int64, error)
+	GetString(string, string) (string, error)
+	GetStruct(string, interface{}) error
```
Raynos: `GetStruct()` will need a default value too `GetStruct(string, interface{}, interface{}) error`

runtime/remote_config.go
```diff
@@ -0,0 +1,300 @@
...
+
+	"github.com/buger/jsonparser"
+	"github.com/pkg/errors"
+	"github.com/uber-go/tally"
+	"go.uber.org/zap"
+)
+
+// RemoteConfig allow accessing values from a periodically refreshed json file
+type RemoteConfig interface {
+	GetBoolean(string) (bool, error)
```
Raynos: `GetBoolean(string, bool) (bool, error)`

runtime/remote_config.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if ret.dataType != vt {
+		return nil, errors.Errorf("Key <%s> type mismatch (expected %s actual %s)", key, vt, ret.dataType)
+	}
+	return ret, nil
+}
+
+// GetBoolean returns the value as a boolean
+func (rc *remoteConfig) GetBoolean(key string) (bool, error) {
+	var (
```
Raynos: no need for this `var` statement ? can just use 

```
ret, err := rc.getValidatedValue(key, jsonparser.Boolean)
if err != nil {
  return false, err
}
v, _ := jsonparser.ParseBoolean(ret.bytes)
return v, nil
```

runtime/remote_config.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func (rc *remoteConfig) loadConfig() RemoteConfigMap {
+	return rc.props.Load().(RemoteConfigMap)
+}
+
+func (rc *remoteConfig) getValidatedValue(key string, vt jsonparser.ValueType) (*RemoteConfigValue, error) {
+	ret, ok := rc.loadConfig()[key]
+	if !ok {
+		return nil, errors.Errorf("Key <%s> does not exist", key)
```
Raynos: A key not existing is not an error; we should return the default value instead.

Maybe this func needs to return `(value, ok, error)`

runtime/remote_config.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// GetFloat returns the value as a float64
+func (rc *remoteConfig) GetFloat(key string, fallback float64) (float64, error) {
+	var (
+		ret *RemoteConfigValue
+		err error
+	)
+	if ret, err = rc.getValidatedValue(key, jsonparser.Number); err != nil {
+		return fallback, err
```
Raynos: if there is a type mismatch returning the fallback is a good idea.

runtime/remote_config.go
```diff
@@ -0,0 +1,300 @@
...
+	)
+	if ret, err = rc.getValidatedValue(key, jsonparser.Boolean); err != nil {
+		return false, err
+	}
+	v, _ := jsonparser.ParseBoolean(ret.bytes)
+	return v, nil
+}
+
+// GetFloat returns the value as a float64
+func (rc *remoteConfig) GetFloat(key string, fallback float64) (float64, error) {
```
Raynos: Honestly having an interface `GetFloat(string, float64) float64` would be simpler.

We can do `rc.logger.Warn("something", error)` if there is a type mismatch.

runtime/remote_config.go
cl1337: mm ... I maybe misunderstood.

current implementation assumes users `know` exactly what key they're listening and they just `remoteConfig.get("that-key")` whenever config reloads, something like:

```
remoteConfig.Subscribe("chao", &func(){
       v, err := remoteConfig.get("some-key")
      // handle err
       reloadNewConfig(v)
})
```

am I getting you right that we actually want better granularity when key changes? like:

```
remoteConfig.Subscribe("chao", "some-key", &func(){
       v, err := remoteConfig.get("some-key")
      // handle err
       reloadNewConfig(v)
})
```

runtime/remote_config.go
cl1337: makes sense

runtime/remote_config.go
cl1337: makes sense

runtime/remote_config.go
cl1337: good catch

runtime/remote_config.go
cl1337: Above 2 comments my take away is we want silent swallow error (?) which I don't think it's a good approach, I get the point that if you pass a fallback value, when fallback happens it's a legit case,  still IMHO caller should decide not remote config

cl1337:
thanks for @Raynos for the review!

runtime/remote_config.go
Raynos: we need to know what has changed so that if our key has NOT changed we do not re-read it for no reason.

runtime/remote_config.go
Raynos: If there is a typo in the remote config UI we do not want to hit an err code branch.

Basically if a key does not exist or its configured poorly we should have the same semantics of using the fallback.

We can use metrics/logs/alerts to handle misconfigured keys but our runtime behaviour should just use the fallback.

We should not have code like

```
val, err := rc.GetBool('should.i.do.a', true)
if err != nil {
  return err
}
if val {
  doA();
}
```

In this example the rc key is about whether we should do a thing. If its misconfigured or non-existant we should do the thing. 

We definitely do not want to swallow the error. we can emit logs and metrics.

runtime/remote_config.go
Raynos: Actually that doesn't make any sense. Maybe `GetStruct(string, interface{}, interface{}) interface{}` but that's a bit annoying.

Probably `GetStruct(string, interface{}) bool` is best then it can be used as

```
var MyData data;
ok := rc.GetStruct('a-key', &data)
if !ok {
  data = defaultData
}
```

cl1337: lol

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+// and polling interval (decides how frequent we check ).
+// remote config file follows the same requirements with runtime/static_config/StaticConfig
+// remote config consider fallback legit case, fallback happens silent without throwing error
+// Why RemoteConfig:
+// 		What diverge staticConfig with remoteConfig is that staticConfig allows value setting
+//		while remoteConfig is read only.
+//		remoteConfig also provides pub/sub to allow caller act on specific key changes
+func NewRemoteConfig(cfg *RemoteConfigOptions, logger *zap.Logger, scope tally.Scope) (RemoteConfig, error) {
+	if logger == nil {
+		logger = zap.NewNop()
```
Raynos: let's make logger mandatory as we need to see `logger.error()`

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	if len(cfg.FilePath) < 1 || cfg.PollingInterval == 0 {
+		return nil, errors.Errorf("invalid remote config options")
+	}
+	if _, err := os.Stat(cfg.FilePath); os.IsNotExist(err) {
+		return nil, errors.Errorf("invalid remote config file path")
+	}
+	if scope == nil {
```
Raynos: let's also make this one mandatory.

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+func (rc *remoteConfig) loadConfig() RemoteConfigMap {
+	return rc.props.Load().(RemoteConfigMap)
+}
+
+func (rc *remoteConfig) getValidatedValue(key string, vt jsonparser.ValueType) (*RemoteConfigValue, bool) {
+	ret, ok := rc.loadConfig()[key]
+	if !ok {
+		rc.logger.Warn(fmt.Sprintf("Key <%s> missing", key))
```
Raynos: we dont use sprintf for logs

```
rc.logger.Warn("RemoteConfig could not find key",
    zap.String("key", key),
)
```

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+		return nil, false
+	}
+	if ret.dataType != vt {
+		rc.logger.Warn(fmt.Sprintf("Key <%s> type mismatch (expected %s actual %s)", key, vt, ret.dataType))
```
Raynos: We dont use sprintf for logs

```
rc.logger.Warn("RemoteConfig could not match type",
    zap.String("key", key),
    zap.String("expectedType", vt),
    zap.String("actualType", ret.dataType),
)
```

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+// GetInt returns the value as int64
+func (rc *remoteConfig) GetInt(key string, fallback int64) int64 {
+	ret, ok := rc.getValidatedValue(key, jsonparser.Number)
+	if !ok {
+		return fallback
+	}
+	if v, err := jsonparser.ParseInt(ret.bytes); err == nil {
+		return v
+	}
+	rc.logger.Warn(fmt.Sprintf("Key <%s> is not int64", key))
```
Raynos: no sprintf.

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+
+// GetStruct loads struct ptr interface{}
+func (rc *remoteConfig) GetStruct(key string, ptr interface{}) bool {
+	ret, ok := rc.getValidatedValue(key, jsonparser.Object)
+	if !ok {
+		return false
+	}
+	if err := json.Unmarshal(ret.bytes, ptr); err != nil {
+		return false
+	}
```
Raynos: we should log the json unmarshal err here too.

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+	rc.execCallbacks(changedMap)
+	return err
+}
+
+// checkAndReturnStat returns file stat and boolean of `whether we should refresh from config file`
+// don't refresh if config file hasn't change since last refresh
+func (rc *remoteConfig) checkAndReturnStat() (os.FileInfo, bool) {
+	stat, err := os.Stat(rc.config.FilePath)
+	if err != nil {
+		rc.logger.Error("Error stat remote config file " + rc.config.FilePath)
```
Raynos: no string concat either.

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+
+func (rc *remoteConfig) poll() {
+	for {
+		select {
+		case <-rc.poller.C:
+			result := "ok"
+			if err := rc.Refresh(); err != nil {
+				result = "err"
+			}
+			rc.scope.Tagged(map[string]string{"result": result}).Counter("remote_config.polling")
```
Raynos: Let's pre-allocate two tagged scopes on the `rc` instance; one for `ok` and one for `err`.

We can also pre-allocate the counter.

Also we need to call `Inc(1)` ;)

runtime/remote_config.go
```diff
@@ -0,0 +1,327 @@
...
+func (rc *remoteConfig) Unsubscribe(identifier string) {
+	rc.mutex.Lock()
+	defer rc.mutex.Unlock()
+	delete(rc.subscribers, identifier)
+}
+
+func (rc *remoteConfig) execCallbacks(changedMap map[string]bool) {
+	for _, s := range rc.subscribers {
+		if _, ok := changedMap[s.Key]; ok {
+			go (*s.Callback)()
```
Raynos: why do we execute each callback in it's own goroutine ? This can get quite large.

runtime/remote_config.go
```diff
@@ -131,11 +132,15 @@ func (rc *remoteConfig) loadConfig() RemoteConfigMap {
 func (rc *remoteConfig) getValidatedValue(key string, vt jsonparser.ValueType) (*RemoteConfigValue, bool) {
 	ret, ok := rc.loadConfig()[key]
 	if !ok {
-		rc.logger.Warn(fmt.Sprintf("Key <%s> missing", key))
+		rc.logger.With(zap.String("key", key)).Warn("Key is missing")
```
Raynos: no need to call `With()` i think that will allocate a child logger.

Can just do `rc.Logger.Warn("Key is missing", zap.String("key", key))`

cl1337: sounds good, does this apply to all logging cases?

cl1337:
ops need to rebase

PR #207, opened 2017-09-19
Add support for URL params to server endpoint.

Raynos:
This PR adds support URL params for the server.

There are a few subtle nuances. If there are required thrift fields
that come from URL params you need to mark them as "not in JSON"

Currently this only supports reading strings from the URL params
and not booleans/floats/ints.

r: @uber/zanzibar-team


codegen/test_data/clients/bar.gogen
```diff
@@ -532,7 +532,7 @@ func (c *barClient) ArgWithParams(
 	req := zanzibar.NewClientHTTPRequest(c.clientID, "ArgWithParams", c.httpClient)
 
 	// Generate full URL.
-	fullURL := c.httpClient.BaseURL + "/bar" + "/" + string(r.UUID) + "/segment" + "/" + string(r.Params.UserUUID)
+	fullURL := c.httpClient.BaseURL + "/bar" + "/argWithParams" + "/" + string(r.UUID) + "/segment" + "/" + string(r.Params.UserUUID)
```
ravirajj: Is `fmt.Sprintf()` faster than string concatenation?

examples/example-gateway/idl/endpoints/bar/bar.thrift
```diff
@@ -35,7 +35,10 @@ struct QueryParamsOptsStruct {
 }
 
 struct ParamsStruct {
-    1: required string userUUID (zanzibar.http.ref = "params.user-uuid")
+    1: required string userUUID (
+        zanzibar.http.ref = "params.user-uuid"
+        go.tag = "json:\"-\""
```
ravirajj: Does this remove the field from the json body?

Raynos: yes it does.

codegen/test_data/clients/bar.gogen
Raynos: not sure.

codegen/method.go
```diff
@@ -431,10 +443,89 @@ func findParamsAnnotation(
 	walkFieldGroups(fields, visitor)
 
 	if identifier == "" {
-		return "", false
+		return "", required, false
```
cl1337: nit: more than one return of the same type may want to label all of them, i.e.   (required, ok bool)

codegen/test_data/clients/bar.gogen
cl1337: fmt.Sprintf allocates a new string object and it's slow, use https://golang.org/pkg/bytes/#Buffer.WriteString if performance is a concern, which seems not here

examples/example-gateway/idl/endpoints/contacts/contacts.thrift
```diff
@@ -30,7 +30,7 @@ struct Contact {
 }
 
 struct SaveContactsRequest {
-    1: required UUID userUUID (zanzibar.http.ref = "params.userUUID");
+    1: required string userUUID (zanzibar.http.ref = "params.userUUID");
```
cl1337: curious why

Raynos: if it's uuid then we have to generate code like `args.Request.userUUID = UUID(req.Params.ByName('userUUID'))`

Generating that convert a string to `UUID` alias logic is non-trivial.

PR #208, opened 2017-09-19
Fix golint issue blocking generation.

Raynos:
r: @uber/zanzibar-team

### DIFF
--- a/codegen/templates/service.tmpl
+++ b/codegen/templates/service.tmpl
@@ -38,9 +38,11 @@ func CreateGateway(
 }
 
 func registerDeps(g *zanzibar.Gateway, deps *module.Dependencies) error {
+	var err error
 	{{- range $class, $instances := $instance.ResolvedDependencies }}
 	{{- range $idx, $instance := $instances }}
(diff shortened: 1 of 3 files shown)


codegen/templates/service.tmpl
```diff
@@ -38,9 +38,11 @@ func CreateGateway(
 }
 
 func registerDeps(g *zanzibar.Gateway, deps *module.Dependencies) error {
+	var err error
```
srivastavankit: instead can you use named response variable?

`func registerDeps(g *zanzibar.Gateway, deps *module.Dependencies) (err error)`

ChuntaoLu: Hmmm, I have always wondered why some prefer named return variables. To me, when I see a return with no value, I have to go back to the function signature to figure out what's returned, and then I need to find out where that named return variable is set, seems too much hassle.

srivastavankit: you dont need to use named variables with empty return. It avoids the variable declaration, thats all.

PR #210, opened 2017-09-20
Add option to use sidecar router for generated clients

ChuntaoLu:
This PR adds a `useSidecarRouter` boolean field in client-config.json to indicate the generated client should use `sidecarRouter.$type.ip` and `sidecarRouter.$type.port` config to communicate with its backend.

The tchannel client integration is complete in this PR, http client integration involves extra headers and will be done in a future PR.


codegen/templates/tchannel_client.tmpl
```diff
@@ -50,9 +51,13 @@ func {{$exportName}}(gateway *zanzibar.Gateway) Client {
 	serviceName := gateway.Config.MustGetString("clients.{{$clientID}}.serviceName")
 	sc := gateway.Channel.GetSubChannel(serviceName, tchannel.Isolated)
 
-	{{/* TODO: (lu) support multiple host port */ -}}
+	{{if $useSidecarRouter -}}
+	ip := gateway.Config.MustGetString("sidecarRouter.tchannel.ip")
```
Raynos: why can we not just configure the ip and port of the client to be "the sidecar router ip".

What if we use multiple different side car routers for different clients ?

srivastavankit: As @Raynos said, waterfall model makes sense --
- client specific sidecar ip
- default sidecar ip

ChuntaoLu: Here it happens to be just ip and port for tchannel client, there could be additional logic (e.g., headers) for http clients.  Besides, if many clients uses the sidecar, it saves from configuring the same ip/port per client, it's not a big deal, but nice to have.

> What if we use multiple different side car routers for different clients ?

Not sure how practical it is, but if it is, we can change the field value from boolean to `useSidecarRouter: $sidecarRouterName` and match the correct ip/port.

codegen/templates/tchannel_client.tmpl
```diff
@@ -50,9 +51,13 @@ func {{$exportName}}(gateway *zanzibar.Gateway) Client {
...
+	port := gateway.Config.MustGetInt("sidecarRouter.tchannel.port")
```
srivastavankit: is there a reason to not merge ip/port into one field?

ChuntaoLu: No particular reason, just being consistent with the existing config format.

codegen/templates/tchannel_client.tmpl
ChuntaoLu: Done.

ChuntaoLu:
@Raynos Updated and added multi sidecar support.

PR #211, opened 2017-09-20
http_client: only emit sent after sending the request

Raynos:
This will not falsey emit a sent metric if we cannot json
encode the body or something else fails like not establishing
an outbound socket.

r: @ravirraj @uber/zanzibar-team

### DIFF
--- a/runtime/client_http_request.go
+++ b/runtime/client_http_request.go
@@ -73,9 +73,6 @@ func (req *ClientHTTPRequest) start() {
 	}
 	req.started = true
 	req.startTime = time.Now()
-
-	// emit metrics
-	req.metrics.Sent.Inc(1)
 }
 
 // WriteJSON will send a json http request out.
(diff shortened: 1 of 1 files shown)


runtime/client_http_request.go
```diff
@@ -128,6 +125,9 @@ func (req *ClientHTTPRequest) Do(
 		return nil, err
 	}
 
+	// emit metrics
+	req.metrics.Sent.Inc(1)
```
ravirajj: Did you want to move this to `finish()` instead?  Just keep it with all the other emitted metrics as I had it earlier.

Raynos: this should be right after we write arg3 to the request.

Raynos: Or for this HTTP case; it should be right here at 128 after we send an outbound http request.

PR #213, opened 2017-09-21
Add rk support

oliviazhang0809:
support adding routing key for catalyst services

### DIFF
--- a/runtime/static_config.go
+++ b/runtime/static_config.go
@@ -206,6 +206,21 @@ func (conf *StaticConfig) MustGetInt(key string) int64 {
 	panic(errors.Errorf("Key (%s) not available", key))
 }
 
+// ContainsKey returns the value as a string or panics.
+func (conf *StaticConfig) ContainsKey(key string) bool {
+	if conf.destroyed {
+		panic(errors.Errorf("Cannot get(%s) because destroyed", key))
+	}
+	if _, contains := conf.seedConfig[key]; contains {
+		return true
+	}
+
+	if _, contains := conf.configValues[key]; contains {
+		return true
+	}
+	return false
+}
+
 // MustGetString returns the value as a string or panics.
 func (conf *StaticConfig) MustGetString(key string) string {
 	if conf.destroyed {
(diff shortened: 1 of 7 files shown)


runtime/tchannel_client.go
```diff
@@ -154,10 +157,13 @@ func (c *tchannelClient) Call(ctx context.Context, thriftService, methodName str
 	retryOpts := &tchannel.RetryOptions{
 		TimeoutPerAttempt: c.timeoutPerAttempt,
 	}
-	ctx, cancel := tchannel.NewContextBuilder(c.timeout).
+	ctxBuilder := tchannel.NewContextBuilder(c.timeout).
 		SetParentContext(ctx).
-		SetRetryOptions(retryOpts).
-		Build()
+		SetRetryOptions(retryOpts)
+	if c.routingKey != nil {
```
zhenghuiwang: Nit: It seems to be simpler to use `string` as type for `routingKey` and check if it is not empty string.

Raynos:
Will also need code changes in template that calls `NewTChannelClient()` ?

runtime/static_config.go
```diff
@@ -206,6 +206,21 @@ func (conf *StaticConfig) MustGetInt(key string) int64 {
 	panic(errors.Errorf("Key (%s) not available", key))
 }
 
+// ContainsKey returns the value as a string or panics.
+func (conf *StaticConfig) ContainsKey(key string) bool {
+	if conf.destroyed {
+		panic(errors.Errorf("Cannot get(%s) because destroyed", key))
```
Raynos: Cannot containsKey(%s)

runtime/tchannel_client.go
ChuntaoLu: +1, seems like `string` type is sufficient

Raynos:
@oliviazhang0809 

Travis says make cover failed ( not enough coverage ).

You can run `make cover ; make view-istanbul` to run the tests with coverage and view lack of coverage.

PR #214, opened 2017-09-21
Add http header validation to HTTP client.

Raynos:
This changes adds mandatory header checking to the http client
just like we have mandatory header checking for the http server.

r: @uber/zanzibar-team

### DIFF
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -94,17 +94,6 @@ func (c *{{$clientName}}) {{$methodName}}(
 	{{end -}}
 	req := zanzibar.NewClientHTTPRequest(c.clientID, "{{$methodName}}", c.httpClient)
 
-	{{if .ReqHeaders }}
-	headerErr := req.CheckHeaders({{.ReqHeaders | printf "%#v"}}, headers)
-	if headerErr != nil {
-		return {{ if eq .ResponseType "" -}}
-			nil, headerErr
-			{{- else -}}
-			defaultRes, nil, headerErr
-			{{- end}}
-	}
-	{{- end}}
-
 	{{if .ReqHeaderGoStatements }}
 	// TODO(jakev): populate request headers from thrift body
 	{{- end}}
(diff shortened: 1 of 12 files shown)


codegen/templates/http_client.tmpl
```diff
@@ -123,6 +119,18 @@ func (c *{{$clientName}}) {{$methodName}}(
 	if err != nil {
 		return {{if eq .ResponseType ""}}nil, err{{else}}defaultRes, nil, err{{end}}
 	}
+
+	{{if .ReqHeaders }}
+	headerErr := req.CheckHeaders({{.ReqHeaders | printf "%#v"}})
+	if headerErr != nil {
+		return {{ if eq .ResponseType "" -}}
```
ChuntaoLu: nit: shorter alternative `{{- if .ResponseType -}}defaultRes, {{- end -}}nil, headerErr`

runtime/client_http_request.go
```diff
@@ -78,6 +78,32 @@ func (req *ClientHTTPRequest) start() {
 	req.metrics.Sent.Inc(1)
 }
 
+// CheckHeaders verifies that the outbound request contaisn required headers
```
ChuntaoLu: typo: contaisn

runtime/client_http_request.go
```diff
@@ -78,6 +78,32 @@ func (req *ClientHTTPRequest) start() {
...
+		panic("must call `req.WriteJSON()` before `req.CheckHeaders()`")
+	}
+
+	actualHeaders := req.httpReq.Header
+
+	for _, headerName := range expected {
+		// TODO: case sensitivity ?
+		headerValue := actualHeaders.Get(headerName)
+		if headerValue == "" {
+			req.Logger.Warn("Got outbound request without mandatory header",
```
ravirajj: Should this be logged as an Error, since we failed the required headers check?  Or are you logging such client errors as Warnings?

runtime/client_http_request_test.go
```diff
@@ -157,6 +157,30 @@ func TestMakingClientCalLWithHeaders(t *testing.T) {
...
+	deps := bgateway.Dependencies.(*exampleGateway.DependenciesTree)
+	bar := deps.Client.Bar
+
+	_, _, err = bar.EchoI8(
+		context.Background(), nil, &clientsBarBar.Echo_EchoI8_Args{Arg: 42},
+	)
+
+	assert.NotNil(t, err)
+	assert.Equal(t, "Missing mandatory header: x-uuid", err.Error())
+}
```
ravirajj: Also test the log output.

runtime/client_http_request.go
Raynos: an `Error` is a pager duty offence. Do you want to get woken up for it ?

Raynos: An error is a bug; aka broken code. In theory this can be an `Error()` but it's possible for a user of edge-gateway to configure "mandatory" headers on the client but no mandatory headers on the endpoint and no header proxying.

PR #217, opened 2017-09-25
Generated tests import relative main/config.

Raynos:
This changes make the generated tests not use the
util package but instead import relative to their location
inside the generated code.

This unblocks edge-gateway

r: @uber/zanzibar-team

### DIFF
--- a/config/production.json
+++ b/config/production.json
@@ -12,6 +12,7 @@
 	"jaeger.sampler.param": 0.001,
 
 	"logger.output": "stdout",
+	"logger.fileName": "",
 
 	"metrics.type": "m3",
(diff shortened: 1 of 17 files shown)


config/production.json
```diff
@@ -12,6 +12,7 @@
 	"jaeger.sampler.param": 0.001,
 
 	"logger.output": "stdout",
+	"logger.fileName": "",
```
ChuntaoLu: hmm, the value being "" seems a bit confusing

Raynos: the default fileName is "" aka no file.

The config library expects keys to exist.

PR #220, opened 2017-09-27
Support url query renaming

ChuntaoLu:


codegen/method.go
```diff
@@ -1013,11 +1008,6 @@ func (ms *MethodSpec) setParseQueryParamStatements(
 
 		identifierName := camelCase(longQueryName) + "Query"
 
-		httpRefAnnotation := field.Annotations[antHTTPRef]
```
Raynos: this is not correct if the field.Annotation is for header then we cannot read from query params and must read from the header.

Same if the antHTTPRef annotation is body

codegen/test_data/clients/bar.gogen
```diff
@@ -542,8 +542,14 @@ func (c *barClient) ArgWithParams(
 	// Generate full URL.
 	fullURL := c.httpClient.BaseURL + "/bar" + "/argWithParams" + "/" + string(r.UUID) + "/segment" + "/" + string(r.Params.UserUUID)
 
+	queryValues := &url.Values{}
```
Raynos: these values are only written into URL params and should not be written into query params.

If the http.ref annotation is params or headers then it must NOT be written into the query parameters.

codegen/method.go
ChuntaoLu: Done

codegen/test_data/clients/bar.gogen
ChuntaoLu: Done

PR #221, opened 2017-09-28
generation for client header

oliviazhang0809:


Makefile
```diff
@@ -121,7 +121,7 @@ install-wrk:
 	cd ./vendor/wrk ; (ls ./wrk 2>/dev/null || make >install_wrk.log)
 
 .PHONY: test
-test: generate lint
+test: check-generate lint
```
Raynos: we want to run `generate` here. `check-generate` takes far too long

codegen/method.go
```diff
@@ -679,6 +686,72 @@ func (ms *MethodSpec) setResponseHeaderFields(
...
+				}
+				if len(seenOptStructs) == 0 {
+					statements.appendf(headerNameValuePair,
+						headerName, bodyIdentifier,
+					)
+				} else {
+					for seenStruct := range seenOptStructs {
+						if strings.HasPrefix(longFieldName, seenStruct) {
+							statements.appendf("if r%s != nil {", seenStruct)
+							statements.appendf(headerNameValuePair,
```
Raynos: I dont think this will work if `len(seenOptStruct) >= 2`

That would generate two if statements and two assignment expressions.

We want to have two loops I guess; one to open the if statements and one to close them.

codegen/templates/http_client.tmpl
```diff
@@ -96,6 +96,9 @@ func (c *{{$clientName}}) {{$methodName}}(
 
 	{{if .ReqHeaderGoStatements }}
 	// TODO(jakev): populate request headers from thrift body
```
Raynos: We can remove this TODO

Makefile
oliviazhang0809: @zhenghuiwang suggested since i had issue with easy json while tests passed locally but failed on travis

Raynos: yeah you sometimes have to run `check-generate` locally but it can take upto 2 minutes instead of 10 seconds. 

This is just something we have to do sometimes to unbreak the code :( 

Something about easyjson is order dependent and you have to re-generate it sometimes.

codegen/method.go
```diff
@@ -679,6 +686,74 @@ func (ms *MethodSpec) setResponseHeaderFields(
...
+				if len(seenOptStructs) == 0 {
+					statements.appendf(headerNameValuePair,
+						headerName, bodyIdentifier,
+					)
+				} else {
+					closeFunction := ""
+					for seenStruct := range seenOptStructs {
+						if strings.HasPrefix(longFieldName, seenStruct) {
+							statements.appendf("if r%s != nil {", seenStruct)
+							closeFunction = closeFunction + "}"
```
Raynos: `go fmt` might fix this but ideally we should do `closeFunction = closeFunction + "}\n"`

oliviazhang0809: `go fmt`ed - didnt work. land the diff for now.

PR #225, opened 2017-09-29
Convert thrift `module` into a .thrift file.

zhenghuiwang:
`Module` is a parsed data structure from a thrift file. This PR implements `Module.Code()`, which converts a `Module` to its thrift file format.

Unit Test:
We parse `expected_baz.thrift` into a `Module`. `Module.Code` should return the exact same file content as `expected_baz.thrift`.
 
r: @uber/zanzibar-team


backend/repository/thrift_compile.go
```diff
@@ -407,3 +434,278 @@ func newEnumSpec(enum *compile.EnumSpec) *EnumSpec {
...
+	panic(fmt.Sprintf(
+		"No typedef, enum, nor struct definition found: typespec %+v", ts))
+}
+
+func enumTypeCodeBlock(e *EnumSpec, line int) *CodeBlock {
+	result := bytes.NewBuffer(nil)
+	result.WriteString(fmt.Sprintf("enum %s {\n", e.Name))
+	items := e.Items
+	for i := 0; i < len(items)-1; i++ {
+		result.WriteString(fmt.Sprintf("\t%s,\n", items[i].Name))
```
fymemon: It looks like the optional value part of an enum item isn't being serialized.

backend/repository/thrift_compile.go
```diff
@@ -407,3 +434,278 @@ func newEnumSpec(enum *compile.EnumSpec) *EnumSpec {
...
+	relPath = filepath.Clean(strings.TrimPrefix(relPath, "../"))
+	return &CodeBlock{
+		Code:  fmt.Sprintf("include \"%s\"", relPath),
+		Order: im.Line,
+	}
+}
+
+// CodeBlock converts a constant to a CodeBlock.
+func (c *Constant) CodeBlock(curFilePath string) *CodeBlock {
+	return &CodeBlock{
```
fymemon: Constants, typedefs, and enum items can have annotations, but doesn't look like they're being handled here.

zhenghuiwang: As i tested out, constants can not have annotations. Added annotations to typedefs and enum items.

backend/repository/thrift_compile.go
zhenghuiwang: The enum definition doesn't have `optional` part.https://git1-us-west.apache.org/repos/asf?p=thrift.git;a=blob;f=tutorial/tutorial.thrift;hb=HEAD#l91

PR #227, opened 2017-10-02
Update to thriftrw v1.8.0

ravirajj:
- Optional fields of generated structs now enable the `omitempty` JSON option
if the field holds a list, set, or map.


codegen/package_test.go
```diff
@@ -131,5 +131,5 @@ func TestGoCustomTypeError(t *testing.T) {
 	spec := &compile.StructSpec{}
 	_, err := h.TypeFullName(spec)
 	assert.Error(t, err)
-	assert.Equal(t, "goCustomType called with native type (*compile.StructSpec) &{false   0 [] map[]}", err.Error())
+	assert.Equal(t, "goCustomType called with native type (*compile.StructSpec) &{false   0 []  map[]}", err.Error())
```
ChuntaoLu: why the extra space?

glide.lock
```diff
@@ -1,6 +1,15 @@
-hash: d6512ee48265f77a1c34516c5ffb73828ccb937533a56454568e6ed2211ba607
-updated: 2017-09-20T11:55:50.9196227-07:00
+hash: 86524339d17f3edec1d59ef1349c00a011ffbd51245f8c3b10b6967c76c513a9
+updated: 2017-10-01T18:35:39.603933045-07:00
 imports:
+- name: cloud.google.com/go
```
Raynos: There is so much random stuff in the glide.lock ; no idea why. 

Can you just delete everything except the thriftrw version bump ?

codegen/package_test.go
ravirajj: I don't know actually!  Are the recent commits causing this?

glide.lock
ravirajj: AFAIK, `glide update` always updates the entire glide.lock file.

Raynos: yeah glide is dumb so we should hand edit the glide.lock to undo the bugs in glide :(

ravirajj: Well glide update also updates the hash at the top of the glide.lock file.  If we hand edit this file, we might break that hash!

Raynos: i dont think the hash matters XD

codegen/package_test.go
ravirajj: Formatting the error with `%+v` revealed that the 2 extra spaces between `false` and `0` are for `Name` and `File` empty strings, the 1 extra space between `[]` and `map[]` is for the `Doc` empty string:

```
expected: "goCustomType called with native type (*compile.StructSpec) &{false   0 []  map[]}"
actual: "goCustomType called with native type (*compile.StructSpec) &{linkOnce:false Name: File: Type:0 Fields:[] Doc: Annotations:map[]}"
```

PR #228, opened 2017-10-02
better tests & fixtures

cl1337:
This PR improves tests

* use consistent TestFixtures across codegen and UI backend, now TestFixtures map allows user to update one test fixture a time, and leave identifiers to later error surfacing
* consolidate testFixtures into endpointSpec, get rid of separate `_test.json` to avoid extra file read at code gen time and extra file write in UI backend 
* get rid of temp fields such as `ResponseString` and `ResponseHeaderKeys`, also the logic to generate those fields. spec json considered source of truth should not store tmp result
* According template change above
* naming: handleID / handleId / handlerId .... and such


codegen/gateway.go
```diff
@@ -564,12 +563,33 @@ func NewEndpointSpec(
...
+
 func augmentHTTPEndpointSpec(
 	espec *EndpointSpec,
 	endpointConfigObj map[string]interface{},
 	midSpecs map[string]*MiddlewareSpec,
 ) (*EndpointSpec, error) {
-	espec.TestFixtures = endpointConfigObj["testFixtures"].([]interface{})
+	espec, err := reloadTestFixtures(espec, endpointConfigObj)
+	if err != nil {
+		return nil, errors.Errorf("Unable to parse test cases")
```
zhenghuiwang: Don't drop the original error here. Use `errors.Wrap()`

codegen/gateway.go
```diff
@@ -564,12 +563,33 @@ func NewEndpointSpec(
 	return augmentHTTPEndpointSpec(espec, endpointConfigObj, midSpecs)
 }
 
+func reloadTestFixtures(espec *EndpointSpec, endpointConfigObj map[string]interface{}) (*EndpointSpec, error) {
```
zhenghuiwang: This function seems to have a fat interface - It is not clear what this function does by just looking at the input/output of this func.

A better way is to define func `testFixtures(endpointConfigObj map[string]interface{}) (map[string]*EndpointTestFixture , error)`

codegen/module_system.go
```diff
@@ -65,44 +63,46 @@ type StructMeta struct {
...
+// FixtureBlob is map[string]interface{} that implements default string
+// used for headers and (http | tchannel) request/response
+type FixtureBlob map[string]interface{}
+
+// String convert map[string]interface{} to string
+func (fb *FixtureBlob) String() string {
+	if str, err := json.Marshal(fb); err == nil {
+		return string(str)
+	}
+	return ""
```
zhenghuiwang: the error for marshaling is dropped here silently. let's at least return `err.Error()`

codegen/module_system.go
```diff
@@ -65,44 +63,46 @@ type StructMeta struct {
...
+}
+
+// EndpointTestFixture saves mocked requests/responses for an endpoint test.
+type EndpointTestFixture struct {
+	TestName           string                        `json:"testName"`
+	EndpointID         string                        `json:"endpointID"`
+	HandleID           string                        `json:"handleID"`
+	EndpointRequest    FixtureBlob                   `json:"endpointRequest"`
```
zhenghuiwang: It seems only JSON request is allowed. How is string request body handled?

codegen/test_data/endpoint_tests/bar_bar_method_argwithheaders_test.gogen
```diff
@@ -0,0 +1,24 @@
+// Code generated by zanzibar
```
zhenghuiwang: Seems the code generation has issues. These are empty files.

codegen/gateway.go
cl1337: this follows the similar IO of `augmentHTTPEndpointSpec` below (and the same idea): augment http endpoint with generated tests.

EndpointSpec and endpointConfigObj seems to be the in-mem model (go struct) and its json representation at some extend. Test fixtures for example, it can be easily merged with json tag annotation, prefer to have them together and consistent with `augmentHTTPEndpointSpec`, what do you think?

codegen/gateway.go
cl1337: good point

backend/repository/config.go
```diff
@@ -56,50 +56,19 @@ type ThriftServiceMap map[string]map[string]*ThriftService
...
-	ClientMethod         string                 `json:"clientMethod"`
-	ClientRequest        map[string]interface{} `json:"clientRequest"`
-	ClientRequestString  string                 `json:"clientRequestString"`
-	ClientReqHeaders     map[string]string      `json:"clientReqHeaders"`
-	ClientReqHeaderKeys  []string               `json:"clientReqHeaderKeys"`
-	ClientResponse       map[string]interface{} `json:"clientResponse"`
-	ClientResponseString string                 `json:"clientResponseString"`
-	ClientResHeaders     map[string]string      `json:"clientResHeaders"`
-	ClientResHeaderKeys  []string               `json:"clientResHeaderKeys"`
+	ID               string                       `json:"endpointID"`
```
ChuntaoLu: This is quite a breaking change that we should not make, it requires zanzibar users to change their JSON configs for no good reason. `id` is conventional in JSON, so does the camelcase of `endpointId`.  And `ID` is mandated by golint.

codegen/module_system.go
cl1337: nope this is used in template, template is dumb don't know how to handle it, actually this should be `{}`

codegen/test_data/endpoint_tests/bar_bar_method_argwithheaders_test.gogen
cl1337: these are files without tests added, we should 1: skip if no tests fixtures 2: enforce every endpoint should have test fixtures, I vote for 2 :)

codegen/module_system.go
cl1337: this pending on discussion

cl1337:
thx @zhenghuiwang for the review

backend/repository/config.go
cl1337: good reason is `ID` and `Id` , `handler` and `handle` are inconsistent across repo, as long as we have consistent naming in the repo I don't mind `Id` or `ID`

ChuntaoLu: `endpointId` is in JSON config while `ID` is in Go code. As long as it's all `somethingId` in JSON and `somethingID` in Go, it's consistent.

cl1337:
per discussion offline with @zhenghuiwang  and @ChuntaoLu  (a lot thanks!)

http/tchannel response/request for clients/endpoint:
```lang=go
type TchannelResponse TestFixtureBody

// for http
type TestFixtureHTTPResponse struct {
     StatusCode int
     Body TestFixtureBody
}

type TestFixtureBody struct {
     BodyString string
     BodyJson map[string]interface{}
}

func (tf *TestFixtureBody) Marshal() (String, error)
```

current `EndpointResponse` is used as response body (should be fixed in template & codegen)

naming inconsistency -> handler/handle is the only thing needs to be fixed
for `endpointId` and `handleId`, keep json key name, and go naming convention

cl1337:
```lang=go
// FixtureBlob is map[string]interface{} that implements default string
// used for headers and (http | tchannel) request/response
type FixtureBlob map[string]interface{}

// String convert FixtureBlob to string
func (fb *FixtureBlob) String() string {
	if str, err := json.Marshal(fb); err == nil {
		return string(str)
	}
	return "{}"
}

// FixtureHTTPBody is used to create http body in test fixtures
type FixtureHTTPBody struct {
	BodyType   string      `json:"bodyType"`
	BodyString string      `json:"bodyString,omitempty"`
	BodyJson   FixtureBlob `json:"bodyJson,omitempty"`
}

// String convert FixtureHTTPBody to string
func (fhb *FixtureHTTPBody) String() string {
	if fhb.BodyType == "string" {
		return fhb.BodyString
	}
	return fhb.BodyJson.String()
}

// TestFixtureHTTPResponse is test fixture for http response
type TestFixtureHTTPResponse struct {
	StatusCode int         `json:"statusCode"`
	Body       FixtureBlob `json:"body,omitempty"`
}

// TestFixtureResponse is test fixture for client/endpoint response
type TestFixtureResponse struct {
	ResponseType     string                  `json:"responseType"`
	HTTPResponse     TestFixtureHTTPResponse `json:"httpResponse,omitempty"`
	TChannelResponse FixtureBlob             `json:"tchannelResponse,omitempty"`
}

// EndpointTestFixture saves mocked requests/responses for an endpoint test.
type EndpointTestFixture struct {
	TestName           string                        `json:"testName"`
	EndpointID         string                        `json:"endpointId"`
	HandleID           string                        `json:"handleId"`
	EndpointRequest    FixtureBlob                   `json:"endpointRequest"` // there's no difference between http or tchannel request
	EndpointReqHeaders FixtureBlob                   `json:"endpointReqHeaders"`
	EndpointResponse   TestFixtureResponse           `json:"endpointResponse"`
	EndpointResHeaders FixtureBlob                   `json:"endpointResHeaders"`
	ClientTestFixtures map[string]*ClientTestFixture `json:"clientTestFixtures"`
	TestServiceName    string                        `json:"testServiceName"` // The service module that mounts the endpoint
}

// ClientTestFixture saves mocked client request/response for an endpoint test.
type ClientTestFixture struct {
	ClientID         string              `json:"clientId"`
	ClientMethod     string              `json:"clientMethod"`
	ClientRequest    FixtureBlob         `json:"clientRequest"` // there's no difference between http or tchannel request
	ClientReqHeaders FixtureBlob         `json:"clientReqHeaders"`
	ClientResponse   TestFixtureResponse `json:"clientResponse"`
	ClientResHeaders FixtureBlob         `json:"clientResHeaders"`
}
```

will fix test/gen/json if no objection on above test fixtures

zhenghuiwang:
Thanks for sending this out. Highlight some changes that I would make.

// FixtureJSON is map[string]interface{} 
type **FixtureJSON** map[string]interface{}

**// Deleted String() for FixtureJSON, which is the same as a marshal func but drops the error**

// **FixtureBody** is used to create http body **and tchannel body** in test fixtures
type **FixtureBody** struct {
	BodyType   string      `json:"bodyType"`
    // BodyString is set if the actual body type is string
	BodyString string      `json:"bodyString,omitempty"`
    // BodyJSON is set if the actual body is a JSON object.
	**BodyJSON**   **FixtureJSON** `json:"bodyJson,omitempty"`
}

// String convert FixtureHTTPBody to string
func (fhb *FixtureBody) String() **(string, error)** {
	if fhb.BodyType == "string" {
		return fhb.BodyString **,nil**
	}
	**b, err :=** json.Mashal(fhb.BodyJSON)
    **reutrn string(b), err**
}

// **FixtureHTTPResponse** is test fixture for http response
type **FixtureHTTPResponse** struct {
	StatusCode int         `json:"statusCode"`
	Body       **FixtureBody** `json:"body,omitempty"`
}

// **FixtureResponse** is test fixture for client/endpoint response
// one of the fields will not be nil
type **FixtureResponse** struct {
	HTTPResponse     *FixtureHTTPResponse `json:"httpResponse,omitempty"`
	TChannelResponse **FixtureBody**             `json:"tchannelResponse,omitempty"`
}

cl1337:
thanks @zhenghuiwang , my 2 cents

1: on
```
// TestFixtureHTTPResponse is test fixture for http response
type TestFixtureHTTPResponse struct {
	StatusCode int         `json:"statusCode"`
	Body       FixtureBlob `json:"body,omitempty"`
}
```
this should be `FixtureHTTPBody ` instead of `FixtureBlob`, `FixtureBlob` is always map[string]interface{}, which is not true if http response is string

2: on
`FixtureHTTPBody`  what I'm saying is:
at codegen in template, there should be no errors / surprises: https://github.com/uber/zanzibar/pull/228/files#diff-a6a4e0f02c678d147f4e698031c8fcc2R86
we should be able to comfortably print the string representation

marshal with error should happen during repository/backend, and surfaced to user asap.

if we see error later on in codegen regarding user input, we're doing it wrong 


other renaming looks fine :)

zhenghuiwang:
As discussed offline with @cl1337 

1. The following struct should be used as the body for http and tchannel response
type FixtureBody struct {
    BodyType string json:"bodyType"
    // BodyString is set if the actual body type is string
    BodyString string json:"bodyString,omitempty"
    // BodyJSON is set if the actual body is a JSON object.
    BodyJSON FixtureJSON json:"bodyJson,omitempty"
}

2. A `String()` is needed for FixtureBody to render the template. Because we can't handle errors during executing the template, we will panic inside `String()`. We will recover the panics from executing the template and return corresponding errors gracefully:
1. Besides `String()`, executing template can potentially throw panic. Adding recover can handle these gracefully too.
2. The original approach is to marshal and pass the serialized strings to the template. There are too many boilerplate code to write.

cl1337:
add more reviewers, this change is expensive let's make sure we do this right (to fix previous issue) and never have to do it again

codegen/module_system.go
```diff
@@ -65,44 +63,104 @@ type StructMeta struct {
...
+		panic(errors.New("invalid response type"))
+	}
+}
+
+// EndpointTestFixture saves mocked requests/responses for an endpoint test.
+type EndpointTestFixture struct {
+	TestName           string                        `json:"testName"`
+	EndpointID         string                        `json:"endpointId"`
+	HandleID           string                        `json:"handleId"`
+	EndpointRequest    FixtureBlob                   `json:"endpointRequest"` // there's no difference between http or tchannel request
```
zhenghuiwang: This is different from your proposal.

codegen/gateway.go
zhenghuiwang: augmentHTTPEndpointSpec does lots of stuff for augmentations. So you can't know what is done by its signature but read the code. (It may not be a good thing but makes sense)

getting testFixture form config is a very specific task, and should have thin readable thin interface.

backend/repository/config.go
```diff
@@ -56,50 +56,19 @@ type ThriftServiceMap map[string]map[string]*ThriftService
...
-	ClientResHeaders     map[string]string      `json:"clientResHeaders"`
-	ClientResHeaderKeys  []string               `json:"clientResHeaderKeys"`
+	ID               string                       `json:"endpointId"`
+	Type             ProtocolType                 `json:"endpointType"`
+	HandleID         string                       `json:"handleId"`
+	ThriftFile       string                       `json:"thriftFile"`
+	ThriftFileSha    string                       `json:"thriftFileSha,omitempty"`
+	ThriftMethodName string                       `json:"thriftMethodName"`
+	WorkflowType     string                       `json:"workflowType"`
+	ClientID         string                       `json:"clientId"`
```
zhenghuiwang: Renaming the json tag(e.g. ClientID) is not essential to this PR and is a breaking change. Please make a separate PR to rename the tags.

cl1337: I thought we agreed to introduce breaking change for the consistency of `go field ID and json tag Id`?
This PR already contains breaking changes would it make sense to reduce the # of breaking changes?
I feel strong about consistent naming

codegen/gateway.go
cl1337: thanks for the explanation, it makes sense, let's use `testFixtures(endpointConfigObj map[string]interface{}) (map[string]*EndpointTestFixture , error)`

codegen/module_system.go
```diff
@@ -65,44 +63,104 @@ type StructMeta struct {
...
+type FixtureHTTPResponse struct {
+	StatusCode int          `json:"statusCode"`
+	Body       *FixtureBody `json:"body,omitempty"`
+}
+
+// FixtureResponse is test fixture for client/endpoint response
+type FixtureResponse struct {
+	ResponseType     ProtocalType         `json:"responseType"`
+	HTTPResponse     *FixtureHTTPResponse `json:"httpResponse,omitempty"`
+	TChannelResponse *FixtureBlob         `json:"tchannelResponse,omitempty"`
```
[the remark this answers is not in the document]
cl1337: just to confirm preference of `FixtureBlob` over `*FixtureBlob`? Why?

codegen/module_system.go
cl1337: per our discussion previously: http request and tchannel request are the same and all can be represented with map[string]interface{}, so we reuse fixtureBlob
any suggestion here?

backend/repository/config.go
zhenghuiwang: sounds good.

codegen/module_system.go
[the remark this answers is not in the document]
zhenghuiwang: FixtureBlob is a map[string]interface. `map` types are reference types, whose value is a memory address for the underlying struct.

`map` type is essentially a pointer type (it is cheap to copy and can be tested again nil).

`*FixtureBlob` is a pointer to a pointer, which is not necessary here. (and not necessary in most cases, unless you want to change the reference to point to another underlying map)

codegen/module_system.go
zhenghuiwang: I thought we agree to have a `request` type for both endpoint and client request, sth like

```
type Request struct{
     Method string
     Headers map[string]interface{}
     Body FixtureBody
}
```

The current definitions have two problems: 1. they duplicate the client request types and endpoint request types and the fields are not bundled together. 2. The request body can be of string type, same as the response body type

cl1337: headers should be aside there's already a field for it
good point on body, http request body being string I didn't see any but seems string is legit https://en.wikipedia.org/wiki/HTTP_message_body

so let's have the same idea with responseBody
```
type FixtureRequest struct {
    RequestType string
    HTTPRequest HTTPRequest
    TchannelRequest JSONBlob
}

type HTTPRequest struct {
    Method string
    Body *FixtureBody
}
```

codegen/module_system.go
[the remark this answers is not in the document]
cl1337: gotcha, * as necessary for explicit underlying null struct to show in json even omitempty, https://goplay.space/#ydrL7OyZc1 here's an example

I guess for map[string]interface{} this is optional, I don't have strong opinion there, we'll move to pointer moving forward if we need underlying struct

codegen/gateway.go
```diff
@@ -564,12 +563,31 @@ func NewEndpointSpec(
 	return augmentHTTPEndpointSpec(espec, endpointConfigObj, midSpecs)
 }
 
+func testFixtures(endpointConfigObj map[string]interface{}) (ret map[string]*EndpointTestFixture, rErr error) {
```
zhenghuiwang: Don't use named returned variables for the sake of avoiding some declarations inside the func. 

The main problem of named return is that it is not clear about the return value at the line of return. Readers have to go back to the previous code to find out the value of each variable.

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

codegen/module_system.go
```diff
@@ -65,44 +63,136 @@ type StructMeta struct {
...
+}
+
+// FixtureResponse is test fixture for client/endpoint response
+type FixtureResponse struct {
+	ResponseType     ProtocalType         `json:"responseType"`
+	HTTPResponse     *FixtureHTTPResponse `json:"httpResponse,omitempty"`
+	TChannelResponse FixtureBlob          `json:"tchannelResponse,omitempty"`
+}
+
+// Body returns the string representation of FixtureResponse
```
zhenghuiwang: Let' leave a comment on why this func is panic instead of returning error.

PR #229, opened 2017-10-02
make json format

cl1337:
we seems to be using eclint for lint only, what's the rationale of not `eclint fix` ??

formatting of json actually matters in tests when compare to golden file, can we have json formatted by linter instead of depend on `--update`? I saw manual crafted json everywhere ...

ideally `make format-json` should not break anything, otherwise the code somehow depend on json format, which is wrong

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ COVER_PKGS = $(shell glide novendor | grep -v "test/..." | \
 GO_FILES := $(shell \
 	find . '(' -path '*/.*' -o -path './vendor' -o -path './workspace' ')' -prune -o -name '*.go' -print | cut -b3-)
 
+JSON_FILES := $(shell \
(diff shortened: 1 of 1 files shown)


ChuntaoLu:
> what's the rationale of not eclint fix ??

We do have [eclint-fix](https://github.com/uber/zanzibar/blob/master/Makefile#L47).

cl1337:
https://github.com/uber/zanzibar/blob/master/config/production.json#L7

Raynos:
is whitespace bad ? I thought it was nice :(

cl1337:
lol white space is awesome as long as it's intended 🕺 

will add more dir to eclint fix & force it

cl1337:
put this change in https://github.com/uber/zanzibar/pull/228 closing this one

PR #230, opened 2017-10-02
add schema validation for middleware configs

jakelarkn:
@uber/zanzibar-team

### DIFF
--- a/backend/repository/endpoint_update.go
+++ b/backend/repository/endpoint_update.go
@@ -25,7 +25,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"encoding/json"
 	"github.com/pkg/errors"
 	"github.com/uber/zanzibar/codegen"
 )
@@ -87,17 +86,9 @@ func (r *Repository) validateEndpointCfg(req *EndpointConfig) error {
 	for _, mid := range req.Middlewares {
 		for _, midCfg := range r.gatewayConfig.Middlewares {
 			if mid.Name == midCfg.Name {
-				result, err := r.validator.ValidateGo(midCfg.SchemaFile, mid.Options)
-				if err != nil && result != nil {
-					resultErrors, _ := json.Marshal(result.Errors())
-					return errors.Wrap(err, "schema validation error\nErrors:\n"+string(resultErrors))
-				}
-				if err != nil {
-					return errors.Wrap(err, "schema validation error unknown")
-				}
-				if !result.Valid() {
-					resultErrors, _ := json.Marshal(result.Errors())
-					return errors.New("schema validation error :\n" + string(resultErrors))
+				err := codegen.SchemaValidateGo(midCfg.SchemaFile, mid.Options)
+				if err != nil  {
+					return err
 				}
 			}
 		}
--- a/backend/repository/endpoint_update.go
+++ b/backend/repository/endpoint_update.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"encoding/json"
 	"github.com/pkg/errors"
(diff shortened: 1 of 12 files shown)


backend/repository/endpoint_update.go
```diff
@@ -82,6 +83,25 @@ func (r *Repository) validateEndpointCfg(req *EndpointConfig) error {
 	} else {
 		return errors.Errorf("client type %q is not supported", clientCfg.Type)
 	}
+
+	for _, mid := range req.Middlewares {
+		for _, midCfg := range r.gatewayConfig.Middlewares {
+			if mid.Name == midCfg.Name {
+				result, err := r.validator.ValidateGo(midCfg.SchemaFile, mid.Options)
```
zhenghuiwang: The error handling logic here is complex. It is better to hide these `if` inside `ValidteGo` and define it as

`func ValidateSchema(schemaFile string, arbitraryObj map[string]interface{}) error`

codegen/schema_validator.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (r *resultStruct) Errors() []ResultError {
+	return r.errors
+}
+
+// SchemaValidator abstracts undlerlying schema validation library
+type SchemaValidator struct{}
+
+// ValidateGo performs JSON schema validation from schema file on to arbitrary Go object
+func (s *SchemaValidator) ValidateGo(schemaFile string, arbitraryObj map[string]interface{}) (Result, error) {
```
zhenghuiwang: `ValidateGo` should be defined as a function instead of a method of an empty struct.

backend/repository/repository.go
```diff
@@ -44,6 +45,7 @@ type Repository struct {
 	// Cached error when getting gateway config.
 	gatewayConfigError error
 	sync.RWMutex
+	validator *codegen.SchemaValidator
```
zhenghuiwang: `validator` is not a state for `repository` and it is cheap to create. So it should be introduced as a local variable right before schema validation.

backend/repository/endpoint_update.go
jakelarkn: ok.  i was using the gojsonschema API as a guideline.  It seems they chose to distinguish general errors (file read, JSON parsing  etc)  from validation errors which are properties on the Result object.  But I am ambivalent and your suggestion would keep this code simpler.

codegen/schema_validator.go
jakelarkn: a Validator is inherently stateful as it caches schema (gojsonschema does this internally) , its arguable better to make this somewhat implied with an instance method  instead of static function.

backend/repository/endpoint_update.go
jakelarkn: ok. i was using the gojsonschema API as a guideline. It seems they chose to distinguish general errors (file read, JSON parsing etc) from validation errors which are properties on the Result object. But I am ambivalent and your suggestion would keep this code simpler.

glide.yaml
```diff
@@ -42,8 +42,8 @@ import:
   version: 6edb48674bd9467b8e91fda004f2bd7202d60ce4
 - package: github.com/uber/jaeger-client-go
   version: 2.8.0
-# tchannel needs master for golang.org/x/net
 - package: github.com/buger/jsonparser
   version: 5b691c8ebc4af5191baa426561d62f1b5115d6e5
 - package: github.com/stretchr/testify
   version: f6abca593680b2315d2075e0f5e2a9751e3f431a
+- package: github.com/xeipuuv/gojsonschema
```
Raynos: we want an explicit version here. not "master"

jakelarkn:
updated w/ comment changes

PR #231, opened 2017-10-02
Configure zap to emit metrics for info/warn/error

Raynos:
This change adds new metrics to zanzibar.Gateway where
we will emit a Counter for info/warn/error showing the 
total log throughput of the gateway.

r: @uber/zanzibar-team

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -111,11 +111,11 @@ func CreateGateway(
 	config.Freeze()
 
 	// order matters for following setup method calls
-	if err := gateway.setupLogger(config); err != nil {
+	if err := gateway.setupMetrics(config); err != nil {
 		return nil, err
 	}
 
-	if err := gateway.setupMetrics(config); err != nil {
+	if err := gateway.setupLogger(config); err != nil {
 		return nil, err
 	}
 
@@ -389,10 +389,13 @@ func (gateway *Gateway) setupLogger(config *StaticConfig) error {
 	}
 
 	zapLogger := zap.New(
-		zapcore.NewCore(
-			zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()),
-			output,
-			zap.InfoLevel,
+		NewLoggingZapCore(
+			zapcore.NewCore(
+				zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()),
+				output,
+				zap.InfoLevel,
+			),
+			gateway.AllHostScope,
 		),
 	)
 
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -388,14 +388,14 @@ func (gateway *Gateway) setupLogger(config *StaticConfig) error {
 		}
 	}
 
+	prodCore := zapcore.NewCore(
+		zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()),
+		output,
+		zap.InfoLevel,
+	)
 	zapLogger := zap.New(
-		NewLoggingZapCore(
-			zapcore.NewCore(
-				zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()),
-				output,
-				zap.InfoLevel,
-			),
-			gateway.AllHostScope,
+		NewInstrumentedZapCore(
+			prodCore, gateway.AllHostScope,
 		),
 	)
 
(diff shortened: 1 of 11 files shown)


runtime/gateway.go
```diff
@@ -389,10 +389,13 @@ func (gateway *Gateway) setupLogger(config *StaticConfig) error {
 	}
 
 	zapLogger := zap.New(
-		zapcore.NewCore(
-			zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()),
-			output,
-			zap.InfoLevel,
+		NewLoggingZapCore(
+			zapcore.NewCore(
```
ChuntaoLu: nit: many levels of indentation, local var for better readability

runtime/logger_metrics.go
```diff
@@ -0,0 +1,116 @@
...
+// THE SOFTWARE.
+
+package zanzibar
+
+import (
+	"github.com/uber-go/tally"
+	"go.uber.org/zap/zapcore"
+)
+
+type loggingZapCore struct {
```
ChuntaoLu: maybe `instrumentedZapCore` or `zapCoreWithMetrics` more close to what the wrapper is?

test/health_test.go
```diff
@@ -123,8 +124,10 @@ func TestHealthMetrics(t *testing.T) {
 
 	cgateway.MetricsWaitGroup.Wait()
 
+	// sleep to avoid race conditions with m3.Client
```
ChuntaoLu: Could we comment a bit more on why the race condition?

test/health_test.go
```diff
@@ -123,8 +124,10 @@ func TestHealthMetrics(t *testing.T) {
...
+	time.Sleep(100 * time.Millisecond)
```
ChuntaoLu: Don't feel good about using timer to get around race conditions, is there no other solution?

test/lib/test_gateway/test_gateway.go
```diff
@@ -107,6 +107,7 @@ type Options struct {
 	KnownHTTPBackends     []string
 	KnownTChannelBackends []string
 	CountMetrics          bool
+	MaxMetrics            int
```
ChuntaoLu: we should drop a comment on when and how it should be used.

test/health_test.go
```diff
@@ -158,12 +169,17 @@ func TestHealthMetrics(t *testing.T) {
...
+	loggedMetrics := metrics[tally.KeyForPrefixedStringMap("zap.logged.info", defaultTags)]
+	value = *loggedMetrics.MetricValue.Count.I64Value
+	assert.Equal(t, int64(2), value, "expected counter to be 2")
 }
 
 func TestRuntimeMetrics(t *testing.T) {
 	gateway, err := testGateway.CreateGateway(t, nil, &testGateway.Options{
 		CountMetrics:         true,
 		EnableRuntimeMetrics: true,
+		MaxMetrics:           31,
```
ChuntaoLu: maybe hoist https://github.com/uber/zanzibar/pull/231/files#diff-5a9b0fba1adefb12f372c4444b3ef8e6R194, and set the value to `numMetrics`?

oliviazhang0809:
looks good to me besides using timer.sleep to escape race condition. Do we have better way?

Raynos:
The reason for the sleep is non-determinism.

Conceptually we wait for e.g. "9 metric names" however these may be emitted as 

```
A
B
C
D = 2
E
F
G
H ( Wait() finishes )
```

Or

```
A
B
C
D = 1
E
F
G
H ( Wait() finishes )
D = 1
```

The metric server has no way of understanding that you are waiting for `D = 2` ; your just waiting for 9 metrics and non determinism means that its either 9 or 10 discrete events.

I would have to add new logic into the M3Service itself basically says `M3Service.Expect(name, value, timeout=100)` 

So instead of waiting for all metrics we just ask M3Service do we have a metric at this value and timeout if it takes too long. This would put the sleep and/or timeout in the `test/lib/m3_server` instead.

runtime/logger_metrics.go
Raynos: good idea.

ChuntaoLu:
lgtm, currently https://github.com/dominikh/go-tools/issues/197 is causing build failures

Raynos:
Yeah tests pass except for staticcheck because honnef.co.fail. 

Why do we install source code from an arbitrary domain on the internet again instead of a central repository :(

Even npm is better then golang package management.

PR #232, opened 2017-10-03
Registering optional fields and tags with tchannel server endpoint

ravirajj:


runtime/tchannel_server.go
```diff
@@ -98,34 +98,42 @@ func NewTChannelRouter(registrar tchannel.Registrar, g *Gateway) *TChannelRouter
 func (s *TChannelRouter) Register(
 	endpointID, handlerID, method string,
 	h TChannelHandler,
+	fields []zapcore.Field,
```
ChuntaoLu: This method is getting completely out of hand, the `register` semantic is simple, and yet now it is bloated with 6 arguments and returns an endpoint? It does way more than what its name means.

Raynos: Taking an options struct is fine. 

You could also pass in the `Endpoint` similar to the `Router` API for http.

ravirajj: Passing a `TChannelEndpoint` to `Register` like `Router` for http.

ravirajj:
Abandoning in favor of #237.

PR #233, opened 2017-10-03
Generate JSON (un)marshallers only for modules that need them

ChuntaoLu:
Currently we generate JSON (un)marshallers for everything under /gen-code, while only the modules that are of http/json need them.
This PR generates the JSON (un)marshallers only for the modules that actually need them.

r: @uber/zanzibar-team


codegen/method.go
```diff
@@ -131,15 +131,19 @@ type StatusCode struct {
...
+	antHTTPReqHeaders = "zanzibar.http.reqHeaders"
+	antHTTPResHeaders = "zanzibar.http.resHeaders"
+	antHTTPRef        = "zanzibar.http.ref"
+	antMeta           = "zanzibar.meta"
+	antHandler        = "zanzibar.handler"
+
+	// AntHTTPReqDefBoxed annotates a method so that the genereted method takes
+	// generated argument directly instead of a struct that warps the argument.
+	// The annotated method should have one and only one argument.
+	AntHTTPReqDefBoxed = "zanzibar.http.req.def"
```
Raynos: we are supposed to deprecate this :(

codegen/runner/pre-steps.sh
```diff
@@ -58,9 +59,37 @@ end=`date +%s`
 runtime=$((end-start))
 echo "Compiled easyjson : +$runtime"
 
+go build -o $RESOLVE_THRIFT_BINARY $RESOLVE_THRIFT_FILE
+
+# find the modules that actually need JSON (un)marshallers
+target_dirs=""
+found_thrifts=""
+config_files=$(find ${CONFIG_DIR} -name "*-config.json" | sort)
```
Raynos: Did you resolve imports ?

If we import X in a http+json thrift file and embed it we need a json marshaller generated for x

ChuntaoLu: There is no need to resolve imports if the `"zanzibar.http.req.def"` annotation is not used, i.e., the struct that wraps all the method arguments is (un)marshalled. E.g., [foo.FooStruct](https://github.com/uber/zanzibar/blob/lu.ej/examples/example-gateway/idl/clients/bar/bar.thrift#L104) is imported from `examples/example-gateway/clients/foo/foo.thrift`, here [Bar_TooManyArgs_Args](https://github.com/uber/zanzibar/blob/lu.ej/examples/example-gateway/build/gen-code/clients/bar/bar/bar_toomanyargs.go#L14) is used and there is unwrapping of it, we don't need a standalone json marshaller for `foo.FooStruct`, because `foo.FooStruct` marshallers are already defined in [examples/example-gateway/build/gen-code/clients/bar/bar/bar_toomanyargs_easyjson.go](https://github.com/uber/zanzibar/blob/lu.ej/examples/example-gateway/build/gen-code/clients/bar/bar/bar_toomanyargs_easyjson.go#L448-L624).

codegen/method.go
ChuntaoLu: If deprecating the annotation means the (un)marshalling are always on the wrapper struct, then it's easier, we don't even need to look for the imported thrifts.

ChuntaoLu:
> i dont think this works with nested imports.
> 
> You have to walk the entire thrift AST object model for all arguments and return types and if there any foreign imported structs in there then those files also need easyjson.

The nesting is taken care of by easyjson itself, as long as there is no `zanzibar.http.req.def` annotation, we don't need to manually generate standalone marshallers.

Raynos:
The nesting is not taking care of by easyjson itself.

That's not how easyjson works.

Easyjson will take a nested struct and call `Unmarshal` which uses stdlib reflection approach UNLESS you run `easyjson` on the other file manually.

ChuntaoLu:
https://github.com/uber/zanzibar/pull/233/files#diff-1212669c4a6d3884c14dd55c2b929d9d shows easyjson does take care of the nested thrifts and generates proper JSON (un)marshallers.

@Raynos Benchmarks show no regression:
```
#
#   benchmark output lu.ej
#

    Latency    27.96ms   16.12ms 239.85ms   84.29%
    Req/Sec     1.23k   248.76     1.86k    70.33%
  439907 requests in 30.09s, 48.25MB read
  Socket errors: connect 0, read 253, write 0, timeout 0
Requests/sec:  14619.01
Transfer/sec:      1.60MB

#
#   benchmark output master
#

    Latency    30.35ms   18.24ms 277.83ms   83.89%
    Req/Sec     1.12k   222.39     1.73k    68.72%
  401955 requests in 30.07s, 44.08MB read
  Socket errors: connect 0, read 254, write 0, timeout 1
  Non-2xx or 3xx responses: 1
Requests/sec:  13369.42
Transfer/sec:      1.47MB
```
```
#
#   benchmark output lu.ej
#

Running 30s test @ http://localhost:8093/googlenow/add-credentials
    Latency    23.03ms   17.30ms   1.35s    93.48%
    Req/Sec     1.46k   223.29     2.23k    72.33%
  525411 requests in 30.07s, 61.63MB read
  Socket errors: connect 0, read 242, write 0, timeout 0
  Non-2xx or 3xx responses: 3
Requests/sec:  17470.76
Transfer/sec:      2.05MB

#
#   benchmark output master
#

Running 30s test @ http://localhost:8093/googlenow/add-credentials
    Latency    23.46ms   14.46ms   1.36s    88.76%
    Req/Sec     1.44k   232.46     2.14k    69.42%
  517872 requests in 30.07s, 60.75MB read
  Socket errors: connect 0, read 246, write 0, timeout 0
  Non-2xx or 3xx responses: 1
Requests/sec:  17221.77
Transfer/sec:      2.02MB
```

PR #234, opened 2017-10-03
Add request error type

zhenghuiwang:
This PR adds `RequestError` type to capture invalid request fields so that the HTTP handlers can return accurate and structured errors.

This PR uses it at two places. There are more places to be changed to using `RequestError`.

r: @uber/zanzibar-team

### DIFF
--- a/backend/repository/handler.go
+++ b/backend/repository/handler.go
@@ -424,22 +424,32 @@ func (h *Handler) WriteErrorResponse(w http.ResponseWriter, code int, err error)
 	type stackTracer interface {
 		StackTrace() errors.StackTrace
 	}
+	errJSON := fmt.Sprintf("{\"error\": %q}", err.Error())
+	causeErr := errors.Cause(err)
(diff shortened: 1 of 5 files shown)


backend/repository/handler.go
```diff
@@ -424,22 +424,32 @@ func (h *Handler) WriteErrorResponse(w http.ResponseWriter, code int, err error)
...
+	if e, ok := causeErr.(*RequestError); ok {
+		code = http.StatusBadRequest
+		errJSON = e.JSON()
+		causeErr = e.Cause
+	}
+
 	var errStack string
-	if causeErr, ok := errors.Cause(err).(stackTracer); ok {
-		errStack = fmt.Sprintf("%+v", causeErr.StackTrace())
+	if e, ok := causeErr.(stackTracer); ok {
```
cl1337: will this tack trace surface to user?

zhenghuiwang: No, they are only for service logs.

PR #235, opened 2017-10-04
fix string return type

oliviazhang0809:


codegen/method.go
```diff
@@ -862,19 +865,36 @@ func (ms *MethodSpec) setTypeConverters(
...
-
 	respConverter := &TypeConverter{
 		LineBuilder: LineBuilder{},
 		Helper:      h,
 	}
 
-	err = respConverter.GenStructConverter(downstreamRespFields, respFields, respTransforms)
-	if err != nil {
-		return err
+	respConverter.append("func convert", pascalCase(ms.Name), "ClientResponse(in ", downstreamMethod.ResponseType, ") ", ms.ResponseType, "{")
```
Raynos: we dont have a lint rule yet for 80 characters but we try to stick to it, can add some linebreaks here for readability.

codegen/templates/endpoint.tmpl
```diff
@@ -171,6 +171,8 @@ func (handler *{{$handlerName}}) HandleRequest(
 
 	{{if eq .ResponseType "" -}}
 	res.WriteJSONBytes({{.OKStatusCode.Code}}, cliRespHeaders, nil)
+	{{- else if eq .ResponseType "string" -}}
+	res.WriteJSONBytes({{.OKStatusCode.Code}}, cliRespHeaders, []byte(response))
```
Raynos: I don't think this is valid json ? `"\"" + response + "\""` ?

oliviazhang0809: the response is a string itself. 
check out `build/endpoints/bar/bar_bar_method_helloworld.go`

test/endpoints/bar/bar_hello_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	assert.Equal(t, "200 OK", res.Status)
+	assert.Equal(t, 1, counter)
+
+	respBytes, err := ioutil.ReadAll(res.Body)
+	if !assert.NoError(t, err, "got http resp error") {
+		return
+	}
+
+	assert.Equal(t, string(respBytes), "hello")
```
Raynos: this is still not valid JSON. We should expect `"hello"` here. 

Also need to test what happens if send `"\"hello\""` and if strings are properly encoded and stuff.

test/endpoints/bar/bar_hello_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if !assert.NoError(t, err, "got bootstrap err") {
+		return
+	}
+	defer gateway.Close()
+
+	gateway.HTTPBackends()["bar"].HandleFunc(
+		"GET", "/bar/hello",
+		func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(200)
+			if _, err := w.Write([]byte(`"hello"`)); err != nil {
```
Raynos: Try doing `json.Marshal("I am\"string\"with\"nestedquotes")` or something and double check that the `json.Unmarshal()` from `res.Body` contains the same string.

PR #239, opened 2017-10-06
Adding default dependencies

ravirajj:
Created `DefaultDependencies` to pass the logger and scope to the modules, instead of passing `gateway` which may expose more than necessary.  `DefaultDependencies` should  eventually replace `gateway`, as we add other dependencies, including `Config`, etc., to the `DefaultDependencies`.


codegen/templates/module_initializer.tmpl
```diff
@@ -42,14 +42,18 @@ func InitializeDependencies(
...
 	tree := &DependenciesTree{}
 
-	{{- range $idx, $className := $instance.DependencyOrder}}
+	{{range $idx, $className := $instance.DependencyOrder}}
 	{{- $moduleInstances := (index $instance.RecursiveDependencies $className)}}
 	initialized{{$className | pascal}}Dependencies := &{{$className | title}}DependenciesNodes{}
 	tree.{{$className | title}} = initialized{{$className | pascal}}Dependencies
 
 	{{- range $idx, $dependency := $moduleInstances}}
 	{{- if $dependency.HasDependencies}}
```
Raynos: this `HasDependencies` check nolonger works because "everyone" has "default dependendices"

I think you want to get rid of this if statement all together.

codegen/templates/endpoint.tmpl
```diff
@@ -50,51 +50,50 @@ import (
...
 		Clients: deps.Client,
 	}
+	handler.endpoint = zanzibar.NewRouterEndpoint(
+		deps.Default.Logger, deps.Default.Scope,
+		"{{$endpointId}}", "{{$handleId}}",
+		{{ if len $middlewares | ne 0 -}}
+		zanzibar.NewStack([]zanzibar.MiddlewareHandle{
+			{{range $idx, $middleware := $middlewares -}}
+			{{$middleware.Name}}.NewMiddleWare(
+				deps.Default.Logger, deps.Default.Scope,
```
Raynos: This breaking change to the middleware api is going to be really annoying.

Let's not do it and continue to pass in `*zanzibar.Gateway` instead.

codegen/templates/module_initializer.tmpl
```diff
@@ -35,21 +33,25 @@ type {{$className | title}}DependenciesNodes struct {
...
-	return nil, {{$instance.PackageInfo.ExportName}}(gateway)
-	{{- else}}
 	tree := &DependenciesTree{}
 
-	{{- range $idx, $className := $instance.DependencyOrder}}
+	initializedDefaultDependencies := &zanzibar.DefaultDependencies{
+		Logger:  g.Logger,
+		Scope:   g.AllHostScope,
+		Config:  g.Config,
+		Channel: g.Channel,
```
Raynos: What's the use case for the `Channel` here ?

Raynos: Ok I understand now. Allocating sub channels and creating custom tchannel clients.

examples/example-gateway/middlewares/example/example.go
```diff
@@ -46,9 +50,13 @@ type MiddlewareState struct {
 // NewMiddleWare creates a new middleware that executes the next middleware
 // after performing it's operations.
 func NewMiddleWare(
```
Raynos: let's not change this API and keep it as `*zanzibar.Gateway`

codegen/templates/endpoint.tmpl
ravirajj: Sure. I can change it back to `*zanzibar.Gateway`.

codegen/templates/module_initializer.tmpl
ravirajj: Yes, it's needed for tchannel clients. I was trying to see if we could customize the default dependencies, so only `Channel` is added for tchannel clients, but it became complicated.

examples/example-gateway/middlewares/example/example.go
ravirajj: Sure.

codegen/templates/endpoint.tmpl
ravirajj: Or I can pass `gateway` and `deps` to all module initializers.

examples/example-gateway/middlewares/example/example.go
ravirajj: Sure, but the middleware logger and scope won't have the endpoint fields and tags.

PR #241, opened 2017-10-07
Removing tchannel stats reporter

ravirajj:
Removing tchannel stats reporter since custom metrics have been added to the tchannel server and client.


ravirajj:
@Raynos Good point, there are other metrics emitted by tchannel we'd want to listen too, apart from the blacklisted ones.

runtime/gateway.go
```diff
@@ -486,14 +488,16 @@ func (gateway *Gateway) setupTChannel(config *StaticConfig) error {
 	serviceName := config.MustGetString("tchannel.serviceName")
 	processName := config.MustGetString("tchannel.processName")
 
-	subScope := gateway.AllHostScope.SubScope("tchannel")
```
Raynos: I liked the subscope idea.

Raynos: Let's keep the tchannel prefix to avoid collisons.

ravirajj: Done.

test/endpoints/tchannel/baz/baz_metrics_test.go
```diff
@@ -166,30 +125,28 @@ func TestCallMetrics(t *testing.T) {
...
+	inboundSuccess := metrics[tally.KeyForPrefixedStringMap(
 		"test-gateway.test.all-workers.inbound.calls.success", endpointTags,
 	)]
 	value = *inboundSuccess.MetricValue.Count.I64Value
 	assert.Equal(t, int64(1), value, "expected counter to be 1")
 
 	tchannelOutboundNames := []string{
-		"test-gateway.test.all-workers.tchannel.outbound.calls.latency",
 		"test-gateway.test.all-workers.tchannel.outbound.calls.per-attempt.latency",
 		"test-gateway.test.all-workers.tchannel.outbound.calls.send",
```
Raynos: we do not want `send` either ; out metrics are called `sent` instead of `send` which is why its not in the blacklist yet.

ravirajj: This is most likely a typo bug in tchannel: https://github.com/uber/tchannel-go/blob/3abb4c025c1663b383452339a22d918cf9d0be0b/outbound.go#L196

Even though the docs say `outbound.calls.sent`: http://tchannel.readthedocs.io/en/latest/metrics/#call-metrics

PR #242, opened 2017-10-09
Add option to provide thriftrw source files

ChuntaoLu:
This PR adds an option to provide thriftrw sources files for Go code generation, the rationale is, depending on how idl files are managed, the idl dir might have thrift files that do not need to generate Go code.

It also includes changes to make the bash script more consistent.

### DIFF
--- a/codegen/runner/pre-steps.sh
+++ b/codegen/runner/pre-steps.sh
@@ -1,21 +1,29 @@
 #!/usr/bin/env bash
 set -e
 
-if [ -z $1 ]; then
+if [ -z "$1" ]; then
 	echo "build dir argument (\$1) is missing"
 	exit 1
 fi
 
-if [ -z $2 ]; then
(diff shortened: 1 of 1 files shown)


Raynos:
I think this has an issue on jenkins

```
goroutine 1 [running]:
main.main()
    /mnt/jenkins/workspace/test-edge-gateway/workspace/src/code.uber.internal/rt/edge-gateway/vendor/github.com/uber/zanzibar/scripts/resolve_thrift/main.go:41 +0xa22
panic: Failed to parse thrift file: ./idl/[]
```

Also the `resolve_thrift` binary doesnt exit 1 or something when it panics like `make generate` needs to exit non-zero ? 

Jenkins got a panic but it continue on anyway and completely ignored the panic :S

ChuntaoLu:
@Raynos issue should be fixed with https://github.com/uber/zanzibar/pull/243

PR #245, opened 2017-10-11
Cleanup dead middleware and bad defaults in config

Raynos:
We've had a logger middleware thats effectively dead sitting
around that's not being used.

We've also run into bugs with edge-gateway where "default"
values from the config were used that did not make any sense
at all like "my-gateway" in metrics.

r: @uber/zanzibar-team


config/production.json
```diff
@@ -25,10 +20,5 @@
 	"metrics.runtime.enableCPUMetrics": true,
 	"metrics.runtime.enableMemMetrics": true,
 	"metrics.runtime.enableGCMetrics": true,
-	"metrics.runtime.collectInterval": 1000,
-
-	"http.port": 4000,
-
-	"tchannel.serviceName": "my-gateway",
```
ChuntaoLu: Do we no longer need these?

Raynos: these are mandatory but the default values are "never useful". we will need to properly configure edge-gateway and now it will just panic instead of using a "useless default value".

PR #246, opened 2017-10-11
Add metric scope prefixes

Raynos:
We lost the metric scope prefixes in a recent refactor.

Adding them back.

r: @uber/zanzibar-team

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -333,10 +333,14 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 	)
 	// As per M3 best practices, creating separate all-host and per-host metrics
 	// to reduce metric cardinality when querying metrics for all hosts.
-	gateway.AllHostScope = gateway.RootScope
-	gateway.PerHostScope = gateway.RootScope.Tagged(
+	gateway.AllHostScope = gateway.RootScope.SubScope(
+		service + ".production.all-workers",
+	)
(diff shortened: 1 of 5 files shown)


runtime/gateway.go
```diff
@@ -333,10 +333,14 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 	)
 	// As per M3 best practices, creating separate all-host and per-host metrics
 	// to reduce metric cardinality when querying metrics for all hosts.
-	gateway.AllHostScope = gateway.RootScope
-	gateway.PerHostScope = gateway.RootScope.Tagged(
+	gateway.AllHostScope = gateway.RootScope.SubScope(
+		service + "." + env + ".all-workers",
```
ravirajj: You can add the service Prefix to tally.ScopeOptions instead when creating the RootScope.
```
	gateway.RootScope, gateway.scopeCloser = tally.NewRootScope(
		tally.ScopeOptions{
			Prefix:         service,
			Tags:           defaultTags,
			CachedReporter: gateway.metricsBackend,
			Separator:      tally.DefaultSeparator,
		},
		time.Duration(config.MustGetInt("metrics.flushInterval"))*time.Millisecond,
	)
```

Raynos: we still have to create two sub scopes. its the same thing. The explicit sub scope call means the prefix is in a single place instead of concatenated from two different places.

ravirajj: Sounds good.

PR #247, opened 2017-10-12
Fix bug map key cannot be non string type

oliviazhang0809:


codegen/type_converter.go
```diff
@@ -346,16 +346,22 @@ func (c *TypeConverter) genConverterForMap(
 		return err
 	}
 
+	valueStruct, isStruct := toFieldType.ValueSpec.(*compile.StructSpec)
+	sourceIdentifier := fromIdentifier
 	_, isStringKey := toFieldType.KeySpec.(*compile.StringSpec)
+
 	if !isStringKey {
-		return errors.Errorf(
```
Raynos: We need a new check here. We do not want to generate code for `map<map<string,string>, string>` 

The key should be a `string` or a typedef for a `string`.

There is a `compile.RootTypeSpec()` method that will take a typedef and give you back the actual type it is.

codegen/type_converter_test.go
```diff
@@ -1102,28 +1102,27 @@ func TestConvertWithBadImportMapOfBadStruct(t *testing.T) {
...
 	)
 
-	assert.Error(t, err)
-	assert.Equal(t, "", lines)
-	assert.Equal(t,
-		"could not convert key (one), map is not string-keyed.",
-		err.Error(),
-	)
+	assert.NoError(t, err)
+	assert.Equal(t, "out.One = make(map[structs.UUID]string, len(in.One))\nfor key, value := range in.One {\n\tout.One[ structs.UUID(key)] = string(value)\n}\nout.Two = make(map[int32]string, len(in.Two))\nfor key, value := range in.Two {\n\tout.Two[ int32(key)] = string(value)\n}", lines)
```
Raynos: for reviewability and sanity please use 

    trim(`
        // code with whitespace
    `)

ChuntaoLu:
For git commit hygiene, you may want to squash the commits of the same message or update the message if each commit does different thing.

screen shot 2017-10-12 at 3 09 59 pm

codegen/type_converter.go
```diff
@@ -346,16 +346,30 @@ func (c *TypeConverter) genConverterForMap(
...
-			"could not convert key (%s), map is not string-keyed.",
-			toField.Name,
+		_, isMapKey := toFieldType.KeySpec.(*compile.MapSpec)
+		if isMapKey {
```
Raynos: Sorry map key was an example. Doing `map<set<string>, string>` ; `map<list<string>, string` will also not work.

We probably need to check if the typedef resolves to `*compile.StringSpec` and reject every other type that's not string or typedef to string.

oliviazhang0809: basically just collections won't work right? Do you have an example how we can leverage compile.RootTypeSpec()?

Raynos: There should be examples of it. `git grep RootTypeSpec`

codegen/type_converter.go
```diff
@@ -346,16 +346,34 @@ func (c *TypeConverter) genConverterForMap(
...
-		return errors.Errorf(
-			"could not convert key (%s), map is not string-keyed.",
-			toField.Name,
-		)
+		realType := compile.RootTypeSpec(toFieldType.KeySpec)
+		switch realType.(type) {
+		case
+			*compile.MapSpec,
+			*compile.SetSpec,
+			*compile.ListSpec:
```
Raynos: We cannot support `*compile.StructSpec` either. 

I'm pretty sure `*compile.Int8Spec` , `*compile.DoubleSpec` and `*compile.BoolSpec` will also fail.

codegen/type_converter.go
```diff
@@ -346,16 +346,38 @@ func (c *TypeConverter) genConverterForMap(
...
+
 	if !isStringKey {
-		return errors.Errorf(
-			"could not convert key (%s), map is not string-keyed.",
-			toField.Name,
-		)
+		realType := compile.RootTypeSpec(toFieldType.KeySpec)
+		switch realType.(type) {
+		case
+			*compile.MapSpec,
```
Raynos: The full list is here ( https://github.com/uber/zanzibar/blob/master/codegen/thrift.go#L201-L230 )

For hygiene purposes and handling all cases it's better to panic in the `default` case and to explicitely handle the `*compile.StringSpec` case.

PR #248, opened 2017-10-12
Verify support for binary in edge-gateway

oliviazhang0809:


test/endpoints/bar/bar_metrics_test.go
```diff
@@ -66,7 +66,7 @@ func TestCallMetrics(t *testing.T) {
 	_, err = gateway.MakeRequest(
 		"POST", "/bar/bar-path", nil,
 		bytes.NewReader([]byte(`{
-			"request":{"stringField":"foo","boolField":true}
+			"request":{"stringField":"foo","boolField":true,"binaryField":"aGVsbG8="}
```
Raynos: We are sending the field here but there is no check to see if the HTTP downstream or tchannel downstream gets the correct binary.

We should have a "client" thrift file return a `binary` / `[]byte` and double check that the gateway HTTP response gets the base64.

oliviazhang0809: added a return param as binary.

Raynos: I meant we need to add more `assert.Equal()` calls to actually verify the body

PR #253, opened 2017-10-17
Add a status tag for HTTP metrics

Raynos:
By adding a tag for status we can utilize in `aliasByTags()`
for pretty printing.

This extra tag should be relatively cheap but some older metrics
will not have this tag at all.

r: @uber/zanzibar-team

### DIFF
--- a/runtime/call_metrics.go
+++ b/runtime/call_metrics.go
@@ -165,7 +165,9 @@ func NewInboundHTTPMetrics(scope tally.Scope) *InboundHTTPMetrics {
 	metrics.Errors = scope.Counter(inboundCallsErrors)
 	metrics.Status = make(map[int]tally.Counter, len(knownStatusCodes))
 	for statusCode := range knownStatusCodes {
-		metrics.Status[statusCode] = scope.Counter(fmt.Sprintf("%s.%d", inboundCallsStatus, statusCode))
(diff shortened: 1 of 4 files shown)


runtime/call_metrics.go
```diff
@@ -165,7 +165,9 @@ func NewInboundHTTPMetrics(scope tally.Scope) *InboundHTTPMetrics {
 	metrics.Errors = scope.Counter(inboundCallsErrors)
 	metrics.Status = make(map[int]tally.Counter, len(knownStatusCodes))
 	for statusCode := range knownStatusCodes {
-		metrics.Status[statusCode] = scope.Counter(fmt.Sprintf("%s.%d", inboundCallsStatus, statusCode))
+		metrics.Status[statusCode] = scope.Tagged(map[string]string{
+			"status": fmt.Sprintf("%d", statusCode),
+		}).Counter(fmt.Sprintf("%s.%d", inboundCallsStatus, statusCode))
```
ravirajj: If we are adding status as a tag, do we need the `statusCode` suffix in the counter name?

Raynos: The tag is only there because M3QL does not implement `aliasSub()`.

We can keep the statusCode in the metric name for more flexible queries

PR #254, opened 2017-10-17
Verify i64

oliviazhang0809:


test/endpoints/bar/bar_normal_test.go
```diff
@@ -206,7 +206,7 @@ func TestMalformedBarExceptionCode(t *testing.T) {
 	res, err := gateway.MakeRequest(
 		"POST", "/bar/bar-path", nil,
 		bytes.NewReader([]byte(`{
-			"request":{"stringField":"foo","boolField":true,"binaryField":"aGVsbG8="}
+			"request":{"stringField":"foo","boolField":true,"binaryField":"aGVsbG8=","timestamp":123}
```
Raynos: When testing i64 we should use values outside of the safe range

For example `9223372036854775807`.

PR #256, opened 2017-10-18
verify enum works

oliviazhang0809:


test/endpoints/bar/bar_normal_test.go
```diff
@@ -168,7 +168,7 @@ func TestBarExceptionCode(t *testing.T) {
 	res, err := gateway.MakeRequest(
 		"POST", "/bar/bar-path", nil,
 		bytes.NewReader([]byte(`{
-			"request":{"stringField":"foo","boolField":true,"binaryField":"AAD//w==","timestamp":9223372036854775807}
+			"request":{"stringField":"foo","boolField":true,"binaryField":"AAD//w==","timestamp":9223372036854775807,"enumField":0}
```
Raynos: Let's also send "APPLE" to verify that the parse accepts both APPLE and 0

PR #257, opened 2017-10-18
Update tchannel metrics to remove bad characters

Raynos:
Currently our tchannel metrics are not being collected
because the tag values we send have characters that are not allowed,
in this case ":" ( allowed is /[a-zA-Z_-]/.

This change should hopefully make tchannel metrics available
for querying.

r: @uber/zanzibar-team

### DIFF
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -88,10 +88,9 @@ func NewTChannelClient(
 			zap.String("serviceMethod", serviceMethod),
 		)
 		metrics[serviceMethod] = NewOutboundTChannelMetrics(scope.Tagged(map[string]string{
-			"client":          opt.ClientID,
-			"method":          methodName,
-			"target-service":  opt.ServiceName,
-			"target-endpoint": serviceMethod,
+			"client":         opt.ClientID,
+			"method":         methodName,
+			"target-service": opt.ServiceName,
 		}))
 	}
 
--- a/runtime/tchannel_server.go
+++ b/runtime/tchannel_server.go
@@ -112,10 +112,13 @@ func NewTChannelEndpointWithPostResponseCB(
 		zap.String("handlerID", handlerID),
 		zap.String("method", method),
 	)
+
+	cleanMethod := strings.Replace(method, ":", "-", -1)
+
 	scope = scope.Tagged(map[string]string{
 		"endpoint": endpointID,
 		"handler":  handlerID,
-		"method":   method,
+		"method":   cleanMethod,
 	})
 	return &TChannelEndpoint{
 		EndpointID: endpointID,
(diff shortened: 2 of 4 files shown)


test/endpoints/tchannel/baz/baz_metrics_test.go
```diff
@@ -158,7 +158,7 @@ func TestCallMetrics(t *testing.T) {
 		"service":  "test-gateway",
 		"endpoint": "bazTChannel",
 		"handler":  "call",
-		"method":   "SimpleService::Call",
+		"method":   "SimpleService--Call",
```
ravirajj: How about just using single dash?

runtime/tchannel_client.go
```diff
@@ -88,10 +88,9 @@ func NewTChannelClient(
 			zap.String("serviceMethod", serviceMethod),
 		)
 		metrics[serviceMethod] = NewOutboundTChannelMetrics(scope.Tagged(map[string]string{
-			"client":          opt.ClientID,
-			"method":          methodName,
-			"target-service":  opt.ServiceName,
-			"target-endpoint": serviceMethod,
```
ravirajj: Can we sanitize `serviceMethod` too?

runtime/tchannel_server.go
```diff
@@ -112,10 +112,13 @@ func NewTChannelEndpointWithPostResponseCB(
 		zap.String("handlerID", handlerID),
 		zap.String("method", method),
 	)
+
+	cleanMethod := strings.Replace(method, ":", "-", -1)
```
ravirajj: Perhaps move this to util method, so it can be used by tchannel client and server.
In fact, I think tally already has a sanitize method.

Raynos: tally has a sanitize class thats more complicated then needed. A string replace is sufficient here.

runtime/tchannel_client.go
Raynos: sure.

runtime/gateway.go
```diff
@@ -325,9 +325,10 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 	}
 	gateway.RootScope, gateway.scopeCloser = tally.NewRootScope(
 		tally.ScopeOptions{
-			Tags:           defaultTags,
-			CachedReporter: gateway.metricsBackend,
-			Separator:      tally.DefaultSeparator,
+			Tags:            defaultTags,
+			CachedReporter:  gateway.metricsBackend,
+			Separator:       tally.DefaultSeparator,
+			SanitizeOptions: &m3.DefaultSanitizerOpts,
```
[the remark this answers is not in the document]
ravirajj: Or update `uber-go/tally` `m3.DefaultSanitizerOpts` instead of creating another opt.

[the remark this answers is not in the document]
Raynos: ... It whitelists `.` `-` and `_` . Its whitelist, not a blacklist.

`:` is excluded because its not in the whitlist. no changes are needed.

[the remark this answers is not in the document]
ravirajj: Yes. Just looked at the tally sanitize function.  It's a whitelist of valid characters. Any other character is replaced.

PR #259, opened 2017-10-19
Add panic handler with logs / metrics.

Raynos:
Currently we have panics in production that are only
visible in the STDOUT logs of our docker container.

This panic handler will ensure we write to kafka/metrics.

This is only for panics in the goroutine for the http proxy.

Any panics that occur in other goroutines ( the main goroutine
or any goroutine spawned by any other code ) will cause the
process to Exit() to the operating system and then cause the
deployment software to restart it.

r: @uber/zanzibar-team

### DIFF
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 
 	"github.com/julienschmidt/httprouter"
+	"github.com/pkg/errors"
 	"github.com/uber-go/tally"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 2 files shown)


runtime/router.go
```diff
@@ -99,9 +100,11 @@ func (endpoint *RouterEndpoint) HandleRequest(
 
 // HTTPRouter data structure to handle and register endpoints
 type HTTPRouter struct {
+	gateway                  *Gateway
```
ravirajj: Nitpick: you could just save the `logger` here, since only the gateway logger is used.

PR #262, opened 2017-10-23
Bump easyjson: support integer map keys

ChuntaoLu:
This PR bumps easyjson for integer map key support added in https://github.com/mailru/easyjson/pull/147 and adds a use case in bar client thrift.


ChuntaoLu:
@Raynos For testing easyjson itself? If so https://github.com/mailru/easyjson/pull/147/files#diff-57c3f0c7bfe20da14e1f619d008720b4

Raynos:
No another echo test in `./test/clients/...` like all the other echo tests.

PR #263, opened 2017-10-23
thrift module validator 

cl1337:
This PR improves user input validation passed in from UI

- added validation before codegen
- BFS walk & compare thrift module ASTs between client and endpoint
- throw user friendly errors

cc @uber/zanzibar-team 

===========
I want to reiterate why we don't want generate logic in validation:
- currently there's no easy way to walk through the whole generate process for validation only, the generate process is scripting on file system, cannot be plugin in to endpoint call without major refactor
- full generate process yield intermediate result, leave error stack confusing for users
- deep in generate stack, there's no obvious lead to exactly what user input was wrong

My original plan was to visit & check separately *on user input*, and leave other errors to ci system log, my rationale is: all user related errors should be able to detect from user input only, if the error is hiding deep that we need to generate to detect, we should probably debug / fix our gen system instead of depend on user to fix input. 

Since we have concern validation diverge with code gen, I'm going to call codegen in validation 
anyway, we need to change codegen, raise errors that links to exactly which user input and why, this require 
[-] 1: error type shared between codegen & backend
2: refactor in type convert

==============

synced offline with @Raynos 
1: backend is already touching file system for diff gen, this is not optimal but consistent with existing structure
2: we're deferring the full generate process to ci jobs, generate -> type convert path we have a way to not generating intermediate result
3: is the only thing we need -> require changes in type converter

### DIFF
--- a/backend/repository/handler.go
+++ b/backend/repository/handler.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/pkg/errors"
+	reqerr "github.com/uber/zanzibar/codegen/errors"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
@@ -385,6 +386,10 @@ func (h *Handler) CreateDiff(w http.ResponseWriter, r *http.Request, ps httprout
 		h.WriteErrorResponse(w, http.StatusInternalServerError, err)
 		return
 	}
+	if err := h.Manager.Validate(repo, req); err != nil {
+		h.WriteErrorResponse(w, http.StatusInternalServerError, err)
+		return
+	}
 	if _, err := repo.LatestGatewayConfig(); err != nil {
 		h.WriteErrorResponse(w, http.StatusInternalServerError, errors.Wrap(err, "invalid gateway config after upadte"))
 		return
@@ -472,7 +477,7 @@ func (h *Handler) WriteErrorResponse(w http.ResponseWriter, code int, err error)
 	errJSON := fmt.Sprintf("{\"error\": %q}", err.Error())
 	causeErr := errors.Cause(err)
 	// In case we know the real cause of the error.
-	if e, ok := causeErr.(*RequestError); ok {
+	if e, ok := causeErr.(*reqerr.RequestError); ok {
 		code = http.StatusBadRequest
 		errJSON = e.JSON()
 		causeErr = e.Cause
--- a/backend/repository/client_update.go
+++ b/backend/repository/client_update.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/uber/zanzibar/codegen"
+	reqerr "github.com/uber/zanzibar/codegen/errors"
 )
 
 // ClientJSONConfig defines the JSON content of the configuration file.
@@ -98,14 +99,16 @@ func (r *Repository) UpdateClientConfigs(req *ClientConfig, clientCfgDir, thrift
 func allExposedMethods(thriftServices map[string]map[string]*ThriftService, thriftFile string) (map[string]string, error) {
 	serviceMap, ok := thriftServices[thriftFile]
 	if !ok {
-		return nil, errors.Errorf("thrift file %q not found", thriftFile)
+		return nil, reqerr.NewRequestError(
+			reqerr.ClientsThriftFile, errors.Errorf("endpoint thrift file %q not found", thriftFile))
 	}
 	exposedMethods := map[string]string{}
 	for _, tservice := range serviceMap {
 		for _, method := range tservice.Methods {
 			exposedName := tservice.Name + "::" + method.Name
 			if pre, ok := exposedMethods[method.Name]; ok {
-				return nil, errors.Errorf("duplicated method name for %q and %q", pre, exposedName)
+				return nil, reqerr.NewRequestError(
+					reqerr.ClientsExposedMethods, errors.Errorf("duplicated method name for %q and %q", pre, exposedName))
 			}
 			exposedMethods[strings.Title(method.Name)] = exposedName
 		}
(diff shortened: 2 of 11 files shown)


backend/repository/validator.go
```diff
@@ -0,0 +1,378 @@
...
+	// check len(thriftFile) > 0
+	clientModule := thriftModules[clientCfg.ThriftFile]
+	endpointModule := thriftModules[endpointCfg.ThriftFile]
+	// endpoint method should be exposed from client
+	if exposedMethod, ok := clientCfg.ExposedMethods[endpointCfg.ClientMethod]; !ok || exposedMethod != endpointCfg.ThriftMethodName {
+		return NewRequestError(
+			ClientsExposedMethods, errors.New("methods not exposed in proxy client"))
+	}
+	funcPath := strings.Split(endpointCfg.ThriftMethodName, "::")
+	clientFunc, err := ResolveNode(funcPath, clientModule)
```
Raynos: This does not seem correct. it should respect `endpointCfg.ClientMethod` not `endpointCfg.ThriftMethodName`.

The `ThriftMethodName` is for `endpointModule`; the `clientCfg.ExposedMethod[endpointCfg.ClientMethod]` should be used for `clientModule`

backend/repository/validator.go
```diff
@@ -0,0 +1,378 @@
...
+	if err != nil {
+		return NewRequestError(EndpointThriftMethodName, errors.New("method not found in endpoint"))
+	}
+	visitor := NewSyntaxVisitor(clientFunc, endpointFunc)
+	return visitor.Visit()
+}
+
+// ResolveNode resolves a node in thrift AST given a string path
+// example: [SimpleService, compare]
+func ResolveNode(nodePaths []string, root *compile.Module) (interface{}, error) {
```
Raynos: we store re-usable functions for thrift in `codegen/thrift.go` ; might be useful to place this there.

backend/repository/validator.go
```diff
@@ -0,0 +1,378 @@
...
+	if exposedMethod, ok := clientCfg.ExposedMethods[endpointCfg.ClientMethod]; !ok || exposedMethod != endpointCfg.ThriftMethodName {
+		return NewRequestError(
+			ClientsExposedMethods, errors.New("methods not exposed in proxy client"))
+	}
+	funcPath := strings.Split(endpointCfg.ThriftMethodName, "::")
+	clientFunc, err := ResolveNode(funcPath, clientModule)
+	if err != nil {
+		return NewRequestError(EndpointThriftMethodName, errors.New("method not found in client"))
+	}
+	endpointFunc, err := ResolveNode(funcPath, endpointModule)
```
Raynos: The return type is an interface here. useful to double check this is actually a `compile.FunctionSpec`

backend/repository/validator.go
```diff
@@ -0,0 +1,378 @@
...
+	if err != nil {
+		return NewRequestError(EndpointThriftMethodName, errors.New("method not found in endpoint"))
+	}
+	visitor := NewSyntaxVisitor(clientFunc, endpointFunc)
```
Raynos: We already have a walker function ( https://github.com/uber/zanzibar/blob/master/codegen/thrift.go#L171 ) that can be used. 

Would be nice to re-use this instead of reimplementing a walker.

backend/repository/validator.go
```diff
@@ -0,0 +1,378 @@
...
+		client   interface{}
+		endpoint interface{}
+		err      error
+	)
+	for {
+		client, endpoint, err = sv.nodeQ.DeQ()
+		if err != nil {
+			return nil
+		}
+		if reflect.TypeOf(client) != reflect.TypeOf(endpoint) {
```
Raynos: I don't think this hold true when we implement rewrite rules ( with transform middleware ).

backend/repository/validator.go
```diff
@@ -0,0 +1,378 @@
...
+	sv.nodeQ.EnQ(client.ArgsSpec, endpoint.ArgsSpec)
+	sv.nodeQ.EnQ(client.ResultSpec, endpoint.ResultSpec)
+	return nil
+}
+
+// VisitArgsSpec visit and syntax check client/endpoint argsSpec
+func (sv *SyntaxVisitor) VisitArgsSpec(clientArgs compile.ArgsSpec, endpointArgs compile.ArgsSpec) error {
+	clientArgList := compile.FieldGroup(clientArgs)
+	endpointArgList := compile.FieldGroup(endpointArgs)
+	if len(clientArgList) != len(endpointArgList) {
```
Raynos: this check is too strict, the client can have extra (optional) arguments.

In fact we convert arguments by argument name so its possible for them to have the same length but different names.

backend/repository/validator.go
```diff
@@ -0,0 +1,378 @@
...
+	return nil
+}
+
+// RuleProxyConsistency validates client / endpoint proxy consistency
+type RuleProxyConsistency struct{}
+
+func (r *RuleProxyConsistency) validate(clientCfgs []ClientConfig, endpointCfgs []EndpointConfig, thriftModules map[string]*compile.Module) error {
+	// trigger
+	if len(clientCfgs) != 1 || len(endpointCfgs) != 1 {
+		return nil
```
ChuntaoLu: shouldn't this be an error?

cl1337:
thanks team for the review, after sync offline, will refactor validator & request error into codegen/type_converter instead separate walk in validator

backend/repository/handler.go
```diff
@@ -361,6 +362,10 @@ func (h *Handler) CreateDiff(w http.ResponseWriter, r *http.Request, ps httprout
 		h.WriteErrorResponse(w, http.StatusInternalServerError, err)
 		return
 	}
+	if err := h.Manager.Validate(repo, req); err != nil {
```
Raynos: This will generate code and therefore when we call `h.DiffCreator.LandDiff()` the diff that gets created will now contain generated code instead of only config files.

I think this `Validate()` call is fine. We just need to update the implementation of `DiffCreater#LandDiff` in our internal codebase to somehow `git reset --hard origin/master -- build/` or something like that.

cl1337: good catch, as synced offline, will add a flag in generate to not alter files in the diff

as discussed, code generate should happen after diff land, just not here :)

common/errors.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package repository
+package common
```
Raynos: Consider using `codegen/errors/errors.go` instead of `common/errors.go` to reduce number of top level folders in the git repo.

PR #264, opened 2017-10-24
Exposing logger and scope to generated and custom client

ravirajj:
- Creating TChannelClient struct to expose loggers/scopes with appropriate fields/tags.
- Replacing TChannelClient interface with TChannelCaller interface.

### DIFF
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -9,9 +9,10 @@ import (
 	"time"
 
 	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
+
 	zanzibar "github.com/uber/zanzibar/runtime"
-	"github.com/uber/tchannel-go"
+	tchannel "github.com/uber/tchannel-go"
+
 	module "{{$instance.PackageInfo.ModulePackagePath}}"
 	{{range $idx, $pkg := .IncludedPackages -}}
 	{{$pkg.AliasName}} "{{$pkg.PackageName}}"
@@ -147,7 +148,7 @@ type {{$clientName}} struct {
 			}
 		}
 		if err != nil {
-			logger.Warn("{{$clientName}}.{{$methodName}} ({{$serviceMethod}}) returned error", zap.Error(err))
+			logger.Warn("Client call returned error", zap.Error(err))
 		{{if eq .ResponseType "" -}}
 			return nil, err
 		{{else -}}
@@ -160,7 +161,7 @@ type {{$clientName}} struct {
 		{{else -}}
 			resp, err = {{.GenCodePkgName}}.{{title $svc.Name}}_{{title .Name}}_Helper.UnwrapResponse(&result)
 			if err != nil {
-				logger.Error("Unable to unwrap {{$clientName}}.{{$methodName}} ({{$serviceMethod}}) response", zap.Error(err))
+				logger.Warn("Unable to unwrap client response", zap.Error(err))
 			}
 			return resp, respHeaders, err
 		{{end -}}
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -148,7 +148,7 @@ type {{$clientName}} struct {
 			}
 		}
 		if err != nil {
-			logger.Warn("Client call returned error", zap.Error(err))
+			logger.Warn("TChannel client call returned error", zap.Error(err))
 		{{if eq .ResponseType "" -}}
 			return nil, err
 		{{else -}}
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -8,12 +8,10 @@ import (
 	"strconv"
 	"time"
 
-	"go.uber.org/zap"	
+	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-
 	zanzibar "github.com/uber/zanzibar/runtime"
 	"github.com/uber/tchannel-go"
-
 	module "{{$instance.PackageInfo.ModulePackagePath}}"
 	{{range $idx, $pkg := .IncludedPackages -}}
 	{{$pkg.AliasName}} "{{$pkg.PackageName}}"
@@ -103,14 +101,12 @@ func {{$exportName}}(
 
 	return &{{$clientName}}{
 		client: client,
-		logger: deps.Default.Logger,
 	}
 }
 
 // {{$clientName}} is the TChannel client for downstream service.
 type {{$clientName}} struct {
-	client        zanzibar.TChannelClient
-	logger *zap.Logger
+	client *zanzibar.TChannelClient
 }
 
 {{range $svc := .Services}}
@@ -118,7 +114,7 @@ type {{$clientName}} struct {
 {{$serviceMethod := printf "%s::%s" $svc.Name .Name -}}
 {{$methodName := (title (index $exposedMethods $serviceMethod)) -}}
 {{if $methodName -}}
-	// {{$methodName}} is a client RPC call for method "{{$svc.Name}}::{{.Name}}"
+	// {{$methodName}} is a client RPC call for method "{{$serviceMethod}}"
 	func (c *{{$clientName}}) {{$methodName}}(
 		ctx context.Context,
 		reqHeaders map[string]string,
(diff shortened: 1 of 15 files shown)


examples/example-gateway/clients/quux/quux.go
```diff
@@ -21,16 +21,49 @@
...
+	"go.uber.org/zap"
+	"golang.org/x/net/context"
 )
 
 // Client is a custom client that does nothing yet
 type Client interface{}
 
-type quux struct{}
+type quux struct {
+	client *zanzibar.TChannelClient
```
ChuntaoLu: This client is meant to be a demo for a custom client. It doesn't make sense to fill in a tchannel client which should be defined in thrift.

codegen/templates/tchannel_client.tmpl
```diff
@@ -161,6 +159,9 @@ type {{$clientName}} struct {
 			return respHeaders, err
 		{{else -}}
 			resp, err = {{.GenCodePkgName}}.{{title $svc.Name}}_{{title .Name}}_Helper.UnwrapResponse(&result)
+			if err != nil {
```
Raynos: not all errors are `.Error()` ; default to `.Warn()` here instead.

codegen/templates/tchannel_client.tmpl
```diff
@@ -150,6 +147,7 @@ type {{$clientName}} struct {
 			}
 		}
 		if err != nil {
+			logger.Warn("{{$clientName}}.{{$methodName}} ({{$serviceMethod}}) returned error", zap.Error(err))
```
Raynos: we do not put meta data in the log message. all of these should be fields the log string needs to be static.

codegen/templates/tchannel_client.tmpl
```diff
@@ -161,6 +159,9 @@ type {{$clientName}} struct {
 			return respHeaders, err
 		{{else -}}
 			resp, err = {{.GenCodePkgName}}.{{title $svc.Name}}_{{title .Name}}_Helper.UnwrapResponse(&result)
+			if err != nil {
+				logger.Error("Unable to unwrap {{$clientName}}.{{$methodName}} ({{$serviceMethod}}) response", zap.Error(err))
```
Raynos: We do not want a dynamic log string here. Put them all in fields.

examples/example-gateway/clients/quux/quux.go
```diff
@@ -21,16 +21,49 @@
...
+	zanzibar "github.com/uber/zanzibar/runtime"
+	"go.uber.org/zap"
+	"golang.org/x/net/context"
 )
 
 // Client is a custom client that does nothing yet
 type Client interface{}
 
-type quux struct{}
+type quux struct {
```
Raynos: I dont think this clients needs changing...

examples/example-gateway/endpoints/bar/endpoint-config.json
```diff
@@ -20,7 +20,8 @@
 	},
 	"dependencies": {
 		"client": [
-			"bar"
+			"bar",
+			"quux"
```
Raynos: why this change ?

runtime/tchannel_client.go
```diff
@@ -53,12 +55,13 @@ type tchannelClient struct {
 	timeout           time.Duration
 	timeoutPerAttempt time.Duration
 	routingKey        *string
-	loggers           map[string]*zap.Logger
+	Loggers           map[string]*zap.Logger
+	Scopes            map[string]tally.Scope
```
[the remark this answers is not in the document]
ravirajj: Yes, I'm debating which approach to take.  Thought I'd put up this diff for feedback.

runtime/tchannel_client.go
```diff
@@ -43,8 +43,10 @@ type TChannelClientOption struct {
 	RoutingKey        *string
 }
 
-// tchannelClient implements TChannelClient and makes outgoing Thrift calls.
-type tchannelClient struct {
+// TChannelClient implements TChannelCaller and makes outgoing Thrift calls.
+type TChannelClient struct {
+	TChannelCaller
```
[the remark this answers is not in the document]
ravirajj: Trying to see if go made sure the struct implements the interface, but that may not be the case.

codegen/templates/tchannel_client.tmpl
ravirajj: Actually, they might be already there, since the logger scoped the method.

Raynos: if you want to drop them then that's fine too. We still want a static string non-dynamic here.

runtime/tchannel_client.go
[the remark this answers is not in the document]
ravirajj: @clu One problem with having TChannelHandler interface that exposes Logger and Scope methods is that it will need to expose the endpoint.EndpointID, endpoint.HandlerID and endpoint.Method too.  Similarly for TChannelClient too, it would need add ClientID, etc to the interface too.  Using struct makes it easier to expose these fields and the Logger and Scope field too.

codegen/templates/tchannel_client.tmpl
```diff
@@ -8,12 +8,10 @@ import (
 	"strconv"
 	"time"
 
-	"go.uber.org/zap"	
+	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-
```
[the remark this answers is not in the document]
ravirajj: Done.

codegen/templates/tchannel_client.tmpl
```diff
@@ -8,12 +8,10 @@ import (
...
 	zanzibar "github.com/uber/zanzibar/runtime"
 	"github.com/uber/tchannel-go"
-
```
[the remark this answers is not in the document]
ravirajj: Done.

codegen/templates/tchannel_client.tmpl
ravirajj: Done.

codegen/templates/tchannel_client.tmpl
ravirajj: Done.

codegen/templates/tchannel_client.tmpl
ravirajj: Done.

examples/example-gateway/clients/quux/quux.go
ravirajj: Just using it as an example to demonstrate how to create a custom tchannel client, but I can remove it.

examples/example-gateway/endpoints/bar/endpoint-config.json
ravirajj: Same. To demostrate creating a custom tchannel client.

runtime/tchannel_client.go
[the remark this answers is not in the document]
ChuntaoLu: why `EndpointID`, `HandlerID` and `Method` are exposed on the client? `ClientID` seems reasonable, what else need to be exposed for the client?

examples/example-gateway/endpoints/bar/endpoint-config.json
Raynos: understood, got it.

codegen/templates/tchannel_client.tmpl
```diff
@@ -147,7 +148,7 @@ type {{$clientName}} struct {
 			}
 		}
 		if err != nil {
-			logger.Warn("{{$clientName}}.{{$methodName}} ({{$serviceMethod}}) returned error", zap.Error(err))
+			logger.Warn("Client call returned error", zap.Error(err))
```
Raynos: nitpick: "TChannel Client call...`

runtime/tchannel_client.go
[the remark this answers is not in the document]
ravirajj: To clarify:

The `TChannelEndpoint` would expose the `TChannelHandler`, `Logger()`, `Scope()`, `EndpointID`, `HandlerID` and `Method`.

The `TChannelClient` would expose the `TChannelCaller`, `Logger(method)`, `Scope(method)` and `ClientID`.

runtime/tchannel_client.go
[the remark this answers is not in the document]
ChuntaoLu: You already implemented `Call` on `TChannelClient`, https://github.com/uber/zanzibar/pull/264/files/87083f25e7e5ada3e823cb8e13488306973590a7#diff-4eed9ca277749573af3f65a5ffd21480R120

PR #265, opened 2017-10-27
Add a new endpoint that takes in raw thrift code and converts it Modu…

fymemon:
…le structure. Changes to use ast structure instead of compile structure from thriftrw.

### DIFF
--- a/backend/repository/data/handler/test_cases.json
+++ b/backend/repository/data/handler/test_cases.json
@@ -861,6 +861,119 @@
 		}
 	},
 	{
+		"name": "get the module structure of thrift code",
+		"request": {
+			"method": "POST",
+			"url": "/thrift-file-code/clients/baz/baz.thrift",
+			"header": {
+				"gateway-id": "example-gateway"
+			},
+			"body": {
+				"content": "namespace java com.uber.zanzibar.clients.baz\ninclude \"base.thrift\"\ntypedef base.UUID SomeThing\nconst SomeThing theThing = 'ABC'\n"
+			}
+		},
+		"response": {
+			"code": 200,
+			"body": {
+				"constants": {
+					"theThing": {
+						"file": "clients/baz/baz.thrift",
+						"line": 4,
+						"name": "theThing",
+						"type": {
+							"name": "SomeThing"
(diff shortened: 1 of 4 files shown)


backend/repository/thrift_compile.go
```diff
@@ -174,19 +174,232 @@ type ResultSpec struct {
...
+			}
+		case *ast.Service:
+			services[definition.Name] = &ServiceSpec{
+				Name:        definition.Name,
+				File:        thriftFile,
+				Line:        definition.Line,
+				Functions:   convertAstFunctions(definition.Functions),
+				Annotations: convertAstAnnotations(definition.Annotations),
+			}
+		}
```
Raynos: do we want to add a default case where we store a `&TypeSpec` in `types`  with `{ Name, File, Line, Annotations }` ?

backend/repository/data/handler/test_cases.json
```diff
@@ -999,45 +1130,19 @@
 									{
 										"default": "Fruit.APPLE",
 										"id": 1,
+										"line": 83,
 										"name": "arg",
-										"required": false,
+										"required": true,
```
Raynos: The required boolean changed here, we should double check this.

backend/repository/data/handler/test_cases.json
```diff
@@ -1141,23 +1256,19 @@
 								"args_spec": [
 									{
 										"id": 1,
+										"line": 100,
 										"name": "arg",
 										"required": true,
 										"type": {
-											"list_value_type": {
```
Raynos: Does the frontend UI support rendering lists without `list_value_type` ? Do you parse `type.name` instead ?

backend/repository/data/handler/test_cases.json
```diff
@@ -1166,283 +1277,125 @@
 								"args_spec": [
 									{
 										"id": 1,
+										"line": 108,
 										"name": "arg",
 										"required": true,
 										"type": {
-											"map_type": {
```
Raynos: `map_type` data also seems really useful for rendering in the frontend UI.

PR #266, opened 2017-10-27
Add thriftrw change to enable i64 marshal/unmarshal to Date/Long

oliviazhang0809:


examples/example-gateway/idl/endpoints/bar/bar.thrift
```diff
@@ -3,19 +3,27 @@ namespace java com.uber.zanzibar.clients.bar
...
-typedef i64 (js.type = 'Date') Timestamp
+typedef i64 Timestamp (json.type = "Date")
+typedef i64 (json.type = "Long") Long
 
 enum Fruit {
     APPLE,
     BANANA
 }
 
+struct I64TestField {
```
Raynos: too many files change every time we add a new field to `BarRequest`. Can we add a new parameter to `tooManyArgs()` method instead so that less files change ?

PR #267, opened 2017-10-28
actually expose middlewares config in backend

jakelarkn:


backend/repository/handler_test.go
```diff
@@ -68,7 +68,7 @@ func TestHandlers(t *testing.T) {
 		for _, tc := range testCases {
 			tc := tc // Capture range variable
 			t.Run(tc.Name, func(t *testing.T) {
-				t.Parallel()
+				//t.Parallel()
```
jakelarkn: i think reading the middleware config files or doing schema validation concurrently;  breaks parallelism. It just hangs for some reason.  open to suggestions.   This runs quickly even in serial  so I think it might be an optimization we can leave out for the the time being

Raynos: dont need parallel in tests for now but there might be race conditions since we handle http requests in parallel.

We should look into this.

backend/repository/thrift_compile.go
```diff
@@ -454,8 +454,8 @@ func (m *Module) Code() string {
 	sort.Sort(codeBlockSlice(codeBlocks))
 	result := bytes.NewBuffer(nil)
 	for i := range codeBlocks {
-		result.WriteString(codeBlocks[i].Code)
-		result.WriteString("\n\n")
+		_, _ = result.WriteString(codeBlocks[i].Code)
```
Raynos: thanks for fixing this.

backend/repository/config.go
```diff
@@ -226,6 +227,10 @@ func (r *Repository) newGatewayConfig() (configuration *Config, cfgErr error) {
 		return nil, errors.Wrap(err, "failed to read client configuration")
 	}
 	config.Endpoints = r.endpointConfigs(config.ThriftRootDir, gatewaySpec)
+
+	if config.Middlewares, err = r.middlewareConfigs(config.MiddlewareConfigDir); err != nil {
```
Raynos: This is a "configDir" but later on you call `readFile()` on it.

backend/repository/config.go
```diff
@@ -339,6 +344,38 @@ func (r *Repository) endpointConfigs(thriftRootDir string, gatewaySpec *codegen.
...
+	if middlewareConfigFile == "" {
+		return cfgs, nil
+	}
+	bytes, err := r.ReadFile(middlewareConfigFile)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(bytes) == 0 {
+		panic("EMPTY MIDDLEWARE CONFIG")
```
Raynos: just return an error here ? Maybe `r.Logger.Error()` ?

PR #269, opened 2017-11-01
Add managed thrift files to CreateDiff()

Raynos:
This means we can create new thrift files checked into
edge-gateway when we try to create the diff.

This allows the UI editor component for creating a thrift file
to save these thrift files.

r: @uber/zanzibar-team


backend/repository/thrift_update.go
```diff
@@ -45,6 +45,39 @@ func (r *Repository) ThriftConfig(idlRoot string) (map[string]*ThriftMeta, error
 	return config, nil
 }
 
+// WriteManagedThriftFiles will write new thrift files to disk
+func (r *Repository) WriteManagedThriftFiles(files []ManagedThriftFile) error {
```
cl1337: would be nice to throw some trackable errors, `NewRequestError( mfile.filename)` maybe?

backend/repository/manager.go
```diff
@@ -229,15 +244,25 @@ func (m *Manager) UpdateEndpoints(r *Repository, endpointCfgDir string, req []En
...
-	EndpointUpdates []EndpointConfig `json:"endpoint_updates"`
+	ThriftFiles        []string            `json:"thrift_files"`
+	ManagedThriftFiles []ManagedThriftFile `json:"managed_thrift_files"`
+	ClientUpdates      []ClientConfig      `json:"client_updates"`
+	EndpointUpdates    []EndpointConfig    `json:"endpoint_updates"`
 }
 
 // UpdateAll writes the updates for thrift files, clients and endpoints.
 func (m *Manager) UpdateAll(r *Repository, clientCfgDir, endpointCfgDir string, req *UpdateRequest) error {
+	if err := m.UpdateManagedThriftFiles(r, req.ManagedThriftFiles); err != nil {
```
cl1337: it would be nice to consolidate this with `UpdateThriftFiles`
thrift either from fetching idl `m.thriftMetaInIDLRegistry(paths)`, or direct user input, the rest of the code path could be reuse? (dependencies ... etc)

what if I put a `include` from idl in my input thrift?

Raynos: I plan to add include support in a seperate PR. the way include support will work is slightly different because this file that is written does not yet exist in `idl-registry` so leveraging `idl-registry` code is a bit harder.

backend/repository/thrift_update.go
Raynos: sure, will change.

PR #271, opened 2017-11-08
Fixing logger error, clientID fields

ravirajj:


runtime/client_http_request.go
```diff
@@ -90,8 +90,6 @@ func (req *ClientHTTPRequest) CheckHeaders(expected []string) error {
 		if headerValue == "" {
 			req.Logger.Warn("Got outbound request without mandatory header",
 				zap.String("headerName", headerName),
-				zap.String("clientId", req.ClientID),
-				zap.String("methodName", req.MethodName),
```
ChuntaoLu: Are these two fields not useful?

ravirajj: They are already set in the req.Logger.

Issue #275, opened 2017-11-11
Change logging levels for zap child logger


Raynos:
Sometimes we want to pass a child logger with different logger levels to sub systems.

We can do this with something like :

```
childLogger := logger.With(
   zap.String("library", "{{.LibraryName}}"),
).WithOptions(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
    copy := c.With()
    copy.LevelEnabler = zap.NewAtomicLevel(zapcore.WarnLevel)
    return copy
})
```

We would need to test this but it should allow us to have a child logger with a "library=X" field and with a different minimum logging level.

cc @srivastavankit @ChuntaoLu

PR #276, opened 2017-11-11
Use configurable sub loggers for different components

ChuntaoLu:
This PR addresses https://github.com/uber/zanzibar/issues/275.

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -86,6 +86,7 @@ type Gateway struct {
 	runtimeMetrics  RuntimeMetricsCollector
 	logEncoder      zapcore.Encoder
 	logWriter       zapcore.WriteSyncer
+	logWriteSyncer  zapcore.WriteSyncer
 	httpServer      *HTTPServer
 	localHTTPServer *HTTPServer
 	tchannelServer  *tchannel.Channel
@@ -431,9 +432,7 @@ func (gateway *Gateway) setupLogger(config *StaticConfig) error {
 	)
 
 	gateway.logEncoder = logEncoder
-	if gateway.logWriter == nil {
-		gateway.logWriter = output
-	}
+	gateway.logWriteSyncer = output
 
 	// Default to a STDOUT logger
 	gateway.Logger = zapLogger.With(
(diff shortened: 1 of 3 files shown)


runtime/gateway.go
```diff
@@ -443,9 +443,10 @@ func (gateway *Gateway) initJaegerConfig(config *StaticConfig) *jaegerConfig.Con
 }
 
 func (gateway *Gateway) setupTracer(config *StaticConfig) error {
+	infoToDebug := config.MustGetBoolean("tchannel.infoToDebug")
 	opts := []jaegerConfig.Option{
 		// TChannel logger implements jaeger logger interface
-		jaegerConfig.Logger(NewTChannelLogger(gateway.Logger)),
+		jaegerConfig.Logger(NewTChannelLogger(gateway.Logger, infoToDebug)),
```
Raynos: its weird to downgrade jaeger here.

runtime/gateway.go
```diff
@@ -443,9 +443,10 @@ func (gateway *Gateway) initJaegerConfig(config *StaticConfig) *jaegerConfig.Con
 }
 
 func (gateway *Gateway) setupTracer(config *StaticConfig) error {
+	infoToDebug := config.MustGetBoolean("tchannel.infoToDebug")
```
Raynos: Use "jaeger.infoToDebug"

runtime/gateway.go
```diff
@@ -407,16 +420,21 @@ func (gateway *Gateway) setupLogger(config *StaticConfig) error {
...
 	)
 	zapLogger := zap.New(
 		NewInstrumentedZapCore(
 			prodCore, gateway.AllHostScope,
 		),
 	)
 
+	gateway.logEncoder = logEncoder
+	if gateway.logWriter == nil {
+		gateway.logWriter = output
```
Raynos: need to always assign this to `output`

In fact; let's not re-use the `logWriter` field and just add a new top level field for `gateway.logWriteSyncer` or something

ChuntaoLu: Done. Curious why do you think it's better to not reuse `logWriter`?

PR #277, opened 2017-11-12
i64 marshal/unmarshal to Timestamp/Long

oliviazhang0809:


examples/example-gateway/build/gen-code/clients/contacts/contacts/types_i64.go
```diff
@@ -0,0 +1 @@
+package contacts
```
Raynos: It looks like we generate unneeded empty `types_i64.go` files. We want to somehow not do that.

scripts/resolve_i64/main.go
```diff
@@ -0,0 +1,159 @@
...
+	sort.Sort(I64Structs(meta.Types))
+	tmpl, err := template.ParseFiles(filepath.Join(getDirName(), templateFileName))
+	if err != nil {
+		fmt.Fprintln(os.Stderr, fmt.Errorf(
+			"error parsing template %s: %s", templateFileName, err.Error()))
+		os.Exit(1)
+		return
+	}
+	tplBuffer := bytes.NewBuffer(nil)
+	err = tmpl.Execute(tplBuffer, meta)
```
Raynos: if `len(meta.Types) == 0` we can just end here and not do any templating at all.

scripts/resolve_i64/main.go
```diff
@@ -0,0 +1,159 @@
...
+func main() {
+	thriftFile := os.Args[1]
+	module, err := compile.Compile(thriftFile)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to parse thrift file: %s", thriftFile))
+	}
+	meta := &Meta{}
+
+	s := strings.TrimSuffix(thriftFile, ".thrift")
+	s = strings.Replace(s, "/idl/", "/build/gen-code/", 1)
```
Raynos: we can leave a TODO here. `/build/gen-code/` is hardcoded and we might want to customize that; maybe via `os.Args[2]`

examples/example-gateway/build/gen-code/clients/contacts/contacts/types_i64.go
oliviazhang0809: sounds good

PR #282, opened 2017-11-17
TypeConverter: use unique variable names

Raynos:
We need to make sure all identifiers are unique so that
we dont use the wrong identifier in nested loops.

e.g.

```
for key, value := ... {
  for key, value := ... {
    out.Drivers[key].Vehicles[key] = ...
  }
}
```

The above code uses the shadowed local variable to index
both maps which leads to panic ( nil pointer dereference ).

r: @uber/zanzibar-team

### DIFF
--- a/codegen/type_converter.go
+++ b/codegen/type_converter.go
@@ -230,22 +230,29 @@ func (c *TypeConverter) genConverterForList(
 	valueStruct, isStruct := toFieldType.ValueSpec.(*compile.StructSpec)
 	sourceIdentifier := fromIdentifier
 	checkOverride := false
+
+	sourceListID := ""
+	isOverriddenID := ""
+
 	if overriddenIdentifier != "" {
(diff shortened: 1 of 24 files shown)


codegen/type_converter.go
```diff
@@ -81,6 +82,12 @@ func NewTypeConverter(h PackageNameResolver) *TypeConverter {
 	}
 }
 
+func (c *TypeConverter) makeUniqIdentifier(prefix string) string {
```
cl1337: it seems you have an assumption here that `makeUniqIdentifier` always used together for key/val (or items at the same level), shall we either generate a pair of strings as it used, or leave a comment to avoid case like
```
for key1, val2 := range ....
```
?

Raynos: this is used for 5/6 different variable names. We could try to "pretty" them but im happy enough with numerically unique postfixes.

PR #287, opened 2017-12-05
Support for recursive data structurs in zanzibar proxies

oliviazhang0809:
Supporting cases verified - 
```
func convertToNormalClientRequest(in *endpointsBarBar.Bar_Normal_Args) *clientsBarBar.Bar_Normal_Args {
	out := &clientsBarBar.Bar_Normal_Args{}

	convertToNormalRequestClientRequest(in, out)
	convertToNormalRequestRecurClientRequest(in, out)

	return out
}

func convertToNormalRequestRecurClientRequest(in *endpointsBarBar.Bar_Normal_Args, out *clientsBarBar.Bar_Normal_Args) {
	if in.RequestRecur != nil {
		out.RequestRecur = &clientsBarBar.BarRequestRecur{}
		out.RequestRecur.Name = string(in.RequestRecur.Name)
		convertToNormalRecurClientRequest(in, out)
	} else {
		out.RequestRecur = nil
	}
}

func convertToNormalRecurClientRequest(in *endpointsBarBar.Bar_Normal_Args, out *clientsBarBar.Bar_Normal_Args) {
	if in.RequestRecur.Recur != nil {
		out.RequestRecur.Recur = &clientsBarBar.BarRequestRecur{}
		out.RequestRecur.Recur.Name = string(in.RequestRecur.Recur.Name)
		convertToNormalRecurClientRequest(in, out)
	} else {
		out.RequestRecur.Recur = nil
	}
}
```

For
```
struct BarRequestRecur {
    1: required string name
    2: optional BarRequestRecur recur
}

BarResponse normal (
        1: required BarRequest request
        2: optional BarRequestRecur requestRecur
    ) throws (
```


codegen/method.go
```diff
@@ -54,8 +55,9 @@ type HeaderFieldInfo struct {
 
 // MethodSpec specifies all needed parts to generate code for a method in service.
 type MethodSpec struct {
-	Name       string
-	HTTPMethod string
+	Name        string
+	MethodNames []string
```
Raynos: what are method names here ?

codegen/test_data/endpoints/bar_bar_method_toomanyargs.gogen
```diff
@@ -215,17 +225,19 @@ func convertToTooManyArgsClientRequest(in *endpointsBarBar.Bar_TooManyArgs_Args)
 		for key1, value2 := range in.Foo.FooMap {
 			out.Foo.FooMap[key1] = string(value2)
 		}
-		if in.Foo.Message != nil {
-			out.Foo.Message = &clientsFooBaseBase.Message{}
-			out.Foo.Message.Body = string(in.Foo.Message.Body)
-		} else {
-			out.Foo.Message = nil
-		}
+		convertToTooManyArgsMessageClientRequest(in, out)
```
Raynos: this method name contains the local field name instead of the absolute field name.

Should probably be `convertToTooManyArgsFooMessageClientRequest(in, out)`

If its a local message name then `{ foo: { message: 'a' }, bar: { message: 'b' } }` would cause duplicate method name bugs.

Raynos:
This implementation doesn't actually support recursion because we pass in the absolute `in` and `out`

So when you try to copy `{ name: 'a' , recur: { name: 'b', recur: { name: 'c', recur: null } } }` you will not get the correct result but instead get `{ name: 'a': recur { name: 'c', recur: null } }`

examples/example-gateway/idl/endpoints/bar/bar.thrift
```diff
@@ -34,6 +34,11 @@ struct BarResponse {
     6: required binary binaryField
 }
 
+struct BarRequestRecur {
```
Raynos: to verify this you want to add some tests to `test/endpoints/bar/` that make requests with example JSON payloads and observe that the proxy returns the correct JSON inputs/outputs over HTTP requests.

codegen/method.go
oliviazhang0809: this is for dedup the already ran methods - 
```
func (ms *MethodSpec) isMethodprinted(methodName string) bool {
	sort.Strings(ms.MethodNames)
	i := sort.SearchStrings(ms.MethodNames, methodName)
	if i < len(ms.MethodNames) && ms.MethodNames[i] == methodName {
		return true
	}
	return false
}
```

Raynos: methodnames is confusing because its on the MethodSpec.

Maybe we want `ConverterFunctionNames`

codegen/type_converter.go
```diff
@@ -938,9 +938,13 @@ func checkOptionalNil(
 	sort.Strings(keys)
 	for _, id := range keys {
 		if strings.HasPrefix(toIdentifier, id) {
+			/* coverage ignore next line */
```
Raynos: can we add a new test for this case ?

codegen/type_converter_test.go
```diff
@@ -836,38 +806,58 @@ func TestConvertListOfStruct(t *testing.T) {
...
 		nil,
-		nil,
+		nil, false,
 	)
 
 	assert.NoError(t, err)
 	assertPrettyEqual(t, trim(`
 		out.One = make([]*structs.Inner, len(in.One))
 		for index1, value2 := range in.One {
-			if value2 != nil {
```
Raynos: this dissapearrance seems like a bug.

codegen/type_converter_test.go
```diff
@@ -1008,38 +999,25 @@ func TestConvertMapOfStruct(t *testing.T) {
...
-			if value4 != nil {
-				out.Two[key3] = &structs.Inner{}
-				out.Two[key3].Field = (*string)(in.Two[key3].Field)
-			} else {
-				out.Two[key3] = nil
-			}
-		}
-	`), lines)
+	out.One = make(map[string]*structs.Inner, len(in.One))
+	for key1, value2 := range in.One {
```
Raynos: the empty loop body also seems like a bug.

Raynos:
It looks like travis failed because npm made a breaking change :/ the package.json should contain versions for modules we install. Should be a minor fix to package.json and should be reproducible with latest version of npm.

codegen/method.go
```diff
@@ -838,6 +840,74 @@ func (ms *MethodSpec) setDownstream(
...
+) error {
+	typeConverter.IsMethodCall = true
+	index := 0
+	for index < len(typeConverter.HelperFunctionStructs) {
+		helperStruct := typeConverter.HelperFunctionStructs[index]
+		index++
+		fromField := helperStruct.FromField
+		toField := helperStruct.ToField
+
+		methodName := "convertTo" + pascalCase(ms.Name) + pascalCase(fromField.Name) + requestType
```
Raynos: I suspect that `fromField.Name` is a shallow name rather then an absolute name.

codegen/method.go
```diff
@@ -838,6 +840,74 @@ func (ms *MethodSpec) setDownstream(
...
+			continue
+		}
+		ms.ConverterFunctionNames = append(ms.ConverterFunctionNames, methodName)
+
+		// from field type
+		fromFieldType, err := typeConverter.getIdentifierName(fromField.Type)
+		if err != nil {
+			return err
+		}
+		fromFieldType = "*" + fromFieldType
```
Raynos: nit: maybe do a sanity check that `fromField.Type` is a struct before we prefix a `'*'`

codegen/type_converter.go
```diff
@@ -142,6 +162,9 @@ func (c *TypeConverter) genConverterForStruct(
 	fromPrefix string,
 	indent string,
 	fieldMap map[string]FieldMapperEntry,
+	requestType string,
```
Raynos: What is `requestType` ? It seems like it goes on the function name.

Can we store the `requestType` parameter on the `TypeConverter` class itself ?

codegen/type_converter.go
```diff
@@ -188,24 +212,57 @@ func (c *TypeConverter) genConverterForStruct(
 		)
 	}
 
-	c.append(indent, "if ", fromIdentifier, " != nil {")
-	c.append(indent, "\t", toIdentifier, " = &", typeName, "{}")
-
 	subFromFields := fromFieldStruct.Fields
+	fromListOrMap := level+1 == 0
```
Raynos: `level == -1` seems a weird magic number thing.

codegen/type_converter.go
```diff
@@ -68,9 +67,29 @@ type PackageNameResolver interface {
...
-	uninitialized map[string]*fieldStruct
-	fieldCounter  int
+	HelperFunctionStructs []HelperFunctionStruct
+	Helper                PackageNameResolver
+	uninitialized         map[string]*fieldStruct
+	fieldCounter          int
+
+	MethodName   string
+	IsMethodCall bool
+	IsTestCall   bool
```
Raynos: Having an `IsTestCall` for tests seems weird.

codegen/type_converter.go
```diff
@@ -68,9 +67,29 @@ type PackageNameResolver interface {
...
-	Helper        PackageNameResolver
-	uninitialized map[string]*fieldStruct
-	fieldCounter  int
+	HelperFunctionStructs []HelperFunctionStruct
+	Helper                PackageNameResolver
+	uninitialized         map[string]*fieldStruct
+	fieldCounter          int
+
+	MethodName   string
+	IsMethodCall bool
```
Raynos: What is `IsMethodCall` ?

codegen/type_converter.go
```diff
@@ -618,11 +718,15 @@ func (c *TypeConverter) genStructConverter(
 			*compile.StringSpec,
 			*compile.TypedefSpec:
 
+			if c.shouldSkipCall(isRecursiveCall, level) {
```
Raynos: I don't understand why we would want to skip

codegen/type_converter.go
```diff
@@ -604,6 +696,14 @@ func (c *TypeConverter) genStructConverter(
 			fromIdentifier = "in." + fromPrefix + pascalCase(fromField.Name)
 		}
 
+		fromFieldShortName, toFieldShortName := fromIdentifier, toIdentifier
+		if fromField != nil && level > 0 {
```
Raynos: Using level checking seems more fragile

codegen/type_converter.go
```diff
@@ -789,33 +933,6 @@ func (f *FieldAssignment) IsTransform() bool {
 	return f.FromIdentifier[in:] != f.ToIdentifier[out:]
 }
 
-// checkOptionalNil nil-checks fields that may not be initialized along
-// the assign path
-func checkOptionalNil(
```
Raynos: I think this problem still exists. 

This was added to make sure that optional structs dont end up with go default values thus corrupting a JSON wire representation from `{ foo: 'bar' }` to `{ foo: 'bar', nestedThing: { a: '', b: 0 }`

In the case that the thrift is something like

```
struct Outer {
  1: required string foo
  2: optional Inner nestedThing
}

struct InnerThing {
  1: optional string a
  2: optional i32 b
}
```

codegen/test_data/endpoints/bar_bar_method_argwithheaders.gogen
```diff
@@ -181,6 +181,29 @@ func convertArgWithHeadersClientResponse(in *clientsBarBar.BarResponse) *endpoin
...
 	return out
 }
+
+func convertToArgWithHeadersRespClientResponse(in *clientsBarBar.BarRequestRecur) *endpointsBarBar.BarRequestRecur {
+	out := &endpointsBarBar.BarRequestRecur{}
+	if in != nil {
+		out.Name = string(in.Name)
+		out.Recur = convertToArgWithHeadersRecurClientResponse(in.Recur)
+	} else {
+		out = nil
```
cl1337: should response set to nil?

PR #289, opened 2017-12-07
Update service config when adding a new endpoint client group

oliviazhang0809:


backend/repository/endpoint_update.go
```diff
@@ -58,6 +59,11 @@ func (r *Repository) WriteEndpointConfig(
 	if err != nil {
 		return errors.Wrap(err, "failed to write endpoint group configuration")
 	}
+	serviceConfigDir := filepath.Join(r.absPath(endpointCfgDir), "../services/", r.gatewayConfig.ID)
```
Raynos: nitpick: `filepath.Join(x, "..", "services", r.gatewayConfig.ID)`

backend/repository/endpoint_update.go
```diff
@@ -94,6 +100,28 @@ func (r *Repository) validateEndpointCfg(req *EndpointConfig) error {
...
+	}
+	endpoints := fileContent.Dependencies["endpoint"]
+	sort.Strings(endpoints)
+	i := sort.SearchStrings(endpoints, cfg.ID)
+	// not update if client id already exist
+	if i < len(endpoints) && endpoints[i] == cfg.ID {
+		return nil
+	}
+	// update endpoint list with the new client id
+	fileContent.Dependencies["endpoint"] = append(fileContent.Dependencies["endpoint"], cfg.ID)
```
Raynos: can we re-sort before we write back to disk ?

PR #292, opened 2017-12-13
Move method name from method to type converter

oliviazhang0809:
refactor

### DIFF
--- a/codegen/method.go
+++ b/codegen/method.go
@@ -852,12 +852,12 @@ func (ms *MethodSpec) setTypeConverters(
 	structType := compile.FieldGroup(funcSpec.ArgsSpec)
 	downstreamStructType := compile.FieldGroup(downstreamSpec.ArgsSpec)
 
-	typeConverter := NewTypeConverter(h, RequestHelper{
-		RequestSuffix:     "ClientRequest",
-		RequestInputType:  ms.RequestType,
-		RequestOutputType: downstreamMethod.RequestType,
-		ResponseType:      downstreamMethod.ShortRequestType,
-		OutputMethodName:  ms.Name,
+	typeConverter := NewTypeConverter(h, ConvertOptions{
+		FromSuffix:       "ClientRequest",
+		FromInputType:    ms.RequestType,
+		FromOutputType:   downstreamMethod.RequestType,
+		ToType:           downstreamMethod.ShortRequestType,
+		OutputMethodName: ms.Name,
 	})
 
 	err := typeConverter.GenStructConverter(structType, downstreamStructType, reqTransforms, false)
@@ -874,12 +874,12 @@ func (ms *MethodSpec) setTypeConverters(
 		return nil
 	}
 
-	respConverter := NewTypeConverter(h, RequestHelper{
-		RequestSuffix:     "ClientResponse",
-		RequestInputType:  downstreamMethod.ResponseType,
-		RequestOutputType: ms.ResponseType,
-		ResponseType:      ms.ShortResponseType,
-		OutputMethodName:  ms.Name,
+	respConverter := NewTypeConverter(h, ConvertOptions{
+		FromSuffix:       "ClientResponse",
+		FromInputType:    downstreamMethod.ResponseType,
+		FromOutputType:   ms.ResponseType,
+		ToType:           ms.ShortResponseType,
+		OutputMethodName: ms.Name,
 	})
 
 	var respFields, downstreamRespFields []*compile.FieldSpec
--- a/codegen/method.go
+++ b/codegen/method.go
@@ -852,38 +852,47 @@ func (ms *MethodSpec) setTypeConverters(
 	structType := compile.FieldGroup(funcSpec.ArgsSpec)
 	downstreamStructType := compile.FieldGroup(downstreamSpec.ArgsSpec)
 
-	typeConverter := NewTypeConverter(h)
+	typeConverter := NewTypeConverter(h, RequestHelper{
+		UpstreamMethod:   ms,
+		DownstreamMethod: downstreamMethod,
+		RequestSuffix:    "ClientRequest",
+	})
 
-	typeConverter.append(
-		"func convertTo",
-		pascalCase(ms.Name),
-		"ClientRequest(in ", ms.RequestType, ") ", downstreamMethod.RequestType, "{")
+	//typeConverter.append(
+	//	"func convertTo",
+	//	pascalCase(ms.Name),
+	//	"ClientRequest(in ", ms.RequestType, ") ", downstreamMethod.RequestType, "{")
 
-	typeConverter.append("out := &", downstreamMethod.ShortRequestType, "{}\n")
+	//typeConverter.append("out := &", downstreamMethod.ShortRequestType, "{}\n")
 
-	err := typeConverter.GenStructConverter(structType, downstreamStructType, reqTransforms)
+	err := typeConverter.GenStructConverter(structType, downstreamStructType, reqTransforms, false)
 	if err != nil {
 		return err
 	}
-	typeConverter.append("\nreturn out")
-	typeConverter.append("}")
+	//typeConverter.append("\nreturn out")
+	//typeConverter.append("}")
 	ms.ConvertRequestGoStatements = typeConverter.GetLines()
 
 	// TODO: support non-struct return types
 	respType := funcSpec.ResultSpec.ReturnType
-	downstreamRespType := funcSpec.ResultSpec.ReturnType
+	downstreamRespType := downstreamSpec.ResultSpec.ReturnType
 
 	if respType == nil || downstreamRespType == nil {
 		return nil
 	}
 
-	respConverter := NewTypeConverter(h)
+	respConverter := NewTypeConverter(h, RequestHelper{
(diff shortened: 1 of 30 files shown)


codegen/method.go
```diff
@@ -895,18 +895,16 @@ func (ms *MethodSpec) setTypeConverters(
...
 		respFields = respType.(*compile.StructSpec).Fields
 		downstreamRespFields = downstreamRespType.(*compile.StructSpec).Fields
-		respConverter.append("out", " := ", "&", ms.ShortResponseType, "{}\t\n")
-		err = respConverter.GenStructConverter(downstreamRespFields, respFields, respTransforms)
-		if err != nil {
-			return err
-		}
+		isPrimitive = false
+	}
+	err = respConverter.GenStructConverter(downstreamRespFields, respFields, respTransforms, isPrimitive)
```
Raynos: `isPrimitive` boolean seems ugly. Can we just pass in `respType` instead of `respFields` ?

codegen/type_converter.go
```diff
@@ -71,14 +71,26 @@ type TypeConverter struct {
 	Helper        PackageNameResolver
 	uninitialized map[string]*fieldStruct
 	fieldCounter  int
+
+	requestTypeHelper RequestHelper
+}
+
+// RequestHelper is the helper struct for method generation
+type RequestHelper struct {
```
Raynos: Can we rename this to `ConvertOptions` and rename `Request` => `From` and `Response` => `To`

Raynos:
Travis is sad; we can add `package-lock.json` to `.gitignore`

codegen/type_converter.go
```diff
@@ -71,14 +71,26 @@ type TypeConverter struct {
 	Helper        PackageNameResolver
 	uninitialized map[string]*fieldStruct
 	fieldCounter  int
+
+	requestTypeHelper ConvertOptions
```
Raynos: fieldName and typeName dont read cleanly.

codegen/type_converter.go
```diff
@@ -71,14 +71,26 @@ type TypeConverter struct {
...
+	FromSuffix       string
+	FromInputType    string
+	FromOutputType   string
+	ToType           string
+	OutputMethodName string
 }
 
 // NewTypeConverter returns *TypeConverter
-func NewTypeConverter(h PackageNameResolver) *TypeConverter {
+func NewTypeConverter(h PackageNameResolver, requestType ConvertOptions) *TypeConverter {
```
Raynos: paramName and typename dont read cleanly.

codegen/type_converter.go
```diff
@@ -71,14 +71,26 @@ type TypeConverter struct {
 	Helper        PackageNameResolver
```
Raynos: whilst we are here can we fix the `Helper PackageNameResolver` naming.

codegen/type_converter.go
```diff
@@ -71,14 +71,26 @@ type TypeConverter struct {
...
 	uninitialized map[string]*fieldStruct
 	fieldCounter  int
+
+	requestTypeHelper ConvertOptions
+}
+
+// ConvertOptions is the helper struct for method generation
+type ConvertOptions struct {
+	FromSuffix       string
+	FromInputType    string
```
Raynos: `InputTypeName` or something.

codegen/type_converter.go
```diff
@@ -71,14 +71,26 @@ type TypeConverter struct {
...
+	FromOutputType   string
```
Raynos: `ReturnTypeName` or something.

codegen/type_converter.go
```diff
@@ -711,28 +723,61 @@ func (c *TypeConverter) genStructConverter(
...
 ) error {
-	// Add compiled FieldSpecs to the FieldMapperEntry
-	fieldMap = addSpecToMap(fieldMap, fromFields, "")
-	// Check for vlaues not populated recursively by addSpecToMap
-	for k, v := range fieldMap {
-		if fieldMap[k].Field == nil {
+	helper := c.requestTypeHelper
+	var requestInput, requestOutput string
+	requestInput = "(in " + helper.FromInputType + ") " + helper.FromOutputType
+	requestOutput = "out := &" + helper.ToType + "{}\n"
```
Raynos: `ToType` and `FromOutputType` seem duplicate options, can we normalize ?

codegen/type_converter.go
```diff
@@ -711,28 +723,61 @@ func (c *TypeConverter) genStructConverter(
...
+
+	c.append("func convertTo", pascalCase(helper.OutputMethodName), c.requestTypeHelper.FromSuffix,
```
Raynos: maybe `helper.ConverterMethodName` and build `pascalCase(OutputMethodName) + "ClientRequest"` in `method.go` ?

codegen/method.go
```diff
@@ -846,69 +846,45 @@ func (ms *MethodSpec) setTypeConverters(
...
-		pascalCase(ms.Name),
-		"ClientRequest(in ", ms.RequestType, ") ", downstreamMethod.RequestType, "{")
+	structType := &compile.StructSpec{Fields: compile.FieldGroup(funcSpec.ArgsSpec)}
+	downstreamStructType := &compile.StructSpec{Fields: compile.FieldGroup(downstreamSpec.ArgsSpec)}
 
-	typeConverter.append("out := &", downstreamMethod.ShortRequestType, "{}\n")
+	typeConverter := NewTypeConverter(h, ConvertOptions{
+		FromSuffix:       "ClientRequest",
+		FromInputType:    ms.RequestType,
+		FromOutputType:   downstreamMethod.RequestType,
```
Raynos: FromInput and FromOutput seem weird.

`FromType` and `ToType` ?

codegen/method.go
```diff
@@ -846,69 +846,45 @@ func (ms *MethodSpec) setTypeConverters(
...
 		return err
 	}
-	typeConverter.append("\nreturn out")
-	typeConverter.append("}")
 	ms.ConvertRequestGoStatements = typeConverter.GetLines()
 
 	// TODO: support non-struct return types
 	respType := funcSpec.ResultSpec.ReturnType
-	downstreamRespType := funcSpec.ResultSpec.ReturnType
+	downstreamRespType := downstreamSpec.ResultSpec.ReturnType
```
Raynos: ^ this seems like a really nuanced bug. Why did this change, is there a test case for this ?

codegen/type_converter.go
```diff
@@ -711,28 +723,61 @@ func (c *TypeConverter) genStructConverter(
...
+		c.append("out", " := in\t\n")
+		c.append("\nreturn out \t\n}")
+		return nil
+	default:
+		fromFieldStructSpec, ok := fromFieldType.(*compile.StructSpec)
+		if !ok {
 			return errors.Errorf(
-				"Failed to find field ( %s ) for transform.",
-				v.QualifiedName,
+				"Failed to convert field ( %s ).",
```
Raynos: `Failed to convert field ( %s ) to struct`  ?

codegen/type_converter_test.go
```diff
@@ -2254,6 +2442,8 @@ func TestConverterMapMapType(t *testing.T) {
 
 	assert.NoError(t, err)
 	assertPrettyEqual(t, trim(`
+		func convertToFromSuffix(in ToType) ToType{
```
Raynos: these parameters are not correct here :(

codegen/type_converter_test.go
```diff
@@ -68,7 +68,15 @@ func (r *naivePackageNameResolver) TypePackageName(
 }
 
 func newTypeConverter() *codegen.TypeConverter {
-	return codegen.NewTypeConverter(&naivePackageNameResolver{})
+	// construct fake request helper
+	requestHelper := codegen.ConvertOptions{
```
Raynos: we need to pass `fromStruct` and `toStruct` into this method and then use them.

We can also make `FromSuffix` just `ConvertRequest`

codegen/type_converter_test.go
```diff
@@ -229,7 +245,11 @@ func TestConvertBoolsOptionalToRequired(t *testing.T) {
 
 	assert.NoError(t, err)
 	assertPrettyEqual(t, trim(`
+		func convertToFromSuffix(in ToType) ToType{
```
Raynos: I would expect `convertToBar(in *Foo) *Bar {`

Raynos:
Travis now fails due to code coverage on TypeConverter.

```
ERROR: Coverage for statements (99.7%) does not meet global threshold (100%)
```

codegen/method.go
oliviazhang0809: i think the previous was a bug 🤷‍♂️

codegen/type_converter.go
oliviazhang0809: what do you suggest

codegen/type_converter.go
oliviazhang0809: i think making it from/to or request/response is more clear  - 
basically it will be 
```
func OutputMethodName (in FromInputType) FromOutputType {
   out := ToType
}
```

codegen/type_converter.go
oliviazhang0809: what do you suggest

PR #294, opened 2017-12-14
Add peekBody

oliviazhang0809:


runtime/server_http_request_test.go
```diff
@@ -737,3 +738,42 @@ func TestGetQueryValues(t *testing.T) {
...
+		"GET", "/foo", zanzibar.NewRouterEndpoint(
+			bgateway.ActualGateway.Logger,
+			bgateway.ActualGateway.AllHostScope,
+			"foo", "foo",
+			func(
+				ctx context.Context,
+				req *zanzibar.ServerHTTPRequest,
+				res *zanzibar.ServerHTTPResponse,
+			) {
+				req.RawBody = []byte(`{"arg1":{"b1":{"c1":"result"}}}`)
```
Raynos: this is a cheat test. You must call `ReadAll()` before we call `PeekBody()`; `ReadAll()` must become idempotent.

runtime/server_http_request_test.go
```diff
@@ -737,3 +738,42 @@ func TestGetQueryValues(t *testing.T) {
...
+				value, vType, err := req.PeekBody("arg1", "b1", "c1")
+				assert.NoError(t, err, "do not expect error")
+				assert.Equal(t, []byte(`result`), value)
+				assert.Equal(t, vType, jsonparser.String)
+				res.WriteJSONBytes(200, nil, []byte(`{"ok":true}`))
+			},
+		),
+	)
+
+	_, err = gateway.MakeRequest("GET", "/foo?foo=bar", nil, nil)
```
Raynos: should put the body here and make this a `POST`

runtime/server_http_request_test.go
```diff
@@ -737,3 +739,41 @@ func TestGetQueryValues(t *testing.T) {
...
+				assert.Equal(t, vType, jsonparser.String)
+				res.WriteJSONBytes(200, nil, []byte(`{"ok":true}`))
+			},
+		),
+	)
+
+	_, err = gateway.MakeRequest("POST", "/foo?foo=bar", nil, bytes.NewReader([]byte(`{"arg1":{"b1":{"c1":"result"}}}`)))
+	if !assert.NoError(t, err) {
+		return
+	}
```
Raynos: you want to assert you get a 200 back ;) this test gets a 405.

runtime/server_http_request_test.go
```diff
@@ -737,3 +739,45 @@ func TestGetQueryValues(t *testing.T) {
...
+			bgateway.ActualGateway.AllHostScope,
+			"foo", "foo",
+			func(
+				ctx context.Context,
+				req *zanzibar.ServerHTTPRequest,
+				res *zanzibar.ServerHTTPResponse,
+			) {
+				rawBody, success := req.ReadAll()
+				assert.True(t, success)
+				req.RawBody = rawBody
```
Raynos: this breaks encapsulation. `RawBody` should be private not public. `ReadAll()` should assign to it.

Can you refactor `RawBody` => `rawBody`

runtime/server_http_request.go
```diff
@@ -416,7 +433,7 @@ func (req *ServerHTTPRequest) ReadAll() ([]byte, bool) {
 		}
 		return nil, false
 	}
-
+	req.rawBody = rawBody
```
Raynos: it still needs to be idempotent.

`if req.rawBody != nil { ... }` 

We need a test that calling `ReadAll()` twice works.

runtime/server_http_request.go
```diff
@@ -416,7 +438,10 @@ func (req *ServerHTTPRequest) ReadAll() ([]byte, bool) {
 		}
 		return nil, false
 	}
-
+	if req.rawBody != nil {
```
Raynos: we can move this check to line 432 without doing a `ReadAll()` call.

PR #296, opened 2017-12-18
remove unnecessary locks

cl1337:
fix global lock on repository instead of each atomic resources inside it
remove unnecessary lock on runtime repo
atomic load / store on global repo

@uber/zanzibar-team

### DIFF
--- a/backend/repository/endpoint_update_test.go
+++ b/backend/repository/endpoint_update_test.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/uber/zanzibar/codegen"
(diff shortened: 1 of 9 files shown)


backend/repository/endpoint_update_test.go
```diff
@@ -57,8 +58,13 @@ func testUpdateEndpointConfig(t *testing.T, tempDir string, requestFile string)
 	err := readJSONFile(requestFile, req)
 	assert.NoError(t, err, "Failed to unmarshal endpoint config.")
 	r := &Repository{
-		localDir: tempDir,
+		localDir:        tempDir,
+		refreshInterval: time.Hour * 65535,
```
oliviazhang0809: what is this magic number `65535`?

cl1337: no magic # :) a number large enough so we don't refresh repo during a test

PR #299, opened 2017-12-30
add middlewares to module system, codegen and inject clients

jakelarkn:
@uber/zanzibar-team 

key changes
- add middleware as a class type in the module system
- needed config changes:  each middleware has its own middleware-config.json which conforms to the module system json shape
- codegen now does the following
  - generates the module dependency struct for each middleware
  - generates a middleware factory instance which 1) holds a ref to the dependencies struct  2) exposes a factory method for MiddlewareHandle, for the endpoints to pass in their respective options


test/lib/compare.go
```diff
@@ -68,3 +79,20 @@ func DiffStrings(t *testing.T, exp, actual string) {
 	}
 	t.Errorf("Unexpected result: \n%s\n", d)
 }
+
+func deepEqualJSON(b1 []byte, b2 []byte) (bool, error) {
```
cl1337: mm ....  how about
```
	a := `{"a":{"b":1}, "c":{"a":2}}`
	b := `{"c":{"a":2}, "a":{"b":1}}`

	assert.JSONEq(t, a, b)
```

backend/repository/data/gateway_config_expected.json
```diff
@@ -882,13 +882,13 @@
...
-			"schema": "./middlewares/example/example_schema.json",
-			"importPath": "github.com/uber/zanzibar/examples/example-gateway/middlewares/example"
+			"importPath": "github.com/uber/zanzibar/examples/example-gateway/middlewares/example",
+			"schema": "./middlewares/example/example_schema.json"
 		},
 		"example_reader": {
 			"name": "example_reader",
-			"schema": "./middlewares/example_reader/example_reader_schema.json",
-			"importPath": "github.com/uber/zanzibar/examples/example-gateway/middlewares/example_reader"
+			"importPath": "github.com/uber/zanzibar/examples/example-gateway/middlewares/example_reader",
```
Matt-Esch: I'm surprised that this is needed, the module system should provide import paths by convention (and it's part of the basic information provided by all module instances)

codegen/templates/endpoint.tmpl
```diff
@@ -68,7 +68,7 @@ func New{{$handlerName}}(
 		{{ if len $middlewares | ne 0 -}}
 		zanzibar.NewStack([]zanzibar.MiddlewareHandle{
 			{{range $idx, $middleware := $middlewares -}}
-			{{$middleware.Name}}.NewMiddleWare(
+			deps.Middleware.{{$middleware.Name | pascal}}.NewMiddlewareHandle(
```
Matt-Esch: the need to have pascal probably means we're going to run into convention errors at some point

codegen/gateway.go
```diff
@@ -112,24 +112,64 @@ type ClientClassConfig struct {
...
-	Dependencies ClientDependencies     `json:"dependencies"`
+	Dependencies Dependencies           `json:"dependencies"`
 }
 
-// ClientDependencies lists all dependencies of a client.
-type ClientDependencies struct {
+// Dependencies lists all dependencies of a module
+type Dependencies struct {
 	Client []string `json:"client"`
+	//	Service []string `json:"service"`  // example extension
```
oliviazhang0809: Should it be better naming as `ClientDependencies`. Not sure about the usecase of service dependencies.

codegen/gateway.go
```diff
@@ -315,54 +355,28 @@ func newClientSpec(
...
-	Name string `json:"name"`
-	// Go import path for the middleware.
-	Path string `json:"path,omitempty"`
+	Name string
 	// Middleware specific configuration options.
-	Options map[string]interface{} `json:"options,omitempty"`
+	Options map[string]interface{}
 	// Options pretty printed for template initialization
-	PrettyOptions map[string]string `json:"prettyOptions,omitempty"`
+	PrettyOptions map[string]string
```
cl1337: why required now?

backend/repository/data/gateway_config_expected.json
jakelarkn: holdover from the old middleware way-of-doing-things,  will take a stab at rm

codegen/module.go
```diff
@@ -1098,7 +1098,7 @@ type ModuleInstance struct {
 	InstanceName string
 	// Config is a reference to the instance "config" key in the instances json
```
oliviazhang0809: this description needs to be updated

codegen/module.go
```diff
@@ -1139,7 +1139,7 @@ type JSONClassConfig struct {
 	Name string `json:"name"`
 	// The configuration object for this class instance. This depends on the
```
oliviazhang0809: update description as well

codegen/module_system.go
```diff
@@ -1054,6 +1076,80 @@ func (generator *GatewayServiceGenerator) Generate(
...
+	//meta := &MiddlewareMeta{
+	//	Instance: instance,
+	//	Go: instance.Config.(*MiddlewareConfigTmp),
+	//}
+	bytes, err := g.templates.ExecTemplate("middleware.tmpl", instance, g.packageHelper)
+	if err != nil {
+		return err
+	}
+
+	//targetMiddlewareFile := filepath.Join(
```
oliviazhang0809: remove commented out code or add TODO?

codegen/module_system.go
```diff
@@ -1054,6 +1076,80 @@ func (generator *GatewayServiceGenerator) Generate(
...
+}
+
+// Generate returns the endpoint build result, which contains a file per
+// endpoint handler and a list of handler specs
+func (g *MiddlewareGenerator) Generate(
+	instance *ModuleInstance,
+) (*BuildResult, error) {
+	ret := map[string][]byte{}
+
+	//	clientSpecs := readClientDependencySpecs(instance)
```
cl1337: remove comment?

codegen/module_system.go
```diff
@@ -1054,6 +1076,80 @@ func (generator *GatewayServiceGenerator) Generate(
...
+	}
+	return &BuildResult{
+		Files: ret,
+		Spec:  nil,
+	}, nil
+}
+
+func (g *MiddlewareGenerator) generateMiddlewareFile(instance *ModuleInstance, out map[string][]byte) error {
+
+	//meta := &MiddlewareMeta{
```
cl1337: is these comments needed?

runtime/middlewares_test.go
```diff
@@ -36,8 +36,9 @@ import (
 
 // Ensures that a middleware stack can correctly return all of its handlers.
 func TestHandlers(t *testing.T) {
-	ex := example.NewMiddleWare(
+	ex := example.NewMiddleware(
 		nil, // *zanzibar.Gateway
+		nil,
```
oliviazhang0809: can we add test with dep.client not null?

oliviazhang0809:
nit: pls `make generate` before land to make copyright 2018. :)

test/lib/compare.go
jakelarkn: nice

codegen/gateway.go
jakelarkn: naming just mirrors the module system JSON config

service interfaces abstract (one or more clients + minimal business logic + feature toggles).  service dependencies are used somewhat extensively in RTAPI and are a part of the design of the module system in zanzibar.   they don't really make a lot of sense for the edge proxy use-cases   but imagine they will be important for fulfillment and other gateways as code base matures

codegen/gateway.go
jakelarkn: I believe this object is no longer serialized with my changes in the repository package

codegen/module_system.go
jakelarkn: had uncommited changes to remove them, doh!

jakelarkn:
going to punt on the import changes for later PR.   hit a weird bug with identifying import path in codegen conflicting with how jenkins test run script utilizes a symlinked project directory,  resulting in a `make check-generate` that passes locally but fails on jenkins.  already lost a day on it ... this PR unblocks a ton of middleware work in edge.

PR #301, opened 2018-01-03
fix lint for naming issue

oliviazhang0809:
fix the lint for Uuid should be UUID, Id should be ID, etc


codegen/gateway.go
```diff
@@ -41,6 +41,14 @@ func getDirName() string {
 	return zanzibar.GetDirnameFromRuntimeCaller(file)
 }
 
+var lintCorrectionMap = map[string]string{
```
cl1337: is it possible to have import some golint rules here? I'm a bit concern about us maintaining a key word list, but :man_shrugging:

codegen/gateway.go
```diff
@@ -184,6 +192,7 @@ func NewTChannelClientSpec(
 	cspec.ExposedMethods = make(map[string]string, len(exposedMethods))
 	reversed := make(map[string]string, len(exposedMethods))
 	for key, val := range exposedMethods {
+		key = correctMethodNaming(key)
```
ChuntaoLu: We should not do this, this creates confusion and inconsistency between config and generated code, i.e., I configured my method name to be `TestUrl`(which is wrong) but the codegen generates a method `TestURL`. The user should provide `TestURL` in the configuration at first place.

examples/example-gateway/clients/bar/client-config.json
```diff
@@ -18,6 +18,7 @@
 			"ArgWithManyQueryParams": "Bar::argWithManyQueryParams",
 			"ArgWithParams": "Bar::argWithParams",
 			"Hello": "Bar::helloWorld",
+			"TestUrl": "Bar::testUrl",
```
ChuntaoLu: The user should just provide `TestURL` to make linter pass in the first place.

oliviazhang0809: @ChuntaoLu it's more like failing our server side proxy method names instead of affecting clients. So my suggestion is to fix it in our side instead of pushing it down to clients. 
Reusing the `commonInitialisms` to simplify the process.

PR #303, opened 2018-01-03
fix lint for initialism

oliviazhang0809:
Uuid -> UUID when creating diff and codegen


backend/repository/client_update.go
```diff
@@ -46,6 +46,16 @@ type configField struct {
 	ExposedMethods map[string]string `json:"exposedMethods,omitempty"`
 }
 
+func correctMethodNaming(key string) string {
+	for k := range codegen.CommonInitialisms {
+		initial := string(k[0]) + strings.ToLower(k[1:])
+		if strings.Contains(key, string(k[0])+strings.ToLower(k[1:])) {
```
ChuntaoLu: isn't `string(k[0])+strings.ToLower(k[1:])` same as `initial`?

backend/repository/client_update.go
```diff
@@ -46,6 +46,16 @@ type configField struct {
...
+			key = strings.Replace(key, initial, k, 1)
```
ChuntaoLu: This means `TestUrlAndUrl` will become `TestURLAndUrl`, not `TestURLAndURL. I think you want replace all occurrences.

PR #304, opened 2018-01-03
Add default headers from the client config

rpatali:
Client config can specific a dict of default headers that should be
added to every request from that client.

### DIFF
--- a/examples/example-gateway/build/clients/bar/bar.go
+++ b/examples/example-gateway/build/clients/bar/bar.go
@@ -208,6 +208,10 @@ func NewClient(
 	port := deps.Default.Config.MustGetInt("clients.bar.port")
 	baseURL := fmt.Sprintf("http://%s:%d", ip, port)
 	timeout := time.Duration(deps.Default.Config.MustGetInt("clients.bar.timeout")) * time.Millisecond
+	defaultHeaders := make(map[string]string)
+	if deps.Default.Config.ContainsKey("clients.bar.defaultHeaders") {
+		deps.Default.Config.MustGetStruct("clients.bar.defaultHeaders", &defaultHeaders)
+	}
 
 	return &barClient{
 		clientID: "bar",
@@ -247,6 +251,7 @@ func NewClient(
 				"EchoTypedef",
 			},
 			baseURL,
+			defaultHeaders,
 			timeout,
 		),
 	}
--- a/runtime/client_http_request_test.go
+++ b/runtime/client_http_request_test.go
@@ -62,6 +62,7 @@ func TestMakingClientWriteJSONWithBadJSON(t *testing.T) {
 		"clientID",
 		[]string{"DoStuff"},
 		"/",
+		map[string]string{},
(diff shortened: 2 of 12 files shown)


runtime/client_http_request_test.go
```diff
@@ -62,6 +62,7 @@ func TestMakingClientWriteJSONWithBadJSON(t *testing.T) {
 		"clientID",
 		[]string{"DoStuff"},
 		"/",
+		map[string]string{},
```
oliviazhang0809: can we have test when header is not empty and verify it's being correctly passed in and set?

runtime/client_http_response_test.go
```diff
@@ -72,6 +72,7 @@ func TestReadAndUnmarshalNonStructBody(t *testing.T) {
 		"bar",
 		[]string{"echo"},
 		baseURL,
+		map[string]string{},
```
oliviazhang0809: same for here

runtime/http_client.go
```diff
@@ -33,6 +33,7 @@ import (
 type HTTPClient struct {
 	Client  *http.Client
 	BaseURL string
+	Headers map[string]string
```
ChuntaoLu: Let's call it `DefaultHeaders` to be consistent with the config filed name.

examples/example-gateway/build/clients/bar/bar.go
```diff
@@ -208,6 +208,10 @@ func NewClient(
 	port := deps.Default.Config.MustGetInt("clients.bar.port")
 	baseURL := fmt.Sprintf("http://%s:%d", ip, port)
 	timeout := time.Duration(deps.Default.Config.MustGetInt("clients.bar.timeout")) * time.Millisecond
+	defaultHeaders := make(map[string]string)
+	if deps.Default.Config.ContainsKey("clients.bar.defaultHeaders") {
```
cl1337: why not always generate `defaultHeaders` empty map in config by default? :)

ChuntaoLu: config file is not necessarily generated. From zanzibar perspective, we assume config files will be read/written by humans, in which sense less clutter is good.

cl1337: make sense, for edge-gateway tho is always gen-ed from ui

runtime/client_http_request_test.go
rpatali: just fyi, i added the test line #134
this test and the next are testing bad input so the request is never sent to the server.

runtime/client_http_request_test.go
```diff
@@ -110,7 +112,7 @@ func TestMakingClientWriteJSONWithBadHTTPMethod(t *testing.T) {
 	assert.Len(t, logs["Could not create outbound request"], 1)
 }
 
-func TestMakingClientCalLWithHeaders(t *testing.T) {
```
cl1337: lol

PR #306, opened 2018-01-04
fix to pass over directories in middlewares that are not middleware modules

jakelarkn:


codegen/gateway.go
```diff
@@ -947,7 +947,10 @@ func parseMiddlewareConfig(
 		}
 		instanceConfig := filepath.Join(fullMiddlewareDir, file.Name(), "middleware-config.json")
 		bytes, err := ioutil.ReadFile(instanceConfig)
-		if err != nil {
+		if os.IsNotExist(err) {
```
cl1337: mm ... if user pass wrong middleware dir should we still error out? Or this is some other cases?

jakelarkn: in edge there is a lib package

PR #308, opened 2018-01-04
Fix initialism for test_server

oliviazhang0809:
Still see errors like in test_server.tmp for tchannel service
```
 type UserServicegetFullPictureUrlFunc should be UserServicegetFullPictureURLFunc
build/clients/populous-client/populous-client_test_server.go:3100:6: func NewUserServiceGetFullPictureUrlHandler should be 
```


codegen/method.go
```diff
@@ -162,6 +163,7 @@ func NewMethod(
 	var err error
 	var ok bool
 	method.Name = funcSpec.MethodName()
+	method.AdjustedName = CorrectMethodNaming(method.Name)
```
cl1337: mm ... what's the purpose for `adjustedName`?

oliviazhang0809: @cl1337 in `tchannel_client_test_server.tmpl` we was using method name to generate test server function type and struct type, which breaks lint since method name can be `testUuid` etc. Since method name is used in a lot of other places, I create the `AdjustedName` for holding initialism adjusted method name.

codegen/casing.go
```diff
@@ -71,6 +71,17 @@ var CommonInitialisms = map[string]bool{
 }
 var camelingRegex = regexp.MustCompile("[0-9A-Za-z]+")
 
+// CorrectMethodNaming correct the naming for initialisms
+func CorrectMethodNaming(key string) string {
```
ChuntaoLu: This function works on a string whose first letter is in upper case and rest are in lower case, imagine a method name that is `myService::uuidTest`, it won't work anymore. The target `key` is a string whose upper case is in the common intialisms.  There is already the `ensureGolangAncronymCasing` method that you can leverage to come up a generic function that works on an arbitrary string. It's also not limited to method naming, so the function name should be generic.

cl1337: +1 on more case check required but scope could go wild

for example `API` check too general might mess up with name like
`MyServiceApicultures` -> `MyServiceAPIcultures` ?

if this can be extract to easily extended rules instead of hardcode check it would be better, we'll just be patching the cases we see

on the other side, this was the concern I had whether we should fix this or just bypass lint

codegen/template.go
```diff
@@ -63,6 +63,7 @@ var defaultFuncMap = tmpl.FuncMap{
 	"jsonMarshal":   jsonMarshal,
 	"isPointerType": isPointerType,
 	"unref":         unref,
+	"lintAcronym":   CorrectMethodNaming,
```
cl1337: nit: would be nice to have same name

codegen/casing.go
```diff
@@ -71,6 +71,18 @@ var CommonInitialisms = map[string]bool{
...
+	for k := range CommonInitialisms {
+		key = pascalCase(key)
```
ChuntaoLu: This line should be moved outside the for loop.

PR #309, opened 2018-01-05
logged field is causing nil reference

rpatali:
T1459853

### DIFF
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -222,12 +222,14 @@ func (c *tchannelOutboundCall) logFields() []zapcore.Field {
 	var hostPort string
 	if c.call != nil {
 		hostPort = c.call.RemotePeer().HostPort
+	} else {
+		hostPort = "unknown"
 	}
 	fields := []zapcore.Field{
+		zap.String("remoteAddr", hostPort),
 		zap.Time("timestamp-started", c.startTime),
 		zap.Time("timestamp-finished", c.finishTime),
 	}
-	fields = append(fields, zap.String("remoteAddr", hostPort))
 
 	for k, v := range c.reqHeaders {
 		fields = append(fields, zap.String("Request-Header-"+k, v))
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -222,14 +222,12 @@ func (c *tchannelOutboundCall) logFields() []zapcore.Field {
 	var hostPort string
 	if c.call != nil {
 		hostPort = c.call.RemotePeer().HostPort
-	} else {
-		hostPort = ""
 	}
 	fields := []zapcore.Field{
-		zap.String("remoteAddr", hostPort),
 		zap.Time("timestamp-started", c.startTime),
 		zap.Time("timestamp-finished", c.finishTime),
 	}
+	fields = append(fields, zap.String("remoteAddr", hostPort))
 
 	for k, v := range c.reqHeaders {
 		fields = append(fields, zap.String("Request-Header-"+k, v))
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -219,8 +219,14 @@ func (c *tchannelOutboundCall) finish(err error) {
 }
 
 func (c *tchannelOutboundCall) logFields() []zapcore.Field {
+	var hostPort string
+	if c.call != nil {
+		hostPort = c.call.RemotePeer().HostPort
+	} else {
+		hostPort = ""
+	}
 	fields := []zapcore.Field{
-		zap.String("remoteAddr", c.call.RemotePeer().HostPort),
+		zap.String("remoteAddr", hostPort),
 		zap.Time("timestamp-started", c.startTime),
 		zap.Time("timestamp-finished", c.finishTime),
 	}
--- a/runtime/tchannel_client_test.go
+++ b/runtime/tchannel_client_test.go
@@ -1,0 +1,56 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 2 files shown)


runtime/tchannel_client.go
```diff
@@ -219,8 +219,14 @@ func (c *tchannelOutboundCall) finish(err error) {
 }
 
 func (c *tchannelOutboundCall) logFields() []zapcore.Field {
+	var hostPort string
```
rpatali: @ChuntaoLu can't find an obvious way to write a test for this change. any suggestions?

cl1337: or `fields = append(fields, hostPort` if call is not nil, less noise in log I think? cc @ravirajj

rpatali: i that case its a broader question about convention. Do we want to always log an empty thing for that type in case of nil or do we always want to skip nil fields in the log altogether? Either way is fine but better to be consistent one way or another.

runtime/tchannel_client.go
```diff
@@ -219,8 +219,14 @@ func (c *tchannelOutboundCall) finish(err error) {
...
+	if c.call != nil {
+		hostPort = c.call.RemotePeer().HostPort
+	} else {
+		hostPort = ""
+	}
 	fields := []zapcore.Field{
-		zap.String("remoteAddr", c.call.RemotePeer().HostPort),
+		zap.String("remoteAddr", hostPort),
```
oliviazhang0809: do we still need to log `hostPort` if it's empty string?
I'd suggest just create a list of fields and append `hostPort` if it's not null

runtime/tchannel_client.go
```diff
@@ -219,8 +219,14 @@ func (c *tchannelOutboundCall) finish(err error) {
 }
 
 func (c *tchannelOutboundCall) logFields() []zapcore.Field {
+	var hostPort string
+	if c.call != nil {
```
ChuntaoLu: The problem here is not `c.call` is `nil`, it's the `c.call.RemotePeer()` being nil, in that case we can set host to `unknown` instead of empty string, because empty string might cause the key to not show in the logs at all.

runtime/tchannel_client.go
ChuntaoLu: A contrived test case would be to config the client with a hostport that no backend server listens at, but for the sake of this function it's okay to not go that far.

rpatali: discussed offline, planning to not log on nil.

runtime/tchannel_client.go
rpatali: we discussed this offline, c.call is indeed nil.

runtime/tchannel_client.go
rpatali: Added a unit test instead for the case when call is nil although i couldn't find a way to trivially write unit test for the positive case.

runtime/tchannel_client.go
```diff
@@ -219,8 +219,14 @@ func (c *tchannelOutboundCall) finish(err error) {
...
+		hostPort = c.call.RemotePeer().HostPort
+	} else {
+		hostPort = "unknown"
+	}
```
cl1337: nit:
```
hostPort := "unknown"
if c.call != nil {
    hostPort = c.call.RemotePeer().HostPort
}
```
prefer no `else` whenever possible

runtime/tchannel_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+
+	"testing"
+	"time"
+)
+
+func TestNilCallReferenceForLogger(t *testing.T) {
+
```
cl1337: remove empty line

runtime/tchannel_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+	headers := map[string]string{
+		"header-key": "header-value",
+	}
+	outboundCall := &tchannelOutboundCall{
+		client:        nil,
```
cl1337: you can skip all nil value in initialization
```
&tchannelOutboundCall{
    methodName:    "GET",
    serviceMethod: "Test",
}
```

runtime/tchannel_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+		call:          nil,
+		methodName:    "GET",
+		serviceMethod: "Test",
+		success:       false,
```
cl1337: same, boolean value default to false, no need in init

runtime/tchannel_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+		startTime:     time.Now(),
```
cl1337: avoid use time.Now() in tests, hardcode a time here, changing variable in test is a perfect time bomb

runtime/tchannel_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+		finishTime:    time.Now(),
+		reqHeaders:    headers,
+		resHeaders:    headers,
+		logger:        nil,
+		metrics:       nil,
+	}
+	fields := outboundCall.logFields()
+
+	// one field for each of the the:
```
cl1337: dup the

also principle for comments: 
comment on why not what

runtime/tchannel_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zanzibar
+
+import (
+	"github.com/stretchr/testify/assert"
+
+	"testing"
```
cl1337: please do follow the style in this doc: https://github.com/golang/go/wiki/CodeReviewComments#imports

runtime/tchannel_client_test.go
rpatali: completely agree with the principle.
i would forget in a day why the fields length is 5 on the next line but i'm happy to remove it if you think it helps.

runtime/tchannel_client_test.go
rpatali: makes sense.

runtime/tchannel_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func TestNilCallReferenceForLogger(t *testing.T) {
+
+	headers := map[string]string{
+		"header-key": "header-value",
+	}
+	outboundCall := &tchannelOutboundCall{
+		client:        nil,
+		call:          nil,
+		methodName:    "GET",
```
ChuntaoLu: trivial but the method name should be "Get" per tchannel client method convention.

runtime/tchannel_client.go
rpatali: we discussed this and decided to leave it as it is. I prefer no `else` when possible but in this case it would mean we always assign an incorrect value of "unknown" in case of `c.call != nil` which feels weird.

cl1337:
lgtm

PR #311, opened 2018-01-09
bump tchannel version

cl1337:
conflict with some pkg in projects


test/endpoints/tchannel/baz/baz_simpleservice_method_call_test.go
```diff
@@ -98,10 +98,9 @@ func TestCallTChannelSuccessfulRequestOKResponse(t *testing.T) {
 	assert.True(t, success)
 
 	allLogs := gateway.AllLogs()
-	assert.Equal(t, 5, len(allLogs))
+	assert.Equal(t, 4, len(allLogs))
```
ChuntaoLu: Was the number of logs change caused by tchannel upgrade?

cl1337: yes, new log use `create active connection` and has len of 2 instead of one log for outgoing and one for incoming

ChuntaoLu: Yeah, that's thing. We really shouldn't depend our tests on log entries from libraries, which could change any time.

cl1337: agreed, I saw this added multiple places thought it was a deprecated way to test tchannel behavior difference between versions T_T

PR #314, opened 2018-01-16
Allow services in same thrift to have same method name

jyang0:
- in generated code, added service name prefix to method names to tell them apart
- added a new example thrift with 2 services having same method name
- updated existing test data


backend/repository/data/handler/test_cases.json
```diff
@@ -2700,6 +2764,18 @@
 							"thriftFile": "clients/googlenow/googlenow.thrift",
 							"type": "http"
 						},
+						"multi": {
+							"exposedMethods": {
+								"HelloA": "ServiceABack::hello",
```
cl1337: we'll have to fix how `HelloA` is generated in here: https://github.com/uber/zanzibar/blob/master/backend/repository/client_update.go#L89

I believe now it still use method name as key, that will be overriden

codegen/method.go
```diff
@@ -881,7 +881,7 @@ func (ms *MethodSpec) setTypeConverters(
 
 	respConverter.append(
 		"func convert",
-		pascalCase(ms.Name),
+		ms.DownstreamService, pascalCase(ms.Name),
```
ChuntaoLu: While conventionally service name is in pascal case, but it's not enforced. So let's apply `pascalCase` on the service name as well.

codegen/module_system.go
```diff
@@ -839,7 +839,7 @@ func (g *EndpointGenerator) generateEndpointFile(
 
 	var workflowName string
 	if method.Downstream != nil {
-		workflowName = strings.Title(method.Name) + "Endpoint"
+		workflowName = thriftServiceName + strings.Title(method.Name) + "Endpoint"
```
ChuntaoLu: Similarly here, let's `string.Title` the service name.

codegen/method.go
jyang0: Got it!

PR #317, opened 2018-01-20
Add support for recursive structs in thrift file

jyang0:
Allow thrift file to define a struct with an optional field for itself, or multiple structs with optional fields forming a cycle

- Prevent infinite recursion in type conversion by enclosing repeating code inside function literals, and then calling the function literal
- Preserves existing behavior, including with fieldMap and optional field nil-checking
- Add new tests, update existing tests
- Fix bug in method.go


oliviazhang0809:
Can you add a new method in bar that contains recursive struct in request or response and make sure the generated result is correct?

codegen/method.go
```diff
@@ -871,7 +871,8 @@ func (ms *MethodSpec) setTypeConverters(
 
 	// TODO: support non-struct return types
 	respType := funcSpec.ResultSpec.ReturnType
-	downstreamRespType := funcSpec.ResultSpec.ReturnType
+
+	downstreamRespType := downstreamMethod.CompiledThriftSpec.ResultSpec.ReturnType
```
oliviazhang0809: good catch 👍

codegen/type_converter.go
```diff
@@ -142,13 +144,14 @@ func (c *TypeConverter) genConverterForStruct(
...
-
-	typeName, err := c.getIdentifierName(toFieldType)
-	if err != nil {
-		return err
+	toIdentifier := "out"
+	if keyPrefix != "" {
+		toIdentifier += "." + keyPrefix
 	}
+
+	typeName, _ := c.getIdentifierName(toFieldType)
```
oliviazhang0809: probably shouldn't skip error check here. :)

codegen/type_converter.go
```diff
@@ -297,9 +320,10 @@ func (c *TypeConverter) genConverterForList(
 			fromFieldType.ValueSpec,
 			valID,
 			keyPrefix+pascalCase(toField.Name)+"["+indexID+"]",
-			strings.TrimPrefix(fromIdentifier, "in.")+"["+indexID+"]",
+			strings.TrimPrefix(strings.TrimPrefix(fromIdentifier, "in."), "inOriginal.")+"["+indexID+"]",
```
oliviazhang0809: nit: `inOriginal` and `outOriginal` are constant declared upfront.

codegen/type_converter.go
```diff
@@ -638,25 +671,78 @@ func (c *TypeConverter) genStructConverter(
...
-				indent,
-				fieldMap,
-			)
+
+			toTypeName, err := c.getIdentifierName(toFieldType)
 			if err != nil {
 				return err
 			}
+
+			if converterMethodName, ok := c.convStructMap[toFieldType.Name]; ok {
```
oliviazhang0809: this struct case is getting big, can we put it in a helper function for better readability?

codegen/type_converter.go
```diff
@@ -638,25 +671,78 @@ func (c *TypeConverter) genStructConverter(
...
+				// the converter for this struct already exists, so just use it
+				c.append(indent, "out.", keyPrefix+pascalCase(toField.Name), " = ", converterMethodName, "(", fromIdentifier, ")")
+			} else if fromTypeName != "" {
+				// generate a converter for this scope
```
cl1337: what's the purpose for scope use here

codegen/type_converter.go
```diff
@@ -638,25 +671,78 @@ func (c *TypeConverter) genStructConverter(
...
+				if err != nil {
+					return err
+				}
+
+				c.append(indent+"\t\t", "return")
+				c.append(indent+"\t", "}") // end of function definition
+
+				c.append(indent+"\t", "out.", keyPrefix+pascalCase(toField.Name), " = ", converterMethodName, "(", fromIdentifier, ")")
+
+				delete(c.convStructMap, toFieldType.Name)
```
cl1337: why the delete

codegen/type_converter.go
```diff
@@ -638,25 +671,78 @@ func (c *TypeConverter) genStructConverter(
...
+
+			toTypeName, err := c.getIdentifierName(toFieldType)
 			if err != nil {
 				return err
 			}
+
+			if converterMethodName, ok := c.convStructMap[toFieldType.Name]; ok {
+				// the converter for this struct already exists, so just use it
+				c.append(indent, "out.", keyPrefix+pascalCase(toField.Name), " = ", converterMethodName, "(", fromIdentifier, ")")
+			} else if fromTypeName != "" {
```
cl1337: this if else block can be more intuitive 
```
case <already exist nested struct>
case <no nested struct, gen new>
case <not nested struct?>
```
???
seems hard to tell what's the different case here

codegen/type_converter.go
```diff
@@ -779,14 +870,14 @@ type FieldAssignment struct {
 
 // IsTransform returns true if field assignment is not proxy
 func (f *FieldAssignment) IsTransform() bool {
-	var (
```
cl1337: I think it's better we do `trimName` that trims all the wired prefix and do 
`trimed(inName) != trimed(outName)`

this logic here is hard to read now

codegen/type_converter.go
```diff
@@ -804,9 +896,13 @@ func checkOptionalNil(
 	for k := range uninitialized {
 		keys = append(keys, k)
 	}
+
```
cl1337: I think we have introduced to many piece of information here to simply identify some name in different layers, it won't be too long before type_converter is completely unmaintainable

codegen/type_converter_test.go
```diff
@@ -1523,22 +1593,38 @@ func TestConverterMapStructWithSubFields2(t *testing.T) {
 
 	assert.NoError(t, err)
 	assertPrettyEqual(t, trim(`
-		if in.Three != nil {
-			out.Three = &structs.NestedBar{}
-			out.Three.One = string(in.Three.One)
-			if in.Four != nil && in.Four.Two != nil {
-				out.Three.One = *(in.Four.Two)
+		inOriginal := in; _ = inOriginal
```
cl1337: this looks wired, what does it do?

codegen/type_converter_test.go
```diff
@@ -2530,3 +2707,553 @@ func TestCoverIsTransform(t *testing.T) {
...
+					}
+				} else {
+					out = nil
+				}
+				return
+			}
+			out.Recur = convertStringFooBHelper(in.Recur)
+		}
+		{
+			var convertBarHelper func(in *structs.Foo) (out *structs.Bar)
```
cl1337: consider such code on prod with 30K qps, we definitely want to know the performance, have we benchmarked this?

also I feel if we ever have any issues around generated code during runtime for type_converter we won't be able to debug it easily

codegen/type_converter_test.go
```diff
@@ -2530,3 +2707,553 @@ func TestCoverIsTransform(t *testing.T) {
...
+ 			2: optional NestedFooB six
+ 		}`,
+		nil,
+		fieldMap,
+	)
+
+	assert.NoError(t, err)
+	assertPrettyEqual(t, trim(`
+		{
+			var convertNestedFooBHelper func(in *structs.NestedFooA) (out *structs.NestedFooB)
```
cl1337: :sadpanda:

codegen/type_converter.go
jyang0: Right :) Will try to make the error testable / try to get value from a previous call

codegen/type_converter.go
jyang0: Yeah good idea!

codegen/type_converter.go
jyang0: Yeah, that's the right understanding, I'll try to clean it up

codegen/type_converter.go
jyang0: Originally it was to avoid name conflicts, but in hindsight that probably won't be an issue, I can try removing them

codegen/type_converter.go
jyang0: The delete is necessary since a converter method might have been generated somewhere deep, which wouldn't be accessible if the same type conversion was needed again somewhere less deep

codegen/type_converter_test.go
jyang0: In the current implementation this is kinda needed to support fieldMap overrides: within a function literal, only the "local" ```in``` and ```out``` are visible, but the fieldMap override could be taking in values from some other part of the original ```in```

codegen/type_converter_test.go
jyang0: In theory, the performance difference should only come from the extra call to the function literal

codegen/type_converter.go
jyang0: More importantly, the generated method is not safe for re-use because that conversion and the current one might not have the same code due to fieldMap overrides

codegen/type_converter.go
jyang0: Okay, I'll try to make it more obvious it's doing that, but there's still the prefix check too 😛

codegen/type_converter_test.go
jyang0: I'll change it so that when the input & output are both non-recursive, the generated code will stay exactly like before

codegen/type_converter.go
jyang0: It seems there's no way to get an error here, since c.getIdentifierName (or equivalent) is now called upstream, and it would error before it reaches here

codegen/type_converter_test.go
ChuntaoLu: > I'll change it so that when the input & output are both non-recursive, the generated code will stay exactly like before

I was about to suggest the same thing, recursive structs will be rare corner cases. Having a function for a non-recursive struct is not necessary.

codegen/type_converter.go
oliviazhang0809: hmm seems the original code covered this case. Why difference?

codegen/type_converter.go
oliviazhang0809: up

codegen/type_converter.go
```diff
@@ -935,7 +1048,101 @@ func (c *TypeConverter) genConverterForPrimitiveOrTypedef(
...
+	c.append(indent, "}") // end of function definition
+
+	c.append(indent, "out.", toSubIdentifier, " = ", converterMethodName, "(", fromIdentifier, ")")
+
+	delete(c.convStructMap, toFieldType.Name)
+
+	return nil
+}
+
+func isRecursiveStruct(spec compile.TypeSpec, seenSoFar map[string]bool) bool {
```
oliviazhang0809: i can see why this function here but delete a map at realtime is not the best practise :sadpanda. 
In general this is not very good to understand. i'm not sure if there's a way to refactor.
If you have to do this, be sure to add more comments otherwise later on it will be a pita for later maintenance.

codegen/type_converter.go
```diff
@@ -638,25 +689,61 @@ func (c *TypeConverter) genStructConverter(
...
+					fromIdentifier,
+					keyPrefix+pascalCase(toField.Name),
+					stFromPrefix,
+					indent,
+					fieldMap,
+					prevKeyPrefixes,
+				)
+				if err != nil {
+					return err
+				}
```
oliviazhang0809: this two if err != nil can be moved out of if else loop

codegen/type_converter.go
jyang0: I added a ```typeName, err := c.getIdentifierName(toFieldType)``` in genStructConverter that wasn't there before, to check if that struct converter has already been generated / generate one if necessary

codegen/type_converter.go
jyang0: True

codegen/type_converter.go
jyang0: What alternative structure should I use? Sorry, I'm not sure I understand why using delete on a map is bad, can you please explain?

Okay, I can try to add better comments

codegen/type_converter.go
jyang0: I have to declare them as constants upfront, otherwise later inside some function literal I won't be able to access the original ```in``` and ```out```. Normally I wouldn't need to but fieldMap overrides are written in terms of the original input (hence inOriginal). outOriginal is needed for nil-checking when a fieldMap has an assignment to an optional to-field with no corresponding from-field

codegen/type_converter.go
```diff
@@ -727,7 +814,18 @@ func (c *TypeConverter) GenStructConverter(
 		}
 	}
 
-	err := c.genStructConverter("", "", "", fromFields, toFields, fieldMap)
+	if c.isRecursiveStruct(toFields) || c.isRecursiveStruct(fromFields) {
+		c.useRecurGen = true
+	} else {
+		c.useRecurGen = false
+	}
```
cl1337: `c.useRecurGen = c.isRecursiveStruct(toFields) || c.isRecursiveStruct(fromFields)` ?

codegen/type_converter.go
```diff
@@ -79,6 +81,8 @@ func NewTypeConverter(h PackageNameResolver) *TypeConverter {
 		LineBuilder:   LineBuilder{},
 		Helper:        h,
 		uninitialized: make(map[string]*fieldStruct),
+		convStructMap: make(map[string]string),
+		useRecurGen:   false,
```
cl1337: don't need to init default value for primary types, int=0, bool=false ... etc

codegen/type_converter.go
cl1337: maybe not nit, I think variable scope is a bit out of control as @oliviazhang0809 suggested, if you need to access some variables set miles away from where it used, it's usually a good sign to refactor :)

I'll leave this one to @oliviazhang0809

codegen/type_converter.go
```diff
@@ -458,9 +489,10 @@ func (c *TypeConverter) genConverterForMap(
 			fromFieldType.ValueSpec,
 			valID,
 			keyPrefix+pascalCase(toField.Name)+"["+keyID+"]",
-			strings.TrimPrefix(fromIdentifier, "in.")+"["+keyID+"]",
+			strings.TrimPrefix(strings.TrimPrefix(fromIdentifier, "in."), "inOriginal.")+"["+keyID+"]",
```
cl1337: mm ... you can do better here :) `TrimPrefix` twice seems too much implication, what's the unambiguous way of saying: 

I want to get the real arg name without all the prefixes I might have added

codegen/type_converter.go
```diff
@@ -604,6 +650,10 @@ func (c *TypeConverter) genStructConverter(
 			fromIdentifier = "in." + fromPrefix + pascalCase(fromField.Name)
 		}
 
+		if prevKeyPrefixes == nil {
+			prevKeyPrefixes = []string{}
```
cl1337: saw you sometimes use `[]string{}`, sometimes `make`, what's the difference? would be nice to have some consistency if no essential difference

codegen/type_converter.go
```diff
@@ -638,25 +689,61 @@ func (c *TypeConverter) genStructConverter(
...
 				stFromPrefix = keyPrefix
 				stFromType   compile.TypeSpec
+				fromTypeName string
 			)
 			if fromField != nil {
 				stFromType = fromField.Type
 				stFromPrefix = keyPrefix + pascalCase(fromField.Name)
+
+				var err error
+				fromTypeName, err = c.getIdentifierName(stFromType)
```
cl1337: if you're not handling the error, this does the same:
```
//var err error
fromTypeName, _ = c.getIdentifierName(stFromType)
// if err != {
// fromTypeName = ""
// }
```
and why don't we handle the error here?

codegen/type_converter.go
```diff
@@ -638,25 +689,61 @@ func (c *TypeConverter) genStructConverter(
...
+
+			toTypeName, err := c.getIdentifierName(toFieldType)
 			if err != nil {
 				return err
 			}
+
+			if converterMethodName, ok := c.convStructMap[toFieldType.Name]; ok {
+				// the converter for this struct has already been generated, so just use it
+				c.append(indent, "out.", keyPrefix+pascalCase(toField.Name), " = ", converterMethodName, "(", fromIdentifier, ")")
+			} else if c.useRecurGen && fromTypeName != "" {
```
cl1337: I always prefer `len(fromTypeName) < 1`, string len compare is cheaper than string compare

codegen/type_converter.go
```diff
@@ -638,25 +689,61 @@ func (c *TypeConverter) genStructConverter(
...
+					toTypeName,
+					toSubIdentifier,
+					fromTypeName,
+					fromIdentifier,
+					stFromType,
+					fieldMap,
+					prevKeyPrefixes,
+					indent,
+				)
+				if err != nil {
```
cl1337: if you don't handle err, do directly return 

```
return c.genConverterForStructWrapped{
```

and again, should you log / handle here?

codegen/type_converter.go
jyang0: I agree it is quite out of control.. I'm open to alternate solutions, but the dilemma here is to have recursive sub-methods (only way to handle recursive structs?), which are naturally local to a type, but also support fieldMap, whose mapping could be from anywhere. How else would this local sub-method have access to any arbitrary variable?

codegen/type_converter.go
jyang0: I may have missed a line, but I think I am using []string{} for arrays and make() for maps

codegen/type_converter.go
jyang0: This is how the nested generated functions disappeared.. 😅 When useRecurGen is false the code generation looks exactly like before, otherwise it's what you saw earlier (without the blank {} scopes), and it's only true if necessary, i.e. when the input or output have some recursive element

codegen/type_converter.go
jyang0: True, do you want me to remove it? I thought it could be nice to have for clarity

codegen/type_converter.go
jyang0: I hope the compiler is magical enough to save me😅

codegen/type_converter.go
jyang0: Hmm okay I'll add a utility method to do this

codegen/type_converter.go
oliviazhang0809: ok. It's generally not the best practice to call the same function twice and skip error. Can we optimize that?

codegen/type_converter.go
oliviazhang0809: what i meant is to have 
```
const (
	inOriginal = "inOriginal"
	outOriginal = "outOriginal"
)
```
upfront -- global scope and use the variable in functions. - genernally best practise.

codegen/type_converter.go
jyang0: There's ```c.getIdentifierName(toFieldType)``` and ```c.getIdentifierName(stFromType)```. For toFieldType, I added a call upstream, with normal error checking, so the downstream one should be fine. For stFromType, it wasn't there before and in some cases it fails (when fieldMap is used a certain way), which is fine because I only need it to generate the function wrappers, and in cases when fieldMap is used in that way, no wrapper is necessary (I just don't generate a wrapper when it fails)

codegen/type_converter.go
jyang0: Oh I see. Though Go is not letting me write
```
const (
    inOriginal = in
)
```
because ```in``` is not a constant

codegen/type_converter.go
jyang0: I agree. I'd save the value and pass it down, but ```genConverterForStruct``` is also called from ```genConverterForMap``` and ```genConverterForList```, which already seem to have their own separate type validity checks, so I'm not able to produce an error from ```c.getIdentifierName(toFieldType)``` coming from the either of them (it errors beforehand)

oliviazhang0809: sure let's leave as it is.

ChuntaoLu: We should drop a comment here for why the error check is skipped.

PR #318, opened 2018-01-22
small fixes for datacenter redirect tests

jakelarkn:


runtime/gateway.go
```diff
@@ -297,7 +297,7 @@ func (gateway *Gateway) setupConfig(config *StaticConfig) {
 
 		config.SetSeedOrDie("datacenter", string(bytes))
 	} else {
-		config.SetSeedOrDie("datacenter", "unknown")
+		config.SetConfigValueOrDie("datacenter", []byte("unknown"), "string")
```
jakelarkn: seed value is not over-writeable by test input

test/lib/bench_gateway/bench_gateway.go
```diff
@@ -122,6 +122,9 @@ func CreateGateway(
 				ExpectContinueTimeout: 30 * time.Second,
 			},
 			Timeout: 30 * 1000 * time.Millisecond,
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
```
jakelarkn: golang http.client follows redirects by default

runtime/gateway.go
cl1337: mm ... the original impl makes sense, it will freeze config after set here, even `SetConfigValueOrDie` will fail after config frozen? How does test overwrite this

jakelarkn: from what I see tests rely on the seed values to inject config. StaticConfig does not allow Seed to be overwritten, even before .freeze(), so this always errors if the test tries to set "datacenter". I don't really see the value of hard set of "unknown" as a seed value vs just a normal config value.

jakelarkn: https://github.com/uber/zanzibar/blob/master/runtime/static_config.go#L323

ChuntaoLu: Makes sense, I think this special case is worth a few lines of comments.

PR #322, opened 2018-01-30
parse method names from thrift code

rpatali:
Frontent requires an list of method names given thrift code.
thrift compilation creates a struct with service and method names which
isn't necessarily ordered but converting that to a list for the frontend
assumes ordering which may break tests hence always sort the list before
returning.

### DIFF
--- a/backend/repository/handler.go
+++ b/backend/repository/handler.go
@@ -402,13 +402,17 @@ func (h *Handler) ThriftCodeToMethods(w http.ResponseWriter, r *http.Request, _ 
 
 	_, err := UnmarshalJSONBody(r, req)
 	if err != nil {
-		h.WriteErrorResponse(w, http.StatusBadRequest, errors.Wrap(err, "Failed to unmarshal request body for converting to a list of methods"))
+		h.WriteErrorResponse(w, http.StatusBadRequest, errors.Wrap(err, "Raw code failed unmarshalling."))
+		return
+	}
+	if req.RawCode == "" {
+		h.WriteErrorResponse(w, http.StatusBadRequest, errors.Errorf("Raw code is empty."))
 		return
 	}
 
 	module, err := h.Manager.CompileThriftCode(req.RawCode)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


backend/repository/handler.go
```diff
@@ -386,6 +388,44 @@ func (h *Handler) ThriftModuleToCode(w http.ResponseWriter, r *http.Request, ps
...
+	Functions []string `json:"functions"`
+}
+
+// ThriftCodeToMethods returns a list of method names for a given thrift file path
+func (h *Handler) ThriftCodeToMethods(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	req := &thriftCodeToMethodsRequest{}
+
+	_, err := UnmarshalJSONBody(r, req)
+	if err != nil {
+		h.WriteErrorResponse(w, http.StatusBadRequest, errors.Wrap(err, "Raw code failed unmarshalling."))
```
ChuntaoLu: The convention for the wrap message is starting with lower case with no punctuation at the end. https://godoc.org/github.com/pkg/errors#ex-Wrap

backend/repository/manager.go
```diff
@@ -179,6 +179,11 @@ func (m *Manager) CompileThriftFile(gateway, path string) (*Module, error) {
 	return CompileThriftFile(absPath, prefix)
 }
 
+// CompileThriftCode parses raw bytes in to compiled thrift.
+func (m *Manager) CompileThriftCode(rawCode string) (*Module, error) {
+	return CompileThriftCode([]byte(rawCode), "", "")
```
cl1337: why the wrapper  / why not support include

rpatali: - wrapper: All of the code that uses thrift compilation goes through Manager. Seems like Manager is acting as an abstraction on top of thrift compiler operations. I didn't want to break the abstraction (generally, seems like a decent idea too).
- include support: the test cases does have include `base.thrift` in there and the compilation succeeds. The controller method isn't looking for the included function definition as it isn't a use case we have from for this controller. We can always come back and update the controller.

PR #323, opened 2018-01-31
Add delete client, endpoint, thrift mechanism

jyang0:
Added additional fields to UpdateRequest to support deletions:
```
type UpdateRequest struct {
	ThriftFiles        []string            `json:"thrift_files"`
	ManagedThriftFiles []ManagedThriftFile `json:"managed_thrift_files"`
	ClientUpdates      []ClientConfig      `json:"client_updates"`
	EndpointUpdates    []EndpointConfig    `json:"endpoint_updates"`
	DIffID             *string             `json:"diff_id,omitempty"`
	**DeleteThrifts      []string            `json:"delete_thrifts,omitempty"`
	DeleteClients      []string            `json:"delete_clients,omitempty"`
	DeleteEndpoints    []string            `json:"delete_endpoints,omitempty"`**
}
```


jyang0:
Going to reopen soon with changes

backend/repository/manager.go
```diff
@@ -28,6 +28,7 @@ import (
 	"go.uber.org/thriftrw/compile"
 
 	"github.com/pkg/errors"
+	"strings"
```
oliviazhang0809: nit import order

backend/repository/manager.go
```diff
@@ -321,6 +364,16 @@ func (m *Manager) UpdateAll(r *Repository, clientCfgDir, endpointCfgDir string,
 	if err := m.UpdateEndpoints(r, endpointCfgDir, req.EndpointUpdates); err != nil {
 		return errors.Wrap(err, "failed to update endpoints")
 	}
+
+	if err := m.DeleteEndpoints(r, endpointCfgDir, req.DeleteEndpoints); err != nil {
+		return errors.Wrap(err, "failed to delete endpoints")
```
oliviazhang0809: Err has been set in previous request - should be able to return `err` directly.

backend/repository/manager.go
```diff
@@ -321,6 +364,16 @@ func (m *Manager) UpdateAll(r *Repository, clientCfgDir, endpointCfgDir string,
...
+	}
+	if err := m.DeleteClients(r, clientCfgDir, req.DeleteClients); err != nil {
+		return errors.Wrap(err, "failed to delete clients")
```
oliviazhang0809: same here and others

backend/repository/manager.go
jyang0: What does that mean?

backend/repository/manager.go
jyang0: Okay, I'll change this

examples/example-gateway/endpoints/googlenow/endpoint-config.json
```diff
@@ -4,8 +4,8 @@
 	"config": {
 		"rateLimit": 100,
 		"endpoints": [
-			"./add_credentials.json",
-			"./check_credentials.json"
+			"add_credentials.json",
```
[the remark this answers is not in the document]
jyang0: My current delete endpoint implementation expects the entry inside the array to be [handleId in snake case].json, which is exactly the convention used when an endpoint is added in the code. It seems that some of the examples don't follow this convention. To support a prefix like `./` , I can do an absolute path comparison, which would eliminate something like `./`. To support a filename that isn't [handleId in snake case], I'll have to basically try to read every file listed in the array and check its contents to see which endpoint it is. What do you think is the best way?

backend/repository/manager.go
oliviazhang0809: https://github.com/golang/go/wiki/CodeReviewComments#imports

examples/example-gateway/endpoints/googlenow/endpoint-config.json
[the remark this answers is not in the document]
oliviazhang0809: which example doesn't match the snake case matching? I think we need to follow the same naming convention. Either way works as long as it's consistent

backend/repository/thrift_update.go
```diff
@@ -144,3 +145,72 @@ func (r *Repository) ThriftFileVersion(thriftFile string) (string, error) {
...
+		}
+	}
+
+	meta, err := r.ThriftConfig(cfg.ThriftRootDir)
+	if err != nil {
+		return errors.Wrap(err, "failed to read current thrift meta")
+	}
+
+	// Delete entry	in meta json and write to file
+	delete(meta, thriftFile)
```
oliviazhang0809: will there be an issue if one client opens a diff to delete thrift (ok) while another client opens a diff to add client based on the thrift? That two diffs open at the same time.

backend/repository/manager.go
```diff
@@ -321,6 +364,16 @@ func (m *Manager) UpdateAll(r *Repository, clientCfgDir, endpointCfgDir string,
 	if err := m.UpdateEndpoints(r, endpointCfgDir, req.EndpointUpdates); err != nil {
 		return errors.Wrap(err, "failed to update endpoints")
 	}
+
+	if err := m.DeleteEndpoints(r, endpointCfgDir, req.DeleteEndpoints); err != nil {
+		return errors.Wrap(err, "failed to delete endpoints")
+	}
+	if err := m.DeleteClients(r, clientCfgDir, req.DeleteClients); err != nil {
```
oliviazhang0809: does this get runs every time when `UpdateAll` is called? I see you created separated path per update thrift/endpoint/clients and some checkings are duplicated. Can we consolidate the delete paths to update path?

examples/example-gateway/endpoints/googlenow/endpoint-config.json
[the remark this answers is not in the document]
jyang0: There could be more, but bar.helloWorld has handleId helloWorld, but the json is named hello.json. I agree we need to follow the same convention. Is the convention defined as how the code adds a new endpoint? If so, then the examples aren't following it consistently. Would you prefer I adjust the examples to follow them, or change my delete implementation so it also supports the current examples?

backend/repository/thrift_update.go
jyang0: I guess that depends if once the 1st diff is landed, will the 2nd diff get re-built based on the new master? If so, the build should fail right

backend/repository/manager.go
jyang0: It does get run every time, but the overhead should be minimal when there aren't any deletions. I originally used the update path for delete, but I switched to a separate one since:

- I can add a flag to ClientConfig and EndpointConfig to indicate deletion, but I'm still going to need a separate path to delete thrifts, since there's no ThriftConfig, just a string right now
- It seems that a "deleteThis" flag doesn't really belong inside a __Config struct. Also, it seems a bit unnecessary to use a __Config struct when only the name and flag are needed for deletion, and this will introduce a new path anyway (e.g. skip the usual input struct validation)

I'm open to other solutions

backend/repository/endpoint_update.go
```diff
@@ -69,6 +69,114 @@ func (r *Repository) WriteEndpointConfig(
...
 }
 
+// DeleteEndpointConfig deletes an individual endpoint config, updates endpoint group config and endpoint meta json
+// endpointName must have the form endpointId.handleId
+func (r *Repository) DeleteEndpointConfig(endpointCfgDir string, endpointName string) error {
+	gatewayCfg, err := r.LatestGatewayConfig()
+	if err != nil {
+		return errors.Wrap(err, "loading gateway config error")
+	}
+	nameParts := strings.Split(endpointName, ".")
```
cl1337: dependency resolution should be in one uniq pkg, answer such question of what depend on what, use case I can think of:

- partial tests
- delete endpoints
- notify client owners for new user case 

... and more

it should not be each module implement their own logic like this

`codegen.ModuleDep` pkg, expose functions 
`GetDepFrom(client, endpoint)` or something

@ChuntaoLu

backend/repository/client_update.go
```diff
@@ -100,6 +100,39 @@ func (r *Repository) UpdateClientConfigs(req *ClientConfig, clientCfgDir, thrift
 	return nil
 }
 
+// DeleteClientConfigs deletes JSON configuration files for a client update request
+func (r *Repository) DeleteClientConfigs(clientName string, clientCfgDir string) error {
```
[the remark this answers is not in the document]
jyang0: Mhm makes sense, I'll change

backend/repository/manager.go
```diff
@@ -236,7 +237,46 @@ func (m *Manager) UpdateClients(r *Repository, clientCfgDir string, req []Client
...
+
+// DeleteClients deletes the list of specified clients
+func (m *Manager) DeleteClients(r *Repository, clientCfgDir string, clients []string) error {
+	clientDep, err := r.GetAllClientDependencies()
+	if err != nil {
+		return errors.Wrap(err, "failed to get list of dependent endpoints per client")
+	}
+
+	for _, client := range clients {
+		if deps, found := clientDep[client]; found {
```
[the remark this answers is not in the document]
jyang0: Maybe I should rename it, but no, right now it just finds all dependent endpoints for each client

examples/example-gateway/endpoints/googlenow/endpoint-config.json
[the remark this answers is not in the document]
jyang0: I guess for best compatibility I'll make it so that if there's no match based on the absolute path, it'll try to read each json to see if the endpoint is actually there

backend/repository/endpoint_update.go
```diff
@@ -69,6 +69,116 @@ func (r *Repository) WriteEndpointConfig(
...
+		if err = writeToJSONFile(endpointGroupConfigFile, groupConfig); err != nil {
+			return errors.Wrapf(err, "failed to write updated endpoint group config for %s", endpointConfig.ID)
+		}
+	}
+
+	return nil
+}
+
+func getEndpointsWithMiddleware(gatewayCfg *Config, middleware string, endpointGroup string) []string {
+	endpoints := make([]string, 0, 2)
```
cl1337: mm ... maybe just `make([]string,0)`

backend/repository/endpoint_update.go
```diff
@@ -69,6 +69,116 @@ func (r *Repository) WriteEndpointConfig(
...
+	for _, endpoint := range gatewayCfg.Endpoints {
+		if endpoint.ID == endpointGroup {
```
cl1337: why not just use a map here? You can still `len(map)`

backend/repository/endpoint_update.go
```diff
@@ -69,6 +69,116 @@ func (r *Repository) WriteEndpointConfig(
...
+			if i := findString(endpointConfig.ClientID, groupConfig.Dependencies["client"]); i != -1 {
+				groupConfig.Dependencies["client"] = append(groupConfig.Dependencies["client"][:i], groupConfig.Dependencies["client"][i+1:]...)
+			}
+		}
+		// Remove middleware dependency from group config if no other endpoints in this group use it
+		for _, m := range endpointConfig.Middlewares {
+			otherEndpoints = getEndpointsWithMiddleware(gatewayCfg, m.Name, endpointConfig.ID)
+			if len(otherEndpoints) == 1 {
+				if i := findString(m.Name, groupConfig.Dependencies["middleware"]); i != -1 {
+					groupConfig.Dependencies["middleware"] = append(groupConfig.Dependencies["middleware"][:i], groupConfig.Dependencies["middleware"][i+1:]...)
```
cl1337: this one liner seems a bit implicate  ...

backend/repository/endpoint_update.go
jyang0: Okay

backend/repository/endpoint_update.go
jyang0: I guess in this case there's not much point to having a key, since the returned list is already limited to an endpoint group, and it can't be usefully stored since it might change when an endpoint is deleted

PR #324, opened 2018-02-02
expose transform middleware in repository API

jakelarkn:
tranform features are not technically a middleware module as defined by
the module system.  the endpoint middleware stack codegen braches on
transforms to use the type converter.  However transform middlewares
config schema still need to be exposed in the repository API for orchestration
so thats what this accomplishes.

### DIFF
--- a/backend/repository/config.go
+++ b/backend/repository/config.go
@@ -378,9 +378,30 @@ func (r *Repository) endpointConfigs(thriftRootDir string, gatewaySpec *codegen.
 func (r *Repository) middlewareConfigs(
 	gatewaySpec *codegen.GatewaySpec,
 ) (map[string]interface{}, map[string]*RawMiddlewareConfig, error) {
-	middlewareSpecs := gatewaySpec.MiddlewareModules
+	orgMiddlewareSpecs := gatewaySpec.MiddlewareModules
 	schemasByName := make(map[string]interface{})
 	middlewareConfigs := make(map[string]*RawMiddlewareConfig)
+
+	// manually add transform request response since they are features of the zanzibar platform
+	//  Note:  the schema files must be available in the middlewares directory
+	tq := &codegen.MiddlewareSpec{
+		Name:              "transformRequest",
+		OptionsSchemaFile: "./middlewares/transform-request/transform_request_schema.json",
+		ImportPath:        "github.com/uber/zanzibar/codegen",
+	}
+
+	ts := &codegen.MiddlewareSpec{
+		Name:              "transformResponse",
+		OptionsSchemaFile: "./middlewares/transform-response/transform_response_schema.json",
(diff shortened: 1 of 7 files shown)


backend/repository/config.go
```diff
@@ -378,10 +378,31 @@ func (r *Repository) endpointConfigs(thriftRootDir string, gatewaySpec *codegen.
 func (r *Repository) middlewareConfigs(
 	gatewaySpec *codegen.GatewaySpec,
 ) (map[string]interface{}, map[string]*RawMiddlewareConfig, error) {
-	middlewareSpecs := gatewaySpec.MiddlewareModules
+	orgMiddlewareSpecs := gatewaySpec.MiddlewareModules
 	schemasByName := make(map[string]interface{})
 	middlewareConfigs := make(map[string]*RawMiddlewareConfig)
 
+	// manually add transform request response since they are features of the zanzibar platform
+	//  Note:  the schema files must be in available the middlewares directory
```
rpatali: typo: `in available` -> `available in`.

backend/repository/config.go
```diff
@@ -378,10 +378,31 @@ func (r *Repository) endpointConfigs(thriftRootDir string, gatewaySpec *codegen.
 func (r *Repository) middlewareConfigs(
 	gatewaySpec *codegen.GatewaySpec,
 ) (map[string]interface{}, map[string]*RawMiddlewareConfig, error) {
-	middlewareSpecs := gatewaySpec.MiddlewareModules
+	orgMiddlewareSpecs := gatewaySpec.MiddlewareModules
```
cl1337: any specific reason we're not doing this in `parseMiddlewareConfig`? it seems to be one single place parsing middlewares, files or not.

jakelarkn: thought about that.  mostly, I was worried that putting them on GatewaySpec would mess something up with the module system  since I think the middleware handler codegen runs off of that list,  also  I think permeating the special casing deeper into the core  just adds complexity.     Pretty easy to revisit if we look at it more.

cl1337: how we create middleware specs (should have all information with all types ... )
and how middleware specs are used, like in here:https://github.com/uber/zanzibar/blob/master/codegen/gateway.go#L645
(it's literally check `transformRequest` and skip it)
is separate concerns.

If middleware consumer mess up we should fix the consumer.

Anyway just raised the concern that middleware should be generated in one place

jakelarkn: the way transforms are implemented was a hack into middleware  i would be more interested in the idea of moving x-form spec and generation to Endpoint.   For the most part this is trying to quickly fix a bug between a few large incompatible subsystems Codegen, Orchestrater API and Module System.

cl1337: 👍

PR #325, opened 2018-02-02
support annotation prefix config

cl1337:
this PR allows annotation prefix to be specified in project config instead hardcode in zanzibar lib
it helps project migration (if they have to keep original annotations), and schema used by different services

@uber/zanzibar-team

### DIFF
--- a/codegen/method.go
+++ b/codegen/method.go
@@ -29,6 +29,22 @@ import (
 	"go.uber.org/thriftrw/compile"
 )
 
+const (
+	antHTTPMethod     = "%s.http.method"
+	antHTTPPath       = "%s.http.path"
+	antHTTPStatus     = "%s.http.status"
+	antHTTPReqHeaders = "%s.http.reqHeaders"
+	antHTTPResHeaders = "%s.http.resHeaders"
+	antHTTPRef        = "%s.http.ref"
+	antMeta           = "%s.meta"
+	antHandler        = "%s.handler"
+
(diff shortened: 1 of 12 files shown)


codegen/method.go
```diff
@@ -186,20 +201,20 @@ func NewMethod(
 		return nil, err
 	}
 
-	method.ReqHeaders = headers(funcSpec.Annotations[antHTTPReqHeaders])
-	method.ResHeaders = headers(funcSpec.Annotations[antHTTPResHeaders])
+	method.ReqHeaders = headers(funcSpec.Annotations[method.annotationMap[antHTTPReqHeaders]])
```
jakelarkn: lookup through the annotationMap seems kind of clunky   i would suggest that you just make the fmt'd strings properties of   MethodSpec  or  a annotations struct

jakelarkn: additionally  do we expect one annotation prefix per gateway,  or a multiple different prefixes  (rtapi, zanzibar, edge, etc) ?     if it's just one  you could potentially store the fmt'd strings in a project scope  rather than recalc'ing them for every method

cl1337: that's correct, annotation prefix is a project scope variable. 

prefix has to resolved from config. I can, potentially create a project global object, to pass global scope down to `methodSpec` here, or equip `methodSpec` with `packageHelper`.

But I'd refrained  to do things like this to allow method generation have access to global state generated miles away during startup

The other reason why this has to be chunky is one annotation is used outside `codegen` in a script to parse thrift

PR #326, opened 2018-02-06
propagate causing error to resp logger

jakelarkn:


jakelarkn:
there are a bunch of req validations (HTTP 400) that don't currently create a causing error .   the message for those is pretty self explanitory  (compared to "Unexpected Server Error")

jakelarkn:
looking through others..  will update where possible

jakelarkn:
updated

PR #328, opened 2018-02-12
Support mock clients based integration testing

ChuntaoLu:
This PR introduces a new integration testing approach based on mock clients.

The current integration testing approach is based on mock backends: when we run an integration test, a bunch of mock backend servers get span up in a separate process to serve fixture responses. This approach has its pros, but also comes with many problems, including client and backend linking, inefficiency, abstractions hard to reason about, flaky tests, etc.

This PR takes a different approach: it generates a test service by mocking the leaf nodes (clients) in the dependency tree, and the test service runs in a separate goroutine in the same process as the test. The client mocks are generated via Mockery and currently exposed on the test service to control the client behavior. This is a first step toward building higher level testing abstractions with mocked clients.

[Example](https://github.com/uber/zanzibar/pull/328/files#diff-8c6e6f2e2f5e68f2b911bbebbd7f1b91) of how an integration can be written with this new approach:
``` go
func TestBazCall(t *testing.T) {
	ms := ms.MustCreateTestService()
	ms.Start()
	defer ms.Stop()

	// fixtures
	reqHeaders := map[string]string{
		"x-token": "token",
		"x-uuid":  "uuid",
	}
	args := &baz.SimpleService_Call_Args{
		Arg: &baz.BazRequest{
			B1: true,
			S2: "hello",
			I3: 42,
		},
	}
	expectedHeaders := map[string]string{
		"some-res-header": "something",
	}

	ctx := context.Background()
	var result baz.SimpleService_Call_Result
	ms.MockClientNodes().Baz.On("Call", mock.Anything, reqHeaders, mock.Anything).
		Return(map[string]string{"some-res-header": "something"}, nil)

	success, resHeaders, err := ms.MakeTChannelRequest(
		ctx, "SimpleService", "Call", reqHeaders, args, &result,
	)
	if !assert.NoError(t, err, "got tchannel error") {
		return
	}
	assert.True(t, success)
	assert.Equal(t, expectedHeaders, resHeaders)
}
```

It is worthy to compare the run time difference:
```
➜  zanzibar git:(lu.mock)  go test github.com/uber/zanzibar/test/endpoints/tchannel/baz -run ^TestCallTChannelSuccessfulRequestOKResponse$ -count 100
ok  	github.com/uber/zanzibar/test/endpoints/tchannel/baz	8.918s
➜  zanzibar git:(lu.mock)  go test -tags mock github.com/uber/zanzibar/examples/example-gateway/endpoints/tchannel/baz -run ^TestBazCall$ -count 100
ok  	github.com/uber/zanzibar/examples/example-gateway/endpoints/tchannel/baz	1.927s
```
Note the `-tags mock`, without which the generated mock service is excluded from the build.
 
@uber/zanzibar-team


codegen/runner/post-steps.sh
```diff
@@ -1,18 +1,36 @@
...
+if [ -d "$DIRNAME/../../vendor" ]; then
+	MOCKERY_RAW_DIR="$DIRNAME/../../vendor/github.com/vektra/mockery"
+else
+	MOCKERY_RAW_DIR="$DIRNAME/../../../../../github.com/vektra/mockery"
+fi
+MOCKERY_DIR="$(cd "$MOCKERY_RAW_DIR";pwd)"
+MOCKERY_MAIN_FILE="$MOCKERY_DIR/cmd/mockery/mockery.go"
+MOCKERY_BINARY="$MOCKERY_DIR/cmd/mockery/mockery"
+
+go build -o "$MOCKERY_BINARY" "$MOCKERY_MAIN_FILE"
```
jacobgreenleaf: You can invoke go binaries directly using `go run $MOCKERY_MAIN_FILE arg1 arg2 ... argN`

codegen/templates/service_mock.tmpl
```diff
@@ -0,0 +1,160 @@
...
+	httpClient     *http.Client
+	tChannelClient zanzibar.TChannelCaller
+}
+
+// TestOptions is the info needed to create a test service
+type TestOptions struct {
+	TChannelClientMethods map[string]string
+}
+
+// NewTestService creates a new MockService
```
jacobgreenleaf: nit: this function has a signature `(MockService, error)` but never returns the error and instead panics. 

If you want it to panic instead of returning the error, it should be called something like `MustCreateTestService`. Otherwise you can create a wrapper like `MustCreateTestService(NewTestService(...))` similar to [`template.Must`](https://golang.org/pkg/text/template/#Must).

codegen/templates/service_mock.tmpl
```diff
@@ -0,0 +1,160 @@
...
+	started        bool
+	server         *zanzibar.Gateway
+	{{camel $mockType}}    *module.{{$mockType}}
+	httpClient     *http.Client
+	tChannelClient zanzibar.TChannelCaller
+}
+
+// TestOptions is the info needed to create a test service
+type TestOptions struct {
+	TChannelClientMethods map[string]string
```
jacobgreenleaf: Please add a line comment for this, since it's what the developer will inspect first to figure out how to pass information to the test runner.

codegen/templates/service_mock.tmpl
ChuntaoLu: It's an oversight that no error was returned, I like the `Must` approach, will change it to that.

codegen/templates/service_mock.tmpl
ChuntaoLu: Totally agree that it is ambiguous and clueless. I actually want to completely hide it from tests because the way a user send request to the test service is via `MakeTChannelRequest` which takes the thrift service and method names from endpoint thrift idl, the tchannel client is not exposed and the `ZanzibarMethodName` is not used at all for whoever writes the test.

The `TChannelClientMethods` is used to create a client and the "ThriftService::method" to "ZanzibarMethodName" map is only useful for logs and metrics of that client, which can be occasionally helpful for debugging tests.

I will probably not exposed this at all or expose optionally. But the comment for the map is on point and I will put it where it should be.

examples/example-gateway/build/clients/bar/mock_client.go
```diff
@@ -0,0 +1,974 @@
+// Code generated by mockery v1.0.0
+
+// +build mock
+
+package barClient
+
+import bar "github.com/uber/zanzibar/examples/example-gateway/build/gen-code/clients/bar/bar"
```
cl1337: can we fix import here?

scripts/generate.sh
```diff
@@ -12,6 +12,6 @@ end=`date +%s`
 runtime=$((end-start))
 echo "Generated endpoints/clients : +$runtime"
 
-bash ./codegen/runner/post-steps.sh "$PREFIX/build"
+bash ./codegen/runner/post-steps.sh "$PREFIX/build" "$PREFIX"
```
cl1337: mm ... seems your $1 == $2 + /build

codegen/template.go
```diff
@@ -93,6 +94,10 @@ func unref(t string) string {
 	return t
 }
 
+func args(v ...interface{}) []interface{} {
```
cl1337: mm ... is this used in templates? Why array instead of kv pair, is order matter?

examples/example-gateway/build/clients/bar/mock_client.go
ChuntaoLu: This is generated by Mockery.

scripts/generate.sh
ChuntaoLu: It is coincidental, the post-steps.sh does not assume any relationship between the two params.

codegen/template.go
ChuntaoLu: Yes, it's used. This function combines variadic params into one param because the `template` function for nested templates only takes one param.

ChuntaoLu:
@cl1337 
> is this integration test for tchannel client only? Didn't see endpoint / http examples

I guess the question is  " is this integration test for tchannel **endpoint** only?", no, it's for all endpoints. I will add a http endpoint example.

> how will this mock tests affect user input test fixtures?

This PR does not affect user test fixtures because there is no interface change for endpoints and clients. How fixtures are accessed will be worked on, but not in the scope of this PR.

backend/repository/endpoint_update.go
```diff
@@ -161,7 +161,7 @@ func (r *Repository) deleteEndpointConfig(endpointCfgDir string, endpointName st
 }
 
 func getEndpointsWithMiddleware(gatewayCfg *Config, middleware string, endpointGroup string) []string {
-	endpoints := make([]string, 0)
+	var endpoints []string
```
srivastavankit: Why is this better? This creates a slice of size 0, that will exponentially grow as we append. Can we not do something like

```
endpoints := make([]string, 0, len(gatewayCfg.Endpoints))
```

examples/example-gateway/clients/quux/quux.go
```diff
@@ -22,17 +22,20 @@ package quux
...
-type Client interface{}
+type Client interface {
+	Echo(string) string
+}
 
 type quux struct{}
 
 // NewClient creates a new Quux client
-func NewClient(g *zanzibar.Gateway, deps *module.Dependencies) Client {
+func NewClient(deps *module.Dependencies) Client {
```
srivastavankit: nice!

examples/example-gateway/config/test.json
```diff
@@ -27,15 +27,15 @@
 	"clients.multi.ip": "127.0.0.1",
 	"clients.multi.port": 4003,
 	"clients.multi.timeout": 10000,
-	"http.port": 7783,
-	"logger.fileName": "/var/log/example-gateway/example-gateway.log",
+	"http.port": 0,
+	"logger.fileName": "./example-gateway.log",
```
srivastavankit: will this work fine for nested folders and if we invoke tests from there?

glide.lock
```diff
@@ -96,6 +99,8 @@ imports:
   - tnet
   - trand
   - typed
+- name: github.com/vektra/mockery
+  version: b6f6c13c25c3f0a17724553fa01580f328158d7a
```
srivastavankit: how would we prevent people to create their own mocks?

runtime/tchannel_client_raw.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+
+	"github.com/uber-go/tally"
+	"github.com/uber/tchannel-go"
+	"go.uber.org/zap"
+)
+
+// RawTChannelClient is like TChannel client, but without ClientID or MethodNames.
+// Its Logs and metrics are not scoped per method, and not used for generate TChannel
```
srivastavankit: Does this mean that in a test environment we will not have same metrics/logs as production environment>

runtime/tchannel_client_raw.go
```diff
@@ -0,0 +1,101 @@
...
+// tchannel to the given thrift service. There is no guarantee that the given thrift
+// service and method is valid for given Channel.
+// It is intended to be used internally for testing.
+func NewRawTChannelClient(
+	ch *tchannel.Channel,
+	logger *zap.Logger,
+	scope tally.Scope,
+	opt *TChannelClientOption,
+) *RawTChannelClient {
+	clientID := "raw"
```
srivastavankit: create a const for raw

runtime/tchannel_client_raw.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Call makes a RPC call to the given service.
+func (r *RawTChannelClient) Call(
+	ctx context.Context,
+	thriftService, methodName string,
+	reqHeaders map[string]string,
+	req, resp RWTStruct,
+) (success bool, resHeaders map[string]string, err error) {
+	serviceMethod := thriftService + "::" + methodName
```
srivastavankit: use a buffer?

Raynos:
```
raynos at raynos-ThinkPad-T440p  
~/gocode/src/github.com/uber/zanzibar on lu.mock
$ ZANZIBAR_CACHE=1 /usr/bin/time -p go test github.com/uber/zanzibar/test/endpoints/tchannel/baz -run ^TestCallTChannelSuccessfulRequestOKResponse$ -count 100
ok  	github.com/uber/zanzibar/test/endpoints/tchannel/baz	9.200s
real 13.96
user 19.95
sys 2.15
raynos at raynos-ThinkPad-T440p  
~/gocode/src/github.com/uber/zanzibar on lu.mock
$ ZANZIBAR_CACHE=1 /usr/bin/time -p go test -tags mock github.com/uber/zanzibar/examples/example-gateway/endpoints/tchannel/baz -run ^TestBazCall$ -count 100
ok  	github.com/uber/zanzibar/examples/example-gateway/endpoints/tchannel/baz	3.697s
real 12.27
user 22.03
sys 1.55
raynos at raynos-ThinkPad-T440p  
~/gocode/src/github.com/uber/zanzibar on lu.mock
$ ZANZIBAR_CACHE=1 /usr/bin/time -p go test github.com/uber/zanzibar/test/endpoints/tchannel/baz -run ^TestCallTChannelSuccessfulRequestOKResponse$ -count 100
ok  	github.com/uber/zanzibar/test/endpoints/tchannel/baz	9.238s
real 14.41
user 19.89
sys 2.34
raynos at raynos-ThinkPad-T440p  
~/gocode/src/github.com/uber/zanzibar on lu.mock
$ ZANZIBAR_CACHE=1 /usr/bin/time -p go test -tags mock github.com/uber/zanzibar/examples/example-gateway/endpoints/tchannel/baz -run ^TestBazCall$ -count 100
ok  	github.com/uber/zanzibar/examples/example-gateway/endpoints/tchannel/baz	3.683s
real 11.71
user 22.28
sys 1.46
```

Watch out when using the "go test" reported test time. 
It's generally not accurate and your better of using unix `time` to measure time taken to run tests. 
Observe real time is reduced but user time is increased, quite funny.

The mocking approach still has other benefits not related to wall clock time.

backend/repository/endpoint_update.go
ChuntaoLu: It simply passes the linter, no idea how it is on master and passing linters.

glide.lock
ChuntaoLu: Mock clients are generated and baked in to the dependency tree, you can create arbitrary mock clients by yourself, but the server in your integration test won't use those mocks.

runtime/tchannel_client_raw.go
ChuntaoLu: We will have the metrics/logs, just not tagged with a `ClinetID` and `MethodName`, e.g., for a generated client named `Foo` with a method `Bar` that does thrift rpc to `MyService::myMethod`, the logs/metrics for that client will  have `Foo`, `Bar`, `MyService` and `myMethod` tags, while the raw client only has `MyService` and `myMethod` tags.

ChuntaoLu:
@Raynos Thanks for the tip, it's good to know the difference between the go test measured time and elapsed time.

backend/repository/endpoint_update.go
jacobgreenleaf: I'm not sure about this repo but it is possible in Github to enforce CI passing before being allowed to land on master using the Github UI. Perhaps we should turn that on to nudge people to pass CI including lint?

codegen/module_system.go
```diff
@@ -1176,6 +1207,20 @@ func GenerateInitializer(
 	)
 }
 
+// GenerateMockInitializer is like GenerateInitializer but with leaf nodes being mocks.
+func GenerateMockInitializer(
```
jacobgreenleaf: nit: The existing code doesn't have a function per template, you should stick with existing style.

codegen/runner/post-steps.sh
```diff
@@ -1,18 +1,36 @@
 #!/usr/bin/env bash
 set -e
+set -o pipefail
 
-if [ -z $1 ]; then
+if [ -z "$1" ]; then
 	echo "prefix argument (\$1) is missing"
```
jacobgreenleaf: nit: Create a usage function like this:

```
usage() { 
    echo "Usage: $0 [prefix] [config-dir]"
}
```

backend/repository/endpoint_update.go
ChuntaoLu: The repo has pre-push git hooks that enforces all checks to pass before pushing to remote, which I feel somewhat enough. But still I don't know how things like that was on master at the first place.

examples/example-gateway/endpoints/contacts/save_contacts_test.go
```diff
@@ -0,0 +1,41 @@
...
+		Contacts: []*endpointContacts.Contact{},
+	}
+	rawBody, _ := endpointReqeust.MarshalJSON()
+
+	clientRequest := &clientContacts.SaveContactsRequest{
+		UserUUID: "foo",
+	}
+	clientResponse := &clientContacts.SaveContactsResponse{}
+
+	ms.MockClientNodes().Contacts.On("SaveContacts", mock.Anything, mock.Anything, clientRequest).
```
jacobgreenleaf: While it's fine for now, the interfaces here are not very discoverable. It would be better if we wrapped `mockery`-generated mocks (whose interface is `On(methodName string, arguments ...interface{})` and `Return(returnArguments ...interface{})`) with the concrete types (in this example, `OnSaveContacts(context.Context,
		reqHeaders map[string]string,
		args *clientsContactsContacts.SaveContactsRequest)` and `Return(*SaveContactsResponse, error)`).

codegen/module_system.go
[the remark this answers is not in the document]
ChuntaoLu: I tried to followed the style with [GenerateInitializer](https://github.com/uber/zanzibar/blob/master/codegen/module_system.go#L1167), I guess the styles aren't consistent...

codegen/runner/post-steps.sh
```diff
@@ -1,18 +1,36 @@
 #!/usr/bin/env bash
 set -e
+set -o pipefail
 
-if [ -z $1 ]; then
+if [ -z "$1" ]; then
```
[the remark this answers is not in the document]
ChuntaoLu: Done

codegen/runner/post-steps.sh
```diff
@@ -1,18 +1,36 @@
...
+fi
 
 start=$(cat .TMP_ZANZIBAR_TIMESTAMP_FILE.txt)
 
-# TODO : Add new post steps :)
+BUILD_DIR="$1"
+CONFIG_DIR="$2"
+
+DIRNAME="$(dirname "$0")"
+if [ -d "$DIRNAME/../../vendor" ]; then
```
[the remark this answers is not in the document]
ChuntaoLu: That's a very good point. I have added some comments to explain the assumptions. The easiest solution is probably just to do a `go get`, while I am not comfortable to pull that into `GOPATH`, we probably already do a bunch of them already...

examples/example-gateway/endpoints/contacts/save_contacts_test.go
ChuntaoLu: That's true, the mockery `On` API is very raw and flexible. Concretizing the mock client API will be a next step in the scope of fixture work.

PR #330, opened 2018-02-15
fix crashes during create diff

cl1337:
This PR fixes some unguarded assign

### DIFF
--- a/backend/repository/config.go
+++ b/backend/repository/config.go
@@ -72,7 +72,7 @@ type EndpointConfig struct {
 	ClientID         string                                  `json:"clientId"`
 	ClientMethod     string                                  `json:"clientMethod"`
 	TestFixtures     map[string]*codegen.EndpointTestFixture `json:"testFixtures"`
-	Middlewares      []*EndpointMiddlewareConfig             `json:"middlewares"`
+	Middlewares      []EndpointMiddlewareConfig              `json:"middlewares"`
 	ReqHeaderMap     map[string]string                       `json:"reqHeaderMap"`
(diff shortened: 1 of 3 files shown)


backend/repository/config.go
```diff
@@ -72,7 +72,7 @@ type EndpointConfig struct {
 	ClientID         string                                  `json:"clientId"`
 	ClientMethod     string                                  `json:"clientMethod"`
 	TestFixtures     map[string]*codegen.EndpointTestFixture `json:"testFixtures"`
-	Middlewares      []*EndpointMiddlewareConfig             `json:"middlewares"`
+	Middlewares      []EndpointMiddlewareConfig              `json:"middlewares"`
```
cl1337: middlewares can be nil and crashes [here](https://github.com/uber/zanzibar/blob/master/backend/repository/endpoint_update.go#L201)

backend/repository/endpoint_update.go
```diff
@@ -241,13 +241,18 @@ func (r *Repository) findEndpointInGroup(endpointDir string, handleID string, js
...
 		err := readJSONFile(metaFilePath, fileContent)
 		if err != nil {
 			return err
 		}
 	}
-	endpoints := fileContent.Dependencies["endpoint"]
+	if fileContent.Dependencies != nil {
+		endpoints = fileContent.Dependencies["endpoint"]
+	} else {
+		fileContent.Dependencies = make(map[string][]string)
```
cl1337: dependency can come from thrift file instead of specify by UI

PR #332, opened 2018-02-16
fix map typedef key bug

jakelarkn:


codegen/type_converter_test.go
```diff
@@ -990,6 +990,80 @@ func TestConvertMapOfString(t *testing.T) {
...
+	} else {
+		out.UUIDMap[key1] = nil
+	}
+}
+	`), lines)
+}
+
+// Todo Test Override cases
+
+func TestConvertMapTypeDefToStruct(t *testing.T) {
```
jakelarkn: this is the bug we hit.  in the Map generation Typedef string alias (e.g. UUID) Key did have proper handling for Struct value

oliviazhang0809:
nice work!

PR #333, opened 2018-02-17
Replace mockery with mockgen for mock generation

ChuntaoLu:
This PR switches the mock generation lib from mockery to mockgen, because 
 - mockery does not support indirect interface types, such as type alias and type redeclaration 
 - mockgen is officially supported by golang

Unfortunately, mockgen currently does not support anonymous struct types (https://github.com/golang/mock/issues/153), so I have to temporarily unexpose client methods that consume/return anonymous struct types.


codegen/runner/post-steps.sh
```diff
@@ -23,26 +22,53 @@ start=$(cat .TMP_ZANZIBAR_TIMESTAMP_FILE.txt)
...
+#end=$(date +%s)
+#runtime=$((end-start))
+#echo "Compiled Mockery: +$runtime"
+#
+#"$MOCKERY_BINARY" -name="^Client$" -dir="$BUILD_DIR/clients" -inpkg -case=underscore -recursive -note="+build mock" > /dev/null
+#"$MOCKERY_BINARY" -name="^Client$" -dir="$CONFIG_DIR/clients" -inpkg -case=underscore -recursive -note="+build mock" > /dev/null
+
+if ! [ -x "$(command -v mockgen)" ]; then
+	go get -u github.com/golang/mock/gomock
+	go get -u github.com/golang/mock/mockgen
```
jacobgreenleaf: Tooling should be pinned in glide.yaml so we have reproducible results.

examples/example-gateway/endpoints/contacts/save_contacts_test.go
```diff
@@ -27,7 +27,7 @@ func TestSaveContactsCall(t *testing.T) {
 	}
 	clientResponse := &clientContacts.SaveContactsResponse{}
 
-	ms.MockClientNodes().Contacts.On("SaveContacts", mock.Anything, mock.Anything, clientRequest).
+	ms.MockClientNodes().Contacts.EXPECT().SaveContacts(gomock.Any(), gomock.Any(), clientRequest).
```
jacobgreenleaf: Nice, I like how the mocks are statically typed.

PR #340, opened 2018-02-27
fix linting error for camel case package name

cl1337:
this PR fix linting error from generated code caused by a linter change: https://github.com/golang/lint/commit/3ea3fa98a8104b2c8f8a7bffaebc7e54dddf99e1


codegen/module.go
```diff
@@ -819,7 +818,7 @@ func readPackageInfo(
 		// uniqueness of the provided package aliases. Note that the default
 		// package is "PackageName".
 		PackageAlias:          defaultAlias + "Static",
-		GeneratedPackageAlias: defaultAlias + "Generated",
+		GeneratedPackageAlias: defaultAlias + "generated",
```
ChuntaoLu: looking at the direct above and below, shouldn't we be consistent on all alias cases?

cl1337: only `GeneratedPackageAlias` is used in the header line `package <GeneratedPackageAlias>` according to templates, other alias are used in import package renaming
```golang
import (
    <alias> "path/to/package"
)
```

ChuntaoLu: That is correct, my point is since we are doing this, we should keep the package alias style consistent.

PR #342, opened 2018-02-27
fix optional assigning header val to request param

oliviazhang0809:
Change method.go to consider both self-defined type and primitives
Modify baz.call method to test case of int64 and uuid


codegen/method.go
```diff
@@ -675,9 +675,34 @@ func (ms *MethodSpec) setEndpointRequestHeaderFields(
...
 
-					statements.appendf("\trequestBody%s = ptr.String(%s)",
-						bodyIdentifier, variableName,
-					)
+					switch fieldThriftType {
+					case "string":
+						statements.appendf("\trequestBody%s = ptr.String(%s)",
+							bodyIdentifier, variableName,
+						)
+					case "int64":
```
cl1337: maybe `int32`

PR #345, opened 2018-03-02
Mock client fixture abstration for generated clients

ChuntaoLu:
This PR implements the mock client fixture abstraction for thrift based clients as described in detail [here](https://gist.github.com/ChuntaoLu/9d5dcf37376b855f64e63156f765d183).


codegen/gateway.go
```diff
@@ -115,12 +117,38 @@ type ClientClassConfig struct {
...
 // Dependencies lists all dependencies of a module
 type Dependencies struct {
 	Client []string `json:"client"`
 	//	Service []string `json:"service"`  // example extension
 }
 
+// Fixture specifies client fixture import path and all scenarios
+type Fixture struct {
+	ImportPath string              `json:"importPath"`
+	Scenarios  map[string][]string `json:"scenarios"`
```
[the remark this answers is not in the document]
ChuntaoLu: what's the benefits of defining a type for it?

codegen/module_system.go
```diff
@@ -443,6 +444,24 @@ func (g *HTTPClientGenerator) Generate(
 		files["module/dependencies.go"] = dependencies
 	}
 
+	if clientMeta.Fixture != nil {
```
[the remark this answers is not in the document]
ChuntaoLu: Maybe or maybe not, that's in the scope of fixture creation.

codegen/gateway.go
```diff
@@ -351,6 +326,46 @@ func newClientSpec(
...
+	if !ok || len(exposedMethods) == 0 {
+		return nil, errors.Errorf(
+			"No methods are exposed in client config: %s",
+			instance.JSONFileName,
+		)
+	}
+	cspec.ExposedMethods = make(map[string]string, len(exposedMethods))
+	reversed := make(map[string]string, len(exposedMethods))
+	for key, val := range exposedMethods {
+		cspec.ExposedMethods[key] = val.(string)
```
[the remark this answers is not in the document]
ChuntaoLu: good catch, done

codegen/module_system.go
```diff
@@ -600,6 +638,31 @@ func hasMethod(cspec *ClientSpec, thriftMethod string) bool {
...
+		return
+	}
+	mock, err = template.ExecTemplate(
+		"client_mock.tmpl",
+		data,
+		packageHelper,
+	)
+	if err != nil {
+		types = nil
+		return
```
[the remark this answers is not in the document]
ChuntaoLu: done

examples/example-gateway/clients/contacts/fixture/fixture.go
```diff
@@ -0,0 +1,23 @@
+package fixture
+
+import (
+	mc "github.com/uber/zanzibar/examples/example-gateway/build/clients/contacts/mock-client"
+	gen "github.com/uber/zanzibar/examples/example-gateway/build/gen-code/clients/contacts/contacts"
+)
+
+var saveContactsFixtures = map[string]*mc.SaveContactsFixture{
+	"success": {
```
jacobgreenleaf: Convert from `map[string]*mc.SaveContactsFixture` to struct defined in generated code but initialized in user code. This will reduce time to detect the error to build time (field is not present) rather than test time (null pointer exception).

ChuntaoLu: Great suggestion. Replaced the map with struct now.

codegen/templates/client_fixture_types.tmpl
```diff
@@ -0,0 +1,71 @@
...
+{{- end -}}
+{{- end -}}
+}
+
+{{range $methodName, $scenarios := $scenariosMap -}}
+{{$methodName := pascal $methodName -}}
+// {{$methodName}}Scenarios defines all fixture scenarios for {{$methodName}}
+type {{$methodName}}Scenarios struct {
+{{range $scenario := $scenarios -}}
+{{pascal $scenario}} *{{$methodName}}Fixture `scenario:"{{$scenario}}"`
```
jacobgreenleaf: I don't see that the tag is used anywhere. What is this for or possibly to be used for?

codegen/gateway.go
```diff
@@ -115,12 +117,38 @@ type ClientClassConfig struct {
...
+
 // Dependencies lists all dependencies of a module
 type Dependencies struct {
 	Client []string `json:"client"`
 	//	Service []string `json:"service"`  // example extension
 }
 
+// Fixture specifies client fixture import path and all scenarios
+type Fixture struct {
+	ImportPath string              `json:"importPath"`
```
jacobgreenleaf: `// ImportPath is the package where the user-defined Fixture global variable is contained. The Fixture object defines, for a given client, the standardized list of fixture scenarios for that client.`

codegen/gateway.go
jacobgreenleaf: `// Scenarios is a map from zanzibar's exposed method name to a list of user-defined fixture scenarios for a client.`

codegen/templates/client_fixture_types.tmpl
ChuntaoLu: This is intended read by people. It links the struct field with the string literal in `client-config.json`, one can define such scenario
```
{
    "scenarios": [
         "not-found"
     ]
}
```
where "not-found" would be converted to "NotFound" to qualify as a struct field, the tag points to the scenario in above json.

codegen/gateway.go
ChuntaoLu: done

codegen/gateway.go
ChuntaoLu: done

PR #352, opened 2018-03-09
Support post build hooks and generate mock clients with hook

ChuntaoLu:
This PR:
- adds support for post build generation hooks which are useful for work that relies on generated build;
- generates mock clients using post gen hook instead of bash script for code gen consistency and better error handling


scripts/generate.sh
```diff
@@ -10,8 +10,8 @@ start=$(cat .TMP_ZANZIBAR_TIMESTAMP_FILE.txt)
 go run codegen/runner/runner.go -config="$PREFIX/gateway.json"
 end=`date +%s`
 runtime=$((end-start))
-echo "Generated endpoints/clients : +$runtime"
+echo "Generated build : +$runtime"
 
-bash ./codegen/runner/post-steps.sh "$PREFIX" "$PREFIX/build"
+#bash ./codegen/runner/post-steps.sh "$PREFIX" "$PREFIX/build"
```
cl1337: mm ... remove comment?

PR #354, opened 2018-03-09
add open tracing to inbound calls

jakelarkn:
This PR starts an tracing span for inbound requests and propagates the span to downstream.
If the inbound requests carry spans, the spans started by Zanzibar will be `chidOf` upstream spans.


runtime/server_http_request.go
```diff
@@ -451,3 +477,8 @@ func (req *ServerHTTPRequest) UnmarshalBody(
 
 	return true
 }
+
+// GetContext returns the http request Context
+func (req *ServerHTTPRequest) GetContext() context.Context {
```
ChuntaoLu: This one liner function seems too simple to be necessary, the only call site is https://github.com/uber/zanzibar/pull/354/files#diff-2fd36df947abf02141a9e4d07dd5367dR834

runtime/server_http_request_test.go
```diff
@@ -780,3 +794,53 @@ func TestPeekBody(t *testing.T) {
...
+			bgateway.ActualGateway.Logger,
+			bgateway.ActualGateway.AllHostScope,
+			bgateway.ActualGateway.Tracer,
+			"foo", "foo",
+			func(
+				ctx context.Context,
+				req *zanzibar.ServerHTTPRequest,
+				res *zanzibar.ServerHTTPResponse,
+			) {
+				value, _, err := req.PeekBody("arg1", "b1", "c1")
```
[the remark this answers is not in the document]
jakelarkn: copypasta  artifact   my bad

runtime/server_http_request.go
```diff
@@ -93,6 +99,26 @@ func (req *ServerHTTPRequest) start() {
 
 	// emit metrics
 	req.metrics.Recvd.Inc(1)
+
+	if req.tracer != nil {
+		// TODO better operation name   normalized path  etc.
+		opName := fmt.Sprintf("%s %s.%s", req.Method, req.EndpointName, req.HandlerName)
+		carrier := opentracing.HTTPHeadersCarrier(req.httpRequest.Header)
+		spanContext, err := req.tracer.Extract(opentracing.HTTPHeaders, carrier)
+		var span opentracing.Span
```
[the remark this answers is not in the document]
jakelarkn: seems like the right thing to do.    also fireball may call edge  to retrieve response object etc.

runtime/server_http_request.go
jakelarkn: needed a way to check span out of package scope in test.  let me know if you have a way to improve

ChuntaoLu: Oh, I see, didn't notice it was in a different package. Makes sense.

runtime/server_http_request.go
```diff
@@ -93,6 +101,25 @@ func (req *ServerHTTPRequest) start() {
...
+		opName := fmt.Sprintf("%s.%s", req.EndpointName, req.HandlerName)
+		urlTag := opentracing.Tag{Key: "URL", Value: req.URL}
+		MethodTag := opentracing.Tag{Key: "Method", Value: req.Method}
+		carrier := opentracing.HTTPHeadersCarrier(req.httpRequest.Header)
+		spanContext, err := req.tracer.Extract(opentracing.HTTPHeaders, carrier)
+		var span opentracing.Span
+		if err != nil {
+			if err != opentracing.ErrSpanContextNotFound {
+				/* coverage ignore next line */
+				req.Logger.Warn("Error Extracting Trace Headers", zap.Error(err))
```
cl1337: tbh I don't think this warn log matters much here

PR #356, opened 2018-03-12
Unify and parallelize client mock/fixture generation for all clients

ChuntaoLu:
This PR:
- generates mock client with fixtures for custom clients;
- unifies client mock/fixture generation for all types of clients into post build hooks;
- flushes generated code to files at the end of all mock/fixture generation to avoid dirty git-status in case of errors;
- parallelizes mock/fixture generation, on my local machine, the overall build time for example-gateway is reduced to ~10s, compared with previously ~16s;
- bumps supported go version to 1.9 in order to use `sync.Map`.


codegen/post_gen_hooks.go
```diff
@@ -23,102 +23,204 @@ package codegen
...
-						"Error writing to file %q",
-						s.path,
+						"Error generating service mock_init.go for %s",
+						instance.InstanceName,
 					)
+					return
 				}
-				if err := FormatGoFile(s.path); err != nil {
-					return err
+				files[filepath.Join(genDir, "mock_init.go")] = mockInit
```
cl1337: should probably use https://golang.org/doc/go1.9#sync-map

ChuntaoLu: good catch, replaced map with sync.Map

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
+	bin, err := NewMockgenBin(h, t)
 	if err != nil {
 		return nil, errors.Wrap(err, "error building mockgen binary")
 	}
 
 	return func(instances map[string][]*ModuleInstance) error {
 		fmt.Println("Generating client mocks:")
 		mockCount := len(instances["client"])
+		ec := make(chan error, mockCount)
+		var files sync.Map
```
jacobgreenleaf: I assume you are bumping the Go version to 1.9 for access to this? If so can you adjust the pull request to note this for historical reasons?

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
+				files.Store(filepath.Join(genDir, "mock_client.go"), mock)
+
+				// generate fixture types and augmented mock client
+				f := mc.Config.Fixture
+				if f != nil && f.Scenarios != nil {
+					types, augMock, err := bin.AugmentMockWithFixture(importPath, f, clientInterface)
+					if err != nil {
+						ec <- errors.Wrapf(
+							err,
+							"error generating mock client with fixtures for client %q",
```
jacobgreenleaf: nit: `"error generating fixture for client %q"`

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
+							instance.InstanceName,
+						)
+						return
+					}
+
+					files.Store(filepath.Join(genDir, "types.go"), types)
+					files.Store(filepath.Join(genDir, "mock_client_with_fixture.go"), augMock)
+				}
+
+				fmt.Printf(
```
jacobgreenleaf: nit: Format strings like this can easily become out of sync with the parameters. Prefer to use a strongly typed function so that adding a new parameter can enforce it's updated in all places.

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
-				instance.ClassName,
-				instance.InstanceName,
-				path.Join(path.Base(h.CodeGenTargetPath()), instance.Directory, "mock-client"),
-				i+1, mockCount,
-			)
+			}(instance, i)
+		}
+		wg.Wait()
+
+		select {
```
jacobgreenleaf: If there is an issue with multiple mocks, then I would need to run this repeatedly to figure out these errors. If you are going to run all of them unconditionally, then you should show the user the errors from all of them. Collect the errors and return a single error that accumulates all of the errors, i.e. `err = errors.New("Encountered %d errors when generating mocks: ", N, strings.Join("\n", err1, err2, ..., errN))`

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
-			}
-			mockService, err := t.ExecTemplate("service_mock.tmpl", instance, h)
-			if err != nil {
-				return errors.Wrapf(
-					err,
-					"Error generating service mock_service.go for %s",
-					instance.InstanceName,
-				)
-			}
+			go func(instance *ModuleInstance, i int) {
```
jacobgreenleaf: Same comment on concurrency structure as above.

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
+						instance.InstanceName,
+					)
+					return
+				}
+
+				buildDir := h.CodeGenTargetPath()
+				genDir := filepath.Join(buildDir, instance.Directory, "mock-client")
+
+				importPath := instance.PackageInfo.GeneratedPackagePath
+				if instance.ClassType == "custom" {
```
jacobgreenleaf: Can you move all literal references to `"custom"` to const? This can be a separate diff.

codegen/reflect_interface.go
```diff
@@ -0,0 +1,150 @@
...
+
+// ReflectInterface uses reflection to obtain interface information identified by symbols in given import path
+// projRoot is the root dir where mockgen is installed as a vendor package
+func ReflectInterface(projRoot string, importPath string, symbols []string) (*model.Package, error) {
+	// We use TempDir instead of TempFile so we can control the filename.
+	tmpDir, err := ioutil.TempDir(projRoot, "gomock_reflect_")
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = os.RemoveAll(tmpDir) }()
```
jacobgreenleaf: nit: you can just do `defer os.RemoveAll(tmpDir)` - no need to assign the variable, and since `tmpDir` is constant there is no need to defer an anonymous function.

codegen/reflect_interface.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+type reflectData struct {
+	ImportPath string
+	Symbols    []string
+}
+
+// ReflectInterface uses reflection to obtain interface information identified by symbols in given import path
+// projRoot is the root dir where mockgen is installed as a vendor package
+func ReflectInterface(projRoot string, importPath string, symbols []string) (*model.Package, error) {
```
jacobgreenleaf: nit: `projRoot` and `importPath` are both `string`, you should use more concise `projRoot, importPath string, symbols []string`

nit: `symbols` is a `[]string` but only ever passed a single parameter. Should this be just a `string`?

codegen/module_system.go
```diff
@@ -197,21 +197,22 @@ type ClientTestFixture struct {
 	ClientResHeaders FixtureBlob     `json:"clientResHeaders"`
 }
 
-// NewDefaultModuleSystemWithMockHook reates a fresh instance of the default zanzibar
+// NewDefaultModuleSystemWithMockHook creates a fresh instance of the default zanzibar
 // module system (clients, endpoints, services) with a post build hook to generate client and service mocks
 func NewDefaultModuleSystemWithMockHook(
```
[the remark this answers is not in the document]
ChuntaoLu: This function calls `NewDefaultModuleSystem` with predefined hooks, and `NewDefaultModuleSystem` already takes variadic arguments, https://github.com/uber/zanzibar/blob/master/codegen/module_system.go#L223. I think the overlap you see is an illusion due to github folding, it's actually the body of `NewDefaultModuleSystem`.

codegen/mockgen.go
```diff
@@ -81,42 +94,140 @@ func NewMockgenBin(projRoot string) (*MockgenBin, error) {
...
+
+	exposedMethods := make([]*model.Method, numMethods, numMethods)
+	for i, methodName := range sortedMethods {
+		exposedMethods[i] = methodsMap[methodName]
+	}
+
+	imports := pkg.Imports()
+
+	// Sort keys to make import alias generation predictable
+	sortedPaths := make([]string, len(imports), len(imports))
```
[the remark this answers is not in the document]
ChuntaoLu: The intention is to make imports ordering deterministic, but I guess this is not needed because we do gofmt anyway.

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
 		return nil, errors.Wrap(err, "error building mockgen binary")
 	}
 
 	return func(instances map[string][]*ModuleInstance) error {
 		fmt.Println("Generating client mocks:")
 		mockCount := len(instances["client"])
+		ec := make(chan error, mockCount)
+		var files sync.Map
+		var wg sync.WaitGroup
+		wg.Add(mockCount)
```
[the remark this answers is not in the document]
ChuntaoLu: Yes, that's something I initially thought about and planned to dig in the future. Now I think I can probably get away with running the subprocess just once.

codegen/post_gen_hooks.go
```diff
@@ -23,114 +23,229 @@ package codegen
...
+					return
+				}
+
+				buildDir := h.CodeGenTargetPath()
+				genDir := filepath.Join(buildDir, instance.Directory, "mock-client")
+
+				importPath := instance.PackageInfo.GeneratedPackagePath
+				if instance.ClassType == "custom" {
+					importPath = mc.Config.CustomImportPath
+					if importPath == "" {
```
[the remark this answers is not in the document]
ChuntaoLu: hmm, the logic here is to decide where the interface is defined: custom import path if type is custom otherwise generated import path. Not sure how validators can avoid the check.

codegen/post_gen_hooks.go
ChuntaoLu: The bump is to get access to `sync.Map`, and I will put that in the PR description.

codegen/post_gen_hooks.go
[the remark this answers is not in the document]
ChuntaoLu: Yes, it is already validated here https://github.com/uber/zanzibar/blob/master/codegen/gateway.go#L260, I will just delete this check.

codegen/reflect_interface.go
ChuntaoLu: I tried to make it generic to extract multiple interfaces if needed, the current use cases happen to only need to extract one interface.

codegen/mockgen.go
```diff
@@ -81,42 +94,140 @@ func NewMockgenBin(projRoot string) (*MockgenBin, error) {
...
+		j++
+	}
+	sort.Strings(sortedPaths)
+
+	pkgPathToAlias := make(map[string]string, len(imports))
+	usedAliases := make(map[string]bool, len(imports))
+	for _, pkgPath := range sortedPaths {
+		base := camelCase(path.Base(pkgPath))
+		pkgAlias := base
+		i := 0
```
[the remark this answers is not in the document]
ChuntaoLu: `i` is used here https://github.com/uber/zanzibar/pull/356/files/c52883403eb9ec9e4bbf7be6a7ba0978b4a9bc30#diff-175ff617d327d55fef63de9b0feaf894R175

codegen/reflect_interface.go
ChuntaoLu: That's what I initially did, but the `errcheck` linter doesn't like unhandled error :(

codegen/post_gen_hooks.go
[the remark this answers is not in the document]
jacobgreenleaf: I meant something like https://godoc.org/gopkg.in/validator.v2?

codegen/post_gen_hooks.go
```diff
@@ -21,116 +21,295 @@
...
+								instance.InstanceName,
+							)
+							return
+						}
+
+						files.Store(filepath.Join(genDir, "types.go"), types)
+						files.Store(filepath.Join(genDir, "mock_client_with_fixture.go"), augMock)
+					}
+
+					printGenLine(
```
jacobgreenleaf: nit: Is fmt.Printf thread safe? If it's a wrapper around C `printf` then the answer is probably no, in which case you'll get mixed output from threads sometimes.

codegen/post_gen_hooks.go
```diff
@@ -21,116 +21,295 @@
...
+					)
+					atomic.AddInt32(&idx, 1)
+				}
+			}()
+		}
+
+		for _, instance := range clientInstances {
+			ic <- instance
+		}
+		close(ic)
```
jacobgreenleaf: If you close the channel before all of the items are dequeued won't your workers abort early?

examples/example-gateway/clients/quux/quux.go
```diff
@@ -21,12 +21,15 @@
...
-	module "github.com/uber/zanzibar/examples/example-gateway/build/clients/quux/module"
+	"github.com/uber/zanzibar/examples/example-gateway/build/clients/quux/module"
+	"github.com/uber/zanzibar/examples/example-gateway/build/gen-code/clients/foo/base/base"
 )
 
 // Client is a custom client that does nothing yet
 type Client interface {
-	Echo(string) string
+	EchoString(string) string
+	//EchoMessage(m *base.Message) *base.Message
```
jacobgreenleaf: Can you add a reference to tracking task for why these are commented out in another comment?

codegen/post_gen_hooks.go
ChuntaoLu: `ic` is not buffered, sending to it is blocking, channel won't be closed until all items are sent.

examples/example-gateway/clients/quux/quux.go
ChuntaoLu: I was experimenting, should not need to be commented out, I will update

PR #357, opened 2018-03-13
Add debug logging for server making downstream request

oliviazhang0809:
Dynamic logging is enabled, adding logging logics
```
#
#   benchmark output add-logging
#

    Latency    35.64ms   32.59ms 462.39ms   64.67%
    Req/Sec     1.06k   151.91     1.87k    72.03%
  381387 requests in 30.04s, 41.83MB read
  Socket errors: connect 0, read 243, write 0, timeout 0
Requests/sec:  12696.06
Transfer/sec:      1.39MB

#
#   benchmark output master
#

    Latency    34.63ms   31.96ms 358.55ms   65.01%
    Req/Sec     1.09k   168.92     2.21k    74.94%
  389272 requests in 30.04s, 42.69MB read
  Socket errors: connect 0, read 243, write 0, timeout 3
  Non-2xx or 3xx responses: 3
Requests/sec:  12958.42
Transfer/sec:      1.42MB
```

### DIFF
--- a/codegen/templates/endpoint.tmpl
+++ b/codegen/templates/endpoint.tmpl
@@ -129,8 +129,11 @@ func (h *{{$handlerName}}) HandleRequest(
 	// log endpoint request to downstream services
 	zfields := []zapcore.Field{
 		zap.String("endpoint", h.endpoint.EndpointName),
-		zap.String("body", fmt.Sprintf("%#v", requestBody)),
 	}
+	{{if ne .RequestType ""}}
+	zfields = append(zfields, zap.String("body", fmt.Sprintf("%#v", requestBody)))
+	{{- end -}}
+
 	{{if (ne (len $reqHeaderMapKeys) 0) }}
 	var headerOk bool
 	var headerValue string
--- a/codegen/templates/endpoint.tmpl
+++ b/codegen/templates/endpoint.tmpl
@@ -126,6 +126,23 @@ func (h *{{$handlerName}}) HandleRequest(
 	{{$line}}
 	{{end}}
 
+	// log endpoint request to downstream services
+	zfields := []zapcore.Field{
+		zap.String("endpoint", h.endpoint.EndpointName),
+		zap.String("body", fmt.Sprintf("%#v", requestBody)),
+	}
+	{{if (ne (len $reqHeaderMapKeys) 0) }}
+	var headerOk bool
+	var headerValue string
+	{{- end -}}
+	{{range $i, $k := $reqHeaderMapKeys}}
+	headerValue, headerOk = req.Header.Get("{{$k}}")
+	if headerOk {
+		zfields = append(zfields, zap.String("{{$k}}", headerValue))
+	}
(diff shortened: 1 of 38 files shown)


codegen/test_data/endpoints/bar_bar_method_argnotstruct.gogen
```diff
@@ -75,6 +77,14 @@ func (h *BarArgNotStructHandler) HandleRequest(
 		return
 	}
 
+	// log endpoint request to downstream services
+	zfields := []zapcore.Field{
+		zap.String("endpoint", h.endpoint.EndpointName),
+	}
+
+	zfields = append(zfields, zap.String("body", fmt.Sprintf("%#v", requestBody)))
```
cl1337: do we have special stringer for requestBody?

codegen/templates/endpoint.tmpl
```diff
@@ -126,6 +126,26 @@ func (h *{{$handlerName}}) HandleRequest(
 	{{$line}}
 	{{end}}
 
+	// log endpoint request to downstream services
+	zfields := []zapcore.Field{
+		zap.String("endpoint", h.endpoint.EndpointName),
+	}
+	{{if ne .RequestType ""}}
+	zfields = append(zfields, zap.String("body", fmt.Sprintf("%#v", requestBody)))
```
ChuntaoLu: `fmt.Sprintf("%#v", requestBody)` is reflection based and therefore could be expensive depending on the `requestBody`, ideally serialization should not happen until `Debug` is called, otherwise it's work done that is of no use. [zap.Object](https://godoc.org/go.uber.org/zap#Object) is better as marshaling is done lazily.

ChuntaoLu: The benchmark looks okay, we can think about `zap.Object` in a later time. Could you put a comment line like `// TODO: potential perf, use zap.Object lazy serialization` in here?

Also instead of  `fmt.Sprintf("%#v", requestBody))` we should just do `requestBody.String()`, the struct type generated by thrfitrw already implements `Stringer` interface.

ChuntaoLu:
I guess I have to take it back on `requestBody.String()`, it is quite expensive compared to `fmt.Printf("%#v", requestBody)`:
```
#
#   benchmark output add-logging
#

    Latency    59.10ms   57.24ms 575.87ms   74.23%
    Req/Sec   665.13    112.21     1.19k    71.03%
  238697 requests in 30.06s, 26.18MB read
  Socket errors: connect 0, read 235, write 0, timeout 0
Requests/sec:   7940.00
Transfer/sec:      0.87MB

#
#   benchmark output master
#

    Latency    45.84ms   43.55ms 542.43ms   65.37%
    Req/Sec   824.91    124.18     1.44k    70.92%
  295910 requests in 30.04s, 32.45MB read
  Socket errors: connect 0, read 282, write 0, timeout 0
Requests/sec:   9849.17
Transfer/sec:      1.08MB
```

PR #360, opened 2018-03-15
Cache method mocks and better method name to get mock clients

ChuntaoLu:
This PR:
- caches the method mocks so that subsequent calls do not create new wrapper objects;
- Renames MockClientNodes() to MockClients() on mockService for better ergonomic;

### DIFF
--- a/codegen/templates/augmented_mock.tmpl
+++ b/codegen/templates/augmented_mock.tmpl
@@ -12,6 +12,12 @@ import (
 type MockClientWithFixture struct {
 	*MockClient
 	fixture *ClientFixture
+
+	{{range $method := $methods}}
+	{{$methodName := $method.Name -}}
+	{{$methodMockType := printf "%sMock" $methodName -}}
+	{{camel $methodMockType}} *{{$methodMockType}}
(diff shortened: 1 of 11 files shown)


codegen/templates/augmented_mock.tmpl
```diff
@@ -41,10 +48,13 @@ type {{$methodMockType}} struct {
 {{$methodMockMethod := printf "Expect%s" $methodName -}}
 // {{$methodMockMethod}} returns an object that allows the caller to choose expected scenario for {{$methodName}}
 func (m *MockClientWithFixture) {{$methodMockMethod}}() *{{$methodMockType}} {
-	return &{{$methodMockType}}{
-		scenarios:  m.fixture.{{$methodName}},
-		mockClient: m.MockClient,
+	if m.{{$methodMockField}} == nil {
```
cl1337: do we ever have to concern about # of fixtures we load in memory?

ChuntaoLu: We might need to think about it when we have tons of unique fixtures.

PR #362, opened 2018-03-16
bump go version in travis

cl1337:
remove support for 1.8 since some new data structure introduced in 1.9 only (syncMap)


cl1337:
:D

Issue #361, opened 2018-03-16
improve header transform and header propagate


cl1337:
currently zanzibar operates headers in following 3 ways:

1: in endpoint thrift schema configure `zanzibar.http.reqHeaders = "header1,header2"`
this schema annotation will invoke `checkHeaders` when endpoint handling an incoming request, and consider all specified endpoint mandatory required, if any of the headers not presented, it will fail the incoming request

2: in endpoint config json configure header propagate, which introduced in [#346](https://github.com/uber/zanzibar/pull/346), that allows headers to be used as request field to downstream client call

3: in endpoint config json configure `{req|res}HeaderMap`, to extract headers from incoming endpoint request, and (if exist) propagate to client request headers with mapped naming

a couple of problems:

1: no single source of truth for headers 
2: no schema for header map and necessary required / optional annotation
3: loose check around headers

going to address this with a header schema dot thrift files that
1: imported from include of endpoint schema
2: annotated in endpoint annotations
3: used as source of truth for headers check / validation

cl1337:
3.1415926

PR #363, opened 2018-03-16
#361 - 3 refactor header config loading

cl1337:
dedup

### DIFF
--- a/codegen/gateway.go
+++ b/codegen/gateway.go
@@ -28,9 +28,8 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
-	"strings"
-
 	"sort"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/uber/zanzibar/runtime"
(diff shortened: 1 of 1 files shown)


codegen/gateway.go
```diff
@@ -578,6 +577,35 @@ func testFixtures(endpointConfigObj map[string]interface{}) (map[string]*Endpoin
 	return ret, err
 }
 
+func loadHeadersFromConfig(endpointConfigObj map[string]interface{}, key string) ([]string, map[string]string, error) {
+	headers, ok := endpointConfigObj[key]
+	if !ok {
+		return nil, nil, errors.Errorf("Unable to parse %s", key)
```
ChuntaoLu: nit: the convention is have the first letter in lower case in the error message so that when the error is wrapped the message is like, "error blah blah: unable to parse ..." instead of "error blah blah: Unable to parse ...". Also "%q" is arguably a better formatter as it makes the key stand out.

codegen/gateway.go
```diff
@@ -578,6 +577,35 @@ func testFixtures(endpointConfigObj map[string]interface{}) (map[string]*Endpoin
...
+	}
+	retHeadersMap := make(map[string]string)
+	castMap := headers.(map[string]interface{})
+	for key, value := range castMap {
+		switch value := value.(type) {
+		case string:
+			retHeadersMap[key] = value
+		default:
+			return nil, nil, errors.Errorf(
+				"Unable to parse string %s in headers %s",
```
ChuntaoLu: same nit here

codegen/gateway.go
cl1337: noted, I addressed this style in https://github.com/uber/zanzibar/pull/366

also added TODO to capture `endpointConfigObj` cast as interface issue there

cl1337:
closing this in favor of https://github.com/uber/zanzibar/pull/366

Issue #364, opened 2018-03-16
correct moduleSpec SetDownstream signature


cl1337:
moduleSpec [setDownstream](https://github.com/uber/zanzibar/blob/master/codegen/gateway.go#L890) 

the fact that endpointSpec passes so many its internal properties to ModuleSpec for setdownstream seems a failure of isolation between class`ModuleSpec` and class `EndpointSpec`

if `ModuleSpec` depend so much on `EndpointSpec` to do set downstream job (and seems like its only job), we'd rather just pass `EndpointSpec` as its anchor param 

again a proper fix will require some serious look at these 2 classes function / responsibility

PR #365, opened 2018-03-16
#361 - 3.1 clean up setDownstream func sig

cl1337:
addresses https://github.com/uber/zanzibar/issues/364

### DIFF
--- a/codegen/service_test.go
+++ b/codegen/service_test.go
@@ -61,7 +61,13 @@ func TestExceptionValidation(t *testing.T) {
 	}
 	assert.NoError(t, err)
 	method.ExceptionsIndex = map[string]codegen.ExceptionSpec{"test": {}}
-	err = m.SetDownstream("Bar", "argNotStruct", clientSpec, "argNotStruct", nil, nil, nil, pkgHelper)
+	e := &codegen.EndpointSpec{
+		ThriftServiceName: "Bar",
+		ThriftMethodName:  "argNotStruct",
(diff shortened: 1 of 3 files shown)


codegen/service_test.go
```diff
@@ -61,7 +61,13 @@ func TestExceptionValidation(t *testing.T) {
 	}
 	assert.NoError(t, err)
 	method.ExceptionsIndex = map[string]codegen.ExceptionSpec{"test": {}}
-	err = m.SetDownstream("Bar", "argNotStruct", clientSpec, "argNotStruct", nil, nil, nil, pkgHelper)
+	e := &codegen.EndpointSpec{
+		ThriftServiceName: "Bar",
+		ThriftMethodName:  "argNotStruct",
+		ClientSpec:        clientSpec,
+		ClientMethod:      "argNotStruct",
```
ChuntaoLu: The fact that you only need to specify these fields to run the tests says you don't really need an endpointSpec.

cl1337: this test took a shortcut to cover exception validation case 

-> `	method.ExceptionsIndex = map[string]codegen.ExceptionSpec{"test": {}}`

look at how it construct all the service, method, clientSpec required for a test, pretty sad

if we still have backend/ , where a place for bad, un-generate-able thrift to live, we can write a better integration test with all its required component resolved by zanzibar modules, and we can just have 

`m.SetDownstream(e, pkgHelper)` here instead of `m.SetDownstream(<10 different positional params>)`

PR #367, opened 2018-03-20
#361 - 3.141 support primary types in header propagate

cl1337:
add support to parse header strings to primary types including bool, int and float

in header propagate, we need to assign headers to client call parameters other than
string, we will parse and attempt to resolve appropriate type before assign

if assign failed, we no op on the assignment, following could happen:

1: if client call takes optional param, this is legit case
2: if client call takes required param, this case might be caught by type-converter later
3: if client call is required and type-converter doesn't capture this case, this will cause 0 value client request without fix from https://github.com/uber/zanzibar/pull/366, in https://github.com/uber/zanzibar/pull/366, we will actively check and fail request without headers used to transform/propagate


codegen/header_propagate.go
```diff
@@ -75,27 +76,58 @@ func (hp *HeaderPropagator) Propagate(
...
+		ret       = []string{}
+		typeParse string
+		typeCast  string
+		assignVal = "key"
+	)
+	switch gotype {
+	case "bool":
+		typeParse = "strconv.ParseBool(key)"
+		assignVal = "v"
+	case "int32":
```
ChuntaoLu: what about `int16`?

codegen/header_propagate.go
```diff
@@ -75,27 +76,58 @@ func (hp *HeaderPropagator) Propagate(
...
+		typeCast = "val := int32(v)\n"
+	case "int64":
+		typeParse = "strconv.ParseInt(key, 10, 64)"
+		assignVal = "v"
+	case "float64":
+		typeParse = "strconv.ParseFloat(key, 64)"
+		assignVal = "v"
+	case "string":
+	default:
+		typeCast = "val := " + gotype + "(key)\n"
```
ChuntaoLu: This can potentially cause runtime panics, e.g. `val := int16("42")`

examples/example-gateway/idl/clients/baz/baz.thrift
```diff
@@ -14,6 +14,16 @@ struct BazRequest {
     3: required i32 i3
 }
 
+struct transHeaderType {
```
ChuntaoLu: Shouldn't we cover all [thrift base types](https://thrift.apache.org/docs/types) ?

codegen/header_propagate.go
cl1337: good catch

examples/example-gateway/idl/clients/baz/baz.thrift
cl1337: I'll raise panic to cover all the types, not all types apply to headers

Issue #371, opened 2018-04-05
Unable to use list<X> in a HTTP query parameter results in panic


jacobgreenleaf:
When decoding URI parameters for HTTP requests, zanzibar lacks an encoder and decoder for list Thrift types, e.g. `list<string>`. Currently this causes a panic due to a failing type switch when trying to create the encoder or decoder. 

A standard encoding of composite types is undefined. [RFC 3986 § 3.4](https://tools.ietf.org/html/rfc3986#section-3.4) merely says: 

> [...] query components are often used to carry identifying information in the form of "key=value" pairs

but defines no convention for delimiters or encodings of complex types like maps or lists, leaving it up to the particular HTTP daemon to decide. 

Golang `net` standard library [will encode a list](https://golang.org/pkg/net/url/#Values) e.g. `foo=[a, b]` as `foo=a&foo=b`:

```
	v := url.Values{}
	v.Set("name", "Ava")
	v.Add("friend", "Jess")
	v.Add("friend", "Sarah")
	v.Add("friend", "Zoe")
	// v.Encode() == "name=Ava&friend=Jess&friend=Sarah&friend=Zoe"
```

Decoding is done by using the map access since `url.Values` is typedef of a `map[string][]string`. 

Zanzibar has support for struct types in URL parameters by using a dotted notation: `a.b=foo` would map `b` as a string property on struct `a`. This creates ambiguity if we add support for list types; `a.b.c=foo&a.b.c=bar` can be decoded to either of these structs:

As `c` is list of string: 
```
struct A {
    1: required B b
}

struct B {
    1: required list<string> c
}
```

Or as `b` is list of struct:

```
struct A {
    1: required list<B> b
}

struct B {
    1: required string c
}
```

To avoid this ambiguity, we should restrict lists to scalar types.

PR #372, opened 2018-04-05
Add test case for HTTP query parameters of type list

jacobgreenleaf:
Creates a test case for #371 and removes the panic on generate. Instead the value will be nil until we add support for this, at which time we can change the test case to assert the `Tags` are decoded correctly.


examples/example-gateway/endpoints/bar/bar_call_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	respBody := &endpointsBarBar.BarResponse{}
+	err = json.Unmarshal(respBytes, respBody)
+	assert.NoError(t, err)
+
+	// t.Logf("Resp body: %s", respBytes)
+	// t.Logf("Resp body: %+v", respBody)
+
+	assert.Equal(t, "foobar", respBody.StringField)
+}
```
ChuntaoLu: > Creates a failing test for #371

Does this test fail?

jacobgreenleaf: I will rephrase. It creates a passing test case that locks in the a incorrect behavior. The new behavior is that query string parameters that are for list-types are dropped instead of being decoded into the thrift struct. This assertion is done via the mock call:

```
+	expectedRequest := &clientsBarBar.Bar_ArgWithQueryParams_Args{
+		Name:     "kot",
+		UserUUID: &userUUID,
+		// This is currently broken due to https://github.com/uber/zanzibar/issues/371
+		// Tags:     []string{"foo", "bar"},
+		Tags: []string{},
+	}
```

We will swap `Tags` from `[]string{}` to `[]string{"foo", "bar"}` to assert the correct behavior.

ChuntaoLu: Got it.

ChuntaoLu:
I see what's done here, but I am not sure if the behavior is desirable. Instead of panic, now code gen succeeds, but the generated code is not behaving as the user expects?

jacobgreenleaf:
> I see what's done here, but I am not sure if the behavior is desirable. Instead of panic, now code gen succeeds, but the generated code is not behaving as the user expects?

Yes. It's a bit non-trivial to implement the code that decodes into a struct but I will hold off on this diff until that is completed.

PR #374, opened 2018-04-07
improve debug logging

cl1337:
avoid debug logging if level is not debug
log marshall-ed bytes for req and res

### DIFF
--- a/codegen/templates/endpoint.tmpl
+++ b/codegen/templates/endpoint.tmpl
@@ -20,6 +20,7 @@ package {{$instance.PackageInfo.PackageName}}
 
 import (
 	"context"
+	"encoding/json"
 	"io/ioutil"
 	"net/http"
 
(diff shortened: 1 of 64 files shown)


codegen/templates/workflow.tmpl
```diff
@@ -30,6 +30,7 @@ import (
 	{{- end}}
 
 	module "{{$instance.PackageInfo.ModulePackagePath}}"
+	"go.uber.org/zap"
```
ChuntaoLu: nit: move it to https://github.com/uber/zanzibar/pull/374/files#diff-f66420b147eeb736922ebdd1ea765a69R20

cl1337: mm ... I was following this: https://github.com/golang/go/wiki/CodeReviewComments#imports

`go.uber.org` considered 3rd party package in this import

codegen/templates/endpoint.tmpl
```diff
@@ -160,6 +156,27 @@ func (h *{{$handlerName}}) HandleRequest(
...
 	{{else}}
 	response, cliRespHeaders, err := w.Handle(ctx, req.Header, &requestBody)
+
+	// log downstream response to endpoint
+	if ce := req.Logger.Check(zapcore.DebugLevel, "stub"); ce != nil {
+		zfields := []zapcore.Field{
+			zap.String("endpoint", h.endpoint.EndpointName),
+		}
+		{{- if ne .ResponseType ""}}
+		if response != nil {
```
cl1337: I'm going to drop this check

some response returned as primary type (like string, int), this will break when compare nil

directly marshal nil struct seems acceptable 
```
	bytes, err := json.Marshal(x)
    // bytes -> `null`
    // err -> `<nil>`
```

PR #375, opened 2018-04-12
Generate workflow integration test helpers

ChuntaoLu:
This PR addes a post-gen-hook to generate workflow test helper for endpoints of custom workflow.
An example test case is added for =Contacts.saveContacts=.


codegen/post_gen_hooks.go
```diff
@@ -330,3 +341,162 @@ func writeAndFormat(path string, data []byte) error {
...
+func WorkflowMockGenHook(h *PackageHelper, t *Template) PostGenHook {
+	return func(instances map[string][]*ModuleInstance) error {
+		errChanSize := 0
+		shouldGenMap := map[*ModuleInstance][]*EndpointSpec{}
+		for _, instance := range instances["endpoint"] {
+			endpointSpecs := instance.genSpec.([]*EndpointSpec)
+			for _, endpointSpec := range endpointSpecs {
+				if endpointSpec.WorkflowType == "custom" {
+					shouldGenMap[instance] = endpointSpecs
+					errChanSize += len(endpointSpecs)
```
cl1337: this `+=` makes me think there're more custom endpoint specs and we'll accumulate, but it follows with a `break` :)

codegen/post_gen_hooks.go
```diff
@@ -330,3 +341,162 @@ func writeAndFormat(path string, data []byte) error {
...
+				if err != nil {
+					ec <- errors.Wrapf(
+						err,
+						"Error generating mock endpoint %s",
+						instance.InstanceName,
+					)
+					return
+				}
+
+				var subWg sync.WaitGroup
```
cl1337: I wish go has something like `async.auto`

codegen/post_gen_hooks.go
```diff
@@ -330,3 +341,162 @@ func writeAndFormat(path string, data []byte) error {
...
+					int(atomic.LoadInt32(&idx)), genEndpointCount,
+				)
+				atomic.AddInt32(&idx, 1)
+			}(instance, endpointSpecs)
+		}
+		wg.Wait()
+
+		select {
+		case err := <-ec:
+			close(ec)
```
cl1337: nit: maybe close(ec) after loop through the channels

codegen/module_system.go
```diff
@@ -219,9 +219,14 @@ func NewDefaultModuleSystemWithMockHook(
 		return nil, errors.Wrap(err, "error creating client mock gen hook")
 	}
 
+	workflowMockGenHook := WorkflowMockGenHook(h, t)
 	serviceMockGenHook := ServiceMockGenHook(h, t)
 
-	allHooks := append([]PostGenHook{clientMockGenHook, serviceMockGenHook}, hooks...)
+	allHooks := append([]PostGenHook{
```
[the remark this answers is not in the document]
ChuntaoLu: Order matters for the console-printed generation output, other than that these hooks don't have dependencies on each other.

codegen/post_gen_hooks.go
ChuntaoLu: hmm, I don't get it

codegen/post_gen_hooks.go
ChuntaoLu: Iol, I am glad that go does not have async.auto :p

codegen/post_gen_hooks.go
ChuntaoLu: We can not loop through the channel here before closing it, otherwise the channel is still open, receiving from the open channel will block.

codegen/post_gen_hooks.go
```diff
@@ -330,3 +341,162 @@ func writeAndFormat(path string, data []byte) error {
 	}
 	return FormatGoFile(path)
 }
+
+// WorkflowMockGenHook returns a PostGenHook to generate endpoint workflow mocks
+func WorkflowMockGenHook(h *PackageHelper, t *Template) PostGenHook {
+	return func(instances map[string][]*ModuleInstance) error {
+		errChanSize := 0
+		shouldGenMap := map[*ModuleInstance][]*EndpointSpec{}
+		for _, instance := range instances["endpoint"] {
```
jacobgreenleaf: It seems that this code only generates the initializer for `endpoint` type. What about user defined types like `task`?

ChuntaoLu: User defined types would require corresponding user defined hooks to generate the initializers.

PR #378, opened 2018-04-19
add trace key to response

cl1337:
propagate trace id from request to response header

so that we'll be able to track endpoint request -> client request -> client response -> endpoint response in our log with uniq trace key

### DIFF
--- a/codegen/package.go
+++ b/codegen/package.go
@@ -55,6 +55,8 @@ type PackageHelper struct {
 	middlewareSpecs map[string]*MiddlewareSpec
 	// Use staging client when this header is set as "true"
 	stagingReqHeader string
+	// traceKey is the key for uniq trace id that identifies request / response pair
+	traceKey string
 }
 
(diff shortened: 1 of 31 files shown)


examples/example-gateway/gateway.json
```diff
@@ -12,6 +12,6 @@
 	"copyrightHeader": "./copyright_header.txt",
 
 	"annotationPrefix": "zanzibar",
-
+	"trace-key": "x-trace-id",
```
ChuntaoLu: Inconsistent style, `traceKey`

codegen/package.go
```diff
@@ -69,6 +71,7 @@ func NewPackageHelper(
 	copyrightHeader string,
 	annotationPrefix string,
 	stagingReqHeader string,
+	traceKey string,
```
ChuntaoLu: No objection here, but feels like `PackageHelper` is out of control, we need to think about how to prevent it from growing endlessly.

PR #379, opened 2018-04-24
Add default header forward

jyang0:
Add
`
"defaultReqHeaderMap": {
	"x-trace-id": "x-trace-id"
},
"defaultResHeaderMap": {
	"x-trace-id": "x-trace-id"
}
`
field to gateway.json, similar to `reqHeaderMap` and `resHeaderMap` in endpoint configs

### DIFF
--- a/codegen/package.go
+++ b/codegen/package.go
@@ -57,6 +57,10 @@ type PackageHelper struct {
 	stagingReqHeader string
 	// traceKey is the key for uniq trace id that identifies request / response pair
 	traceKey string
+	// A map of default headers that all endpoints will forward to their clients
+	defaultReqHeaderMap map[string]string
+	// A map of default headers that all clients will forward back to their endpoints
+	defaultResHeaderMap map[string]string
 }
 
 // NewPackageHelper creates a package helper.
@@ -72,6 +76,8 @@ func NewPackageHelper(
 	annotationPrefix string,
 	stagingReqHeader string,
 	traceKey string,
+	defaultReqHeaderMap map[string]string,
+	defaultResHeaderMap map[string]string,
 ) (*PackageHelper, error) {
 	absConfigRoot, err := filepath.Abs(configRoot)
 	if err != nil {
@@ -101,6 +107,8 @@ func NewPackageHelper(
 		annotationPrefix:    annotationPrefix,
 		stagingReqHeader:    stagingReqHeader,
 		traceKey:            traceKey,
+		defaultReqHeaderMap: defaultReqHeaderMap,
+		defaultResHeaderMap: defaultResHeaderMap,
 	}
 	return p, nil
 }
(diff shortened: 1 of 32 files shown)


examples/example-gateway/gateway.json
```diff
@@ -13,5 +13,12 @@
 
 	"annotationPrefix": "zanzibar",
 	"traceKey": "x-trace-id",
-	"genMock": true
+	"genMock": true,
+
+	"defaultReqHeaderMap": {
+		"x-trace-id": "x-trace-id"
```
ChuntaoLu: Should we have this config as a map? To me, this is saying "**transform** endpoint header `x-trace-id` to client header `x-trace-id` for every endpoint". I think a global setting of forwarding is more proper:

```
"defaultReqHeadersToForward": [
    "x-trace-id"
]
```
This says "forward endpoint header `x-trace-id` to client header for each endpoint".

I think it is important to express the config semantic clearly. It's fine that header forwarding is achieved/implemented via transformation with the same header key, but the config semantic should be clear that it's forwarding.

examples/example-gateway/build/endpoints/bar/workflow/bar_bar_method_argnotstruct.go
```diff
@@ -74,8 +74,12 @@ func (w barArgNotStructWorkflow) Handle(
 	if ok {
 		clientHeaders["X-Zanzibar-Use-Staging"] = h
 	}
+	h, ok = reqHeaders.Get("x-trace-id")
```
ChuntaoLu: the casing is not consistent with existing header casing in this file.

codegen/package.go
```diff
@@ -72,6 +76,8 @@ func NewPackageHelper(
 	annotationPrefix string,
 	stagingReqHeader string,
 	traceKey string,
+	defaultReqHeaderMap map[string]string,
```
ChuntaoLu: @cl1337 fyi, another example of `packageHelper` growing

examples/example-gateway/build/endpoints/bar/workflow/bar_bar_method_argnotstruct.go
cl1337: although under the hood we camelcase the header passed in, but good to stay case consistency

examples/example-gateway/gateway.json
jyang0: I do agree that a global setting of forwarding is more proper. I just thought that I might as well try to follow the style of the endpoint configs, which have a `reqHeaderMap` and `resHeaderMap`, so it'd be consistent to have a `default[Req/Res]HeaderMap` that behaves the same way. What do you think?

jyang0: @cl1337

examples/example-gateway/build/endpoints/bar/workflow/bar_bar_method_argnotstruct.go
jyang0: I can update the codegen to format/canonicalize whatever headers are in gateway.json, instead of preserving the original casing. Although right now the library is case insensitive with `Get` and canonicalizes all headers that are set

PR #380, opened 2018-04-25
#361 - 3.141592 fix header propagate on void request

cl1337:
cover the case of using header propagate without request

```
response method() throws (
    1 Exception exp (zanzibar.http.statusCode = "401")
)
```

client request will be created in 
```
clientRequest := propagateHeaders{{title .Name}}ClientRequests(nil, reqHeaders)
```


codegen/method.go
```diff
@@ -919,6 +919,11 @@ func (ms *MethodSpec) setHeaderPropagator(
 		downstreamMethod.RequestType,
 		"{",
 	)
+
+	hp.append("if in == nil {")
+	hp.append(fmt.Sprintf(`in = %s{}`, strings.Replace(downstreamMethod.RequestType, "*", "&", 1)))
```
cl1337: client request is always 2 cases, either structs or null, no primary type here

PR #385, opened 2018-05-11
fix nil response headers triggers NPE in zanzibar

herainman:
Issue: 
Returning nil from an endpoint instead of zanzibar.Header (usually a map cast into zanzibar.TChannelHeaders) triggers a null pointer exception in zanzibar-generated code.

Fix:
add nil check for headers

### DIFF
--- a/codegen/templates/tchannel_endpoint.tmpl
+++ b/codegen/templates/tchannel_endpoint.tmpl
@@ -93,8 +93,10 @@ func (h *{{$handlerName}}) Handle(
 	{{end}}
 
 	resHeaders := map[string]string{}
-	for _, key := range wfResHeaders.Keys() {
-		resHeaders[key], _ = wfResHeaders.Get(key)
+	if wfResHeaders != nil {
+		for _, key := range wfResHeaders.Keys() {
+			resHeaders[key], _ = wfResHeaders.Get(key)
+		}
 	}
 
 	{{if eq (len .Exceptions) 0 -}}
@@ -134,11 +136,13 @@ func (h *{{$handlerName}}) Handle(
 	{{end}}
 
 	{{- if .ResHeaders}}
-	if err := wfResHeaders.Ensure({{.ResHeaders | printf "%#v" }}, h.endpoint.Logger); err != nil {
-		return false, nil, nil, errors.Wrapf(
-			err, "%s.%s (%s) missing response headers",
-			h.endpoint.EndpointID, h.endpoint.HandlerID, h.endpoint.Method,
-		)
+	if wfResHeaders != nil {
+		if err := wfResHeaders.Ensure({{.ResHeaders | printf "%#v" }}, h.endpoint.Logger); err != nil {
+			return false, nil, nil, errors.Wrapf(
+				err, "%s.%s (%s) missing response headers",
+				h.endpoint.EndpointID, h.endpoint.HandlerID, h.endpoint.Method,
+			)
+		}
 	}
 	{{- end}}
 
--- a/codegen/templates/tchannel_endpoint.tmpl
+++ b/codegen/templates/tchannel_endpoint.tmpl
@@ -137,7 +137,11 @@ func (h *{{$handlerName}}) Handle(
 
 	{{- if .ResHeaders}}
 	if wfResHeaders == nil {
-		return false, nil, nil, errors.Errorf(
+		return false, nil, nil, errors.Wrapf(
+			errors.Errorf(
+				"Missing mandatory headers: %s",
+				strings.Join({{.ResHeaders | printf "%#v" }}, ", "),
+			),
 			"%s.%s (%s) missing response headers",
 			h.endpoint.EndpointID, h.endpoint.HandlerID, h.endpoint.Method,
 		)
--- a/codegen/templates/tchannel_endpoint.tmpl
+++ b/codegen/templates/tchannel_endpoint.tmpl
@@ -136,13 +136,18 @@ func (h *{{$handlerName}}) Handle(
 	{{end}}
 
 	{{- if .ResHeaders}}
-	if wfResHeaders != nil {
-		if err := wfResHeaders.Ensure({{.ResHeaders | printf "%#v" }}, h.endpoint.Logger); err != nil {
-			return false, nil, nil, errors.Wrapf(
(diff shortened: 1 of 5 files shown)


ChuntaoLu:
I am not sure this is the right fix, because `nil` is valid response header.
 
It's not clear where the NPE happens, can you give an concrete example? Maybe you can write a test where the workflow return nil response header and see where the NPE is and see how we should fix it.

examples/example-gateway/endpoints/tchannel/baz/baz_call.go
```diff
@@ -68,7 +68,7 @@ func (w Workflow) Handle(
 	}
 
 	clientRespHeaders, err := w.Clients.Baz.Call(ctx, clientReqHeaders, clientReq)
-	respHeaders := zanzibar.ServerTChannelHeader(clientRespHeaders)
```
herainman: It seems there is  a chance that when casting to zanzibar.ServerTChannelHeader, it might return a nil, which would be triggers NPE at [here](https://github.com/uber/zanzibar/blob/5274bc56f715972efb39b3665a5104d6e2a34fe6/examples/example-gateway/build/endpoints/tchannel/baz/baz_simpleservice_method_call_tchannel.go#L91)

jacobgreenleaf:
Here is a stack trace for the panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x14f51a5]

goroutine 45 [running]:
code.uber.internal/marketplace/fulfillment.git/build/endpoints/tincup-exchange-getpairwiseexchange.(*TincupExchangeServiceGetPairwiseExchangeRateHandler).Handle(0xc42057ada0, 0x17159e0, 0xc42051df80, 0xc42051de00, 0xc4208f4230, 0x1a6ad40, 0x16bcf00, 0xc420915e58, 0x135b332, 0xc4200c81a0, ...)
        /Users/jacobg/gocode/src/code.uber.internal/marketplace/fulfillment.git/build/endpoints/tincup-exchange-getpairwiseexchange/tincup-exchange-getpairwiseexchange_tincupexchangeservice_method_getpairwiseexchangerate_tchannel.go:68 +0x1b5
code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime.(*tchannelInboundCall).handle(0xc4206a4c60, 0x17159e0, 0xc42051df80, 0xc4208f4230, 0x0, 0x0, 0x0, 0x0)
        /Users/jacobg/gocode/src/code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime/tchannel_server.go:369 +0xb9
code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime.(*TChannelRouter).handle(0xc420342780, 0x17159e0, 0xc42051df80, 0xc4206a4c60, 0x0, 0x0)
        /Users/jacobg/gocode/src/code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime/tchannel_server.go:211 +0x1e7
code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime.(*TChannelRouter).Handle.func1(0xc4206a4c00, 0xc420342780, 0x17159e0, 0xc42051dcb0, 0xc4206a4c60)
        /Users/jacobg/gocode/src/code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime/tchannel_server.go:178 +0x49
created by code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime.(*TChannelRouter).Handle
        /Users/jacobg/gocode/src/code.uber.internal/marketplace/fulfillment.git/vendor/github.com/uber/zanzibar/runtime/tchannel_server.go:178 +0x254
FAIL    code.uber.internal/marketplace/fulfillment.git/endpoints/tincup-exchange-getpairwiseexchange    0.052s
```

examples/example-gateway/endpoints/tchannel/baz/baz_call.go
ChuntaoLu: If the link is right, `wfResHeaders, err := workflow.Handle(ctx, wfReqHeaders, &req)` doesn't seem like where a NPE would happen.

herainman: Sorry here is the correct https://github.com/uber/zanzibar/blob/5274bc56f715972efb39b3665a5104d6e2a34fe6/examples/example-gateway/build/endpoints/tchannel/baz/baz_simpleservice_method_call_tchannel.go#L94

ChuntaoLu: Why would that line cause NPE? Calling `Keys()` on a nil header is okay, see https://play.golang.org/p/HsRL1EzYZ0e

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -134,11 +136,13 @@ func (h *{{$handlerName}}) Handle(
 	{{end}}
 
 	{{- if .ResHeaders}}
-	if err := wfResHeaders.Ensure({{.ResHeaders | printf "%#v" }}, h.endpoint.Logger); err != nil {
-		return false, nil, nil, errors.Wrapf(
-			err, "%s.%s (%s) missing response headers",
-			h.endpoint.EndpointID, h.endpoint.HandlerID, h.endpoint.Method,
-		)
+	if wfResHeaders != nil {
```
ChuntaoLu: This is not correct. `Ensure` checks if the header contain certain keys and return error if not. If the header is nil, it should return error instead of skipping the check.

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -134,6 +136,13 @@ func (h *{{$handlerName}}) Handle(
 	{{end}}
 
 	{{- if .ResHeaders}}
+	if wfResHeaders == nil {
+		return false, nil, nil, errors.Errorf(
+			"%s.%s (%s) missing response headers",
```
ChuntaoLu: We should inform what headers are missing in the error message.

herainman:
seems pushed the wrong version. will make a new PR soon...

PR #386, opened 2018-05-14
test default param value

oliviazhang0809:


examples/example-gateway/build/gen-code/endpoints/bar/bar/types.go
```diff
@@ -156,12 +157,17 @@ func (v *BarRequest) ToWire() (wire.Value, error) {
 		err    error
 	)
 
-	w, err = wire.NewValueString(v.StringField), error(nil)
-	if err != nil {
-		return w, err
+	if v.StringField == nil {
+		v.StringField = ptr.String("abc")
```
oliviazhang0809: @cl1337 seems the default value is supported in zanzibar.

PR #390, opened 2018-05-16
Add static value to header propogator

oliviazhang0809:
To support case 
```
{
  "value": "header",
  "to": "specific_path_to_param"
}
```

### DIFF
--- a/codegen/header_propagate.go
+++ b/codegen/header_propagate.go
@@ -36,12 +36,25 @@ type HeaderPropagator struct {
 	Helper PackageNameResolver
 }
 
+const (
+	staticValue = "static"
+	fromValue   = "from"
+)
+
 // NewHeaderPropagator returns an instance of HeaderPropagator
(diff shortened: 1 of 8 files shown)


codegen/header_propagate.go
```diff
@@ -70,6 +70,17 @@ func (hp *HeaderPropagator) Propagate(
 				val.QualifiedName, field.Name)
 		}
 
+		if val.StaticValue != "" {
```
cl1337: can we do type switch here to emphasis 2 different approach (static value vs from field)

oliviazhang0809: sure

cl1337:
as discussed offline, all header propagate, header schema, type converting is migrated to our internal project, this case seems general enough, I'm open to either here / where you actually use it

ChuntaoLu:
Setting a default value for a downstream request has no logical relationship with header propagation. This does not seem to be a proper solution for the problem it tries to solve.

PR #392, opened 2018-05-18
Serialize module tree in build folder

jacobgreenleaf:
Many of the build steps require as input the module tree. For example, to build the module struct we must know the direct dependencies. 

Currently the dependency tree is resolved once at the beginning of the build process and passed to the generators in memory either as a side-effect (through the `ResolvedDependencies` and `RecursiveDependencies` properties). As a prerequisite of incremental builds we need to be able to execute the entire build process for a single module instance, but by passing the dependency tree only through memory we would need to re-resolve the tree each time. 

This diff stores the module tree onto disk during the build process so we can resolve it once, read it from disk in the individual build steps, and remove it once the build is done.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -43,6 +43,8 @@ const (
 	SingleModule moduleClassType = iota
 	// MultiModule defines a module class type with multiple nested directories
 	MultiModule moduleClassType = iota
+
+	serializedModuleTreePath = "zanzibar.tree"
 )
 
 const configSuffix = "-config.json"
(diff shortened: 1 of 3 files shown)


codegen/module.go
```diff
@@ -868,6 +870,15 @@ func (system *ModuleSystem) GenerateBuild(
 		return nil, err
 	}
 
+	serializedModules, err := json.Marshal(resolvedModules)
```
ChuntaoLu: Instead of JSON, maybe Gob?

jacobgreenleaf: neat +1

jacobgreenleaf: On second thought, maybe we should leave it open for bash scripting using `jq`; it's not being transferred over the network yet.

PR #394, opened 2018-05-25
fix type converting with typedef-ed map key struct

cl1337:
bug fixing for type converting map struct with type-def key, example see clients/baz.thrift and endpoints/baz.thrift

there're 2 things went wrong:
1: client thrift and endpoint thrift might define their own `typedef string UUID`, and use this type as map key (hashable), but go does not know these two typedefs in endpoint thrift and client thrift are pointing to the same base type under the hood, hence we need to type convert one to another

2: there's a prefix set bug that misplace `fromprefix` to `toprefix` when converting struct 

this fix in type converer does not deal with transform, i.e. there will be bug as long as we're not using same thrift for endpoint thrift and client thrift, hence not falling into the type convert deprecate plan


jacobgreenleaf:
Can you add a test in `codegen/type_converter_test.go` for `map<UUID, string>`?

cl1337:
will do, although the test failed due to some missing generated file that git status != 0 ;)

codegen/type_converter.go
```diff
@@ -469,14 +469,27 @@ func (c *TypeConverter) genConverterForMap(
...
 		}
 
+		toFieldKeyID := keyID
+
+		toTypeKeyName := keyType
+		fromTypeKeyName, err := c.getGoTypeName(fromFieldType.KeySpec)
+		if err != nil {
+			return err
+		}
+		if fromTypeKeyName != toTypeKeyName {
```
jacobgreenleaf: If the from-type is not convertible to the to-type this will result in a compile error in the generated code. For example, if from-type is `[]byte` and to-type is `int`. I'm not sure if there is an easy way to solve this but maybe you can think about it as a potential issue. Ideally we would point the user directly to the Thrift line that is the issue rather than in the generated code.

jacobgreenleaf:
I will approve when build is passing. Looks good.

codegen/type_converter.go
cl1337: that's correct, ideally key type should be exactly the same, theoretically user can set any hash-able type as key. This will fail at build time.

Raising issues for thrift files is unfortunately not an option for my use case: asking users to use `client.UUID` instead of endpoint `UUID` 3 layers down in a map struct inside endpoint thrift schema ...

On production the key types are 95% string, and 5% typedef string, we could limit this with downcast-ed basetype compare?

jacobgreenleaf: That is fine with me.

cl1337: so in this case specifically, map key is restricted to strings / typedef strings here:
https://github.com/uber/zanzibar/pull/394/files#diff-b01b2823b821664c3242667b138861d1R405

unconvertible case can never happen

if we remove this restriction in the future we can add 

```
t1 := reflect.TypeOf(mapkeytype1)
t2 := reflect.TypeOf(mapkeytyp2)
return t1.ConvertibleTo(t2) && t2.ConvertibleTo(t1)
```

cl1337:
cover complaining 99.75 != 100 :)
screen shot 2018-05-25 at 3 06 46 pm
although I have 100 cover on my change 🤷‍♂️

PR #396, opened 2018-05-31
expose pending respose bytes/obj/statusCode

r-patali:
there are use cases outside of zanzibar to have access to pending
response information


cl1337:
master failure should be unblocked by https://github.com/uber/zanzibar/pull/397

runtime/server_http_response.go
```diff
@@ -280,3 +280,9 @@ func (res *ServerHTTPResponse) writeBytes(bytes []byte) {
 		)
 	}
 }
+
+// GetPendingResponse lets you read the pending body bytes, obj and status code
+// which isn't sent back yet.
+func (res *ServerHTTPResponse) GetPendingResponse() ([]byte, interface{}, int) {
+	return res.pendingBodyBytes, res.pendingBodyObj, res.pendingStatusCode
```
cl1337: why we're returning `res.pendingBodyObj` ?

r-patali: in case you wish to deserialize the bytes to its object for manipulation.

cl1337: we should probably not allow field manipulation, in case some pointer value inside bodyObj set accidentally

r-patali: The whole idea of exposing these is so that we can modify the response from the middleware, isn't it?

cl1337: for push case, my 2 cents:

1: we expose bytes for push to access such payload, so that push middleware can send such payload as it is to push client

2: it is the case that push endpoint returns nil body with 204, but this could be achieved by other ways

3: allowing mutate for push object, for **A** case, without restriction to such case, may lead to unexpected abuse

what do you think?

r-patali: yes, it makes sense. I think you convinced me that we should start restrictive and open up other fields if/when necessary.

PR #397, opened 2018-06-07
fix master

cl1337:
zanzibar master is broken: https://travis-ci.org/uber/zanzibar/builds/379957528

some lint lib we pull adhoc during build run bumped with new rules, that fails package inside zanzibar

build them as dependency instead and pin to a version that not breaking anything

a bump of such linting lib may come with style fixes in zanzibar

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,6 @@ lint: check-licence eclint-check
 	@go get golang.org/x/lint/golint
 	@$(foreach dir,$(PKGS),golint $(dir) 2>&1 | $(FILTER_LINT) | tee -a lint.log;)
 	@echo "Checking errcheck..."
-	@go get github.com/kisielk/gotool
 	@go run vendor/github.com/kisielk/errchek/main.go $(PKGS) 2>&1 | $(FILTER_LINT) | tee -a lint.log
 	@echo "Checking staticcheck..."
 	@go build -o vendor/honnef.co/go/tools/cmd/staticcheck/staticcheck   vendor/honnef.co/go/tools/cmd/staticcheck/staticcheck.go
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -78,11 +78,11 @@ lint: check-licence eclint-check
 	@go get golang.org/x/lint/golint
 	@$(foreach dir,$(PKGS),golint $(dir) 2>&1 | $(FILTER_LINT) | tee -a lint.log;)
 	@echo "Checking errcheck..."
-	@go get github.com/kisielk/errcheck
-	@errcheck $(PKGS) 2>&1 | $(FILTER_LINT) | tee -a lint.log
+	@go get github.com/kisielk/gotool
```
jacobgreenleaf: Why do you need to `go get` here if you're just using `go run vendor/.../errchek`?

cl1337: actually it's `vendor/honnef.co/go/tools/cmd/staticcheck/staticcheck` has dependency on `github.com/kisielk/gotool`, not  `vendor/github.com/kisielk/errchek`

I can put this in `glide.yaml`

glide.yaml
```diff
@@ -52,3 +52,8 @@ import:
   version: master
 - package: github.com/pborman/uuid
   version: ^1.1.0
+- package: honnef.co/go/tools/cmd/staticcheck
+  version: 8ed405e85c65fb38745a8eafe01ee9590523f172
```
jacobgreenleaf: Add an issue that describes why you did this. It can be annoying to figure out why deps are pinned at particular hashes without it.

cl1337: good advice

cl1337: https://github.com/uber/zanzibar/issues/398

PR #399, opened 2018-06-21
Fix mock workflow initilization

ChuntaoLu:
This PR fixes the dependency initilization and assigments in mock workflow constructor.


ChuntaoLu:
@jacobgreenleaf Do we want to keep the `repository` module class and `kappa` test case?

jacobgreenleaf:
Ideally we would change the example gateway to define `repository` and `kappa` so that it's not an exported module type from zanzibar default module system, but you can delete it from this diff for now.

PR #404, opened 2018-07-06
Disable request and response body logging.

gjtrowbridge:
Disable logging the full request and response bodies on every request.


gjtrowbridge:
If necessary we can put together something that reads this in from the config in the future, but it sounds like for now it's preferable to disable this logging by default.

PR #405, opened 2018-07-06
Swap deputy subchannel to scAlt at tchannel_client.go

herainman:
- The value of request header "x-deputy-forwarded" is sub-channel name. 

- If a coming request has header "x-deputy-forwarded", then swap scAlt with deputy sub-channel
at tchannel_client.go, then call scAlt channel instead of sc


herainman:
Abondon this PR

PR #406, opened 2018-07-07
Better sidecar routing support

ChuntaoLu:
This PR:
- adds routing headers for http clients;
- sets routing delegate head for tchannel clients.


runtime/tchannel_client.go
```diff
@@ -193,6 +193,10 @@ func (c *TChannelClient) call(
 	if c.routingKey != nil {
 		ctxBuilder.SetRoutingKey(*c.routingKey)
 	}
+	rd, ok := ctx.Value("rd").(string)
```
jacobgreenleaf: You should expose a function like `WithRoutingDelegate(Context, string) Context` so that the implementation (i.e. string `"rd"`) is hidden to callers and more semantic.

runtime/tchannel_client.go
```diff
@@ -193,6 +193,10 @@ func (c *TChannelClient) call(
...
+	if ok {
+		ctxBuilder.SetRoutingDelegate(rd)
```
jacobgreenleaf: It would probably be a bit cleaner if instead of passing this parameter through the context it was passed as an optional parameter, e.g. [`yarpc.CallOption`](https://godoc.org/go.uber.org/yarpc#CallOption)

runtime/tchannel_client.go
ChuntaoLu: Good point, it's done now.

runtime/tchannel_client.go
ChuntaoLu: That's not something easy to do at this moment, client call signature is `someMethod(ctx context.Context, reqHeaders map[string]string, args *SomeArgType)`.

PR #407, opened 2018-07-10
Add deputy hostport to subtchannel based on value of x-deputy-forwarded per request

herainman:
If request had header of x-deputy-forwarded, then get the value of hostport, then add it to peer at the outbound tchannel per request


herainman:
abandon this and move to https://github.com/uber/zanzibar/pull/408

PR #408, opened 2018-07-11
Add deputy hostport to tchanel client

herainman:
This commits will send a request to a dedicated hostport. If inbound request had header of **x-deputy-forwarded**, tchannel client would get hostport from it, then add it to peer list at the outbound tchannel, the inbound request with header of **x-deputy-forwarded** would go to the dedicated peer destination. This is thread safe commits since production request and deputy request would go to their hostports separately .

### DIFF
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -60,6 +60,7 @@ type TChannelClient struct {
 	ch                *tchannel.Channel
 	sc                *tchannel.SubChannel
 	scAlt             *tchannel.SubChannel
+	selectedPeers     map[string]struct{}
 	serviceName       string
 	ClientID          string
 	methodNames       map[string]string
@@ -115,6 +116,7 @@ func NewTChannelClient(
 	client := &TChannelClient{
 		ch:                ch,
 		sc:                ch.GetSubChannel(opt.ServiceName),
+		selectedPeers:     map[string]struct{}{},
 		serviceName:       opt.ServiceName,
 		ClientID:          opt.ClientID,
 		methodNames:       opt.MethodNames,
@@ -150,7 +152,7 @@ func (c *TChannelClient) Call(
 		metrics:       c.metrics[serviceMethod],
 	}
 
-	return c.call(ctx, call, reqHeaders, req, resp, false)
+	return c.call(ctx, call, reqHeaders, req, resp, false, "")
 }
 
 // CallThruAltChannel makes a RPC call using a configured alternate channel
@@ -171,7 +173,29 @@ func (c *TChannelClient) CallThruAltChannel(
 		metrics:       c.metrics[serviceMethod],
 	}
 
-	return c.call(ctx, call, reqHeaders, req, resp, true)
+	return c.call(ctx, call, reqHeaders, req, resp, true, "")
+}
+
+// CallToHostPort makes a RPC call using a dedicated hostport
+func (c *TChannelClient) CallToHostPort(
+	ctx context.Context,
+	thriftService, methodName, hostPort string,
+	reqHeaders map[string]string,
+	req, resp RWTStruct,
+	useAltSubchannel bool,
+) (success bool, resHeaders map[string]string, err error) {
+	serviceMethod := thriftService + "::" + methodName
+
+	call := &tchannelOutboundCall{
+		client:        c,
+		methodName:    c.methodNames[serviceMethod],
+		serviceMethod: serviceMethod,
+		reqHeaders:    reqHeaders,
+		logger:        c.Loggers[serviceMethod],
+		metrics:       c.metrics[serviceMethod],
+	}
+
+	return c.call(ctx, call, reqHeaders, req, resp, useAltSubchannel, hostPort)
 }
 
 func (c *TChannelClient) call(
@@ -180,6 +204,7 @@ func (c *TChannelClient) call(
 	reqHeaders map[string]string,
 	req, resp RWTStruct,
 	useAltSubchannel bool,
+	hostport string,
 ) (success bool, resHeaders map[string]string, err error) {
 	defer func() { call.finish(err) }()
 	call.start()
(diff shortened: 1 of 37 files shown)


herainman:
this is a thread safe version for the reason that deputy request is using peer.beginCall, production traffic if using subchannel.begincall without calling preselected peers which contains deputy hostport. Please review, thank you.

runtime/tchannel_client.go
```diff
@@ -207,10 +232,33 @@ func (c *TChannelClient) call(
...
+			c.selectedPeers[hostport] = struct{}{}
+			p := sc.Peers().Add(hostport)
+			call.call, cerr = p.BeginCall(ctx, sc.ServiceName(), call.serviceMethod, &tchannel.CallOptions{
+				Format:       tchannel.Thrift,
+				RequestState: rs,
+			})
+
+			defer func() {
+				sc.Peers().Remove(hostport)
+				delete(c.selectedPeers, hostport)
```
jacobgreenleaf: I don't see how this is thread safe. Can you please add a test that proves it using the race detector? Specifically, what if a request to a client is happening at the same time as this deferred function? Won't the `delete(...)` conflict with the assignment `c.selectedPeers[hostport] = ...`?

herainman: Yes, you are right it is not fully thread safe but it is thread safe for scenario that the call with deputy hostport won't be in the production peer list. E.g. a thread of request to deputy peer parking between line 244 and 245, a thread of request to production peer still has pre-seleced peers of c.selectedPeers[**hostport**], but this deputy peer won't be in production peer list for deputy scenario.

To be fully thread safe, a request go to production needs to use peer.begincall instead of subchannel.begincall since the later one would use peer score to decide which peer the request goes to. However, picking up a peer from subchannel.peerlist would violate load-balancing except there is only one peer(127.0.0.1:5437) in the peerlist.

herainman: Add a RWLock for deputy request only. Please see comments inline

herainman:
verified E2E working with tincup-exchange

runtime/tchannel_client.go
```diff
@@ -211,10 +238,45 @@ func (c *TChannelClient) call(
...
+				}
+
+				c.mutex.Unlock()
+			}()
+		} else {
+			if rs.SelectedPeers == nil {
+				rs.SelectedPeers = map[string]struct{}{}
+			}
+
+			for k, v := range c.selectedPeers {
```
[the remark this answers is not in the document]
herainman: is there any other way that I could avoid locking here? Would readlock block all the reading threads at here which would affect performance?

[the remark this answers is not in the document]
jacobgreenleaf: Read locks commute (you can take a read while another goroutine is also holding a read lock), but write locks do not and write locks also block concurrent read locks.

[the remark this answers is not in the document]
herainman: I see. Thank you.
Readlock added

examples/example-gateway/build/endpoints/tchannel/baz/baz_simpleservice_method_call_tchannel.go
```diff
@@ -150,6 +158,51 @@ func (h *SimpleServiceCallHandler) Handle(
...
+
+	timeoutPerAttempt := time.Millisecond * time.Duration(
+		h.Deps.Default.Config.MustGetInt("tchannel.deputy.timeoutPerAttempt"),
+	)
+
+	methodNames := map[string]string{
+		"SimpleService::Call": "Call",
+	}
+
+	h.Deps.Default.Channel.GetSubChannel(serviceName, tchannel.Isolated)
```
ChuntaoLu: the subchannel here has a peer list that is designated for the subchannel, we should just add the hostport as peer to the peer list.

runtime/tchannel_client.go
```diff
@@ -171,7 +175,29 @@ func (c *TChannelClient) CallThruAltChannel(
 		metrics:       c.metrics[serviceMethod],
 	}
 
-	return c.call(ctx, call, reqHeaders, req, resp, true)
+	return c.call(ctx, call, reqHeaders, req, resp, true, "")
+}
+
+// CallToHostPort makes a RPC call using a dedicated hostport
+func (c *TChannelClient) CallToHostPort(
```
ChuntaoLu: This is unnecessary, the hostport should just be added as a peer when the client is created. No need to modify the `call` function either.

examples/example-gateway/build/endpoints/tchannel/baz/baz_simpleservice_method_call_tchannel.go
```diff
@@ -96,6 +98,12 @@ func (h *SimpleServiceCallHandler) Handle(
 			h.endpoint.EndpointID, h.endpoint.HandlerID, h.endpoint.Method,
 		)
 	}
+
+	if hostPort, ok := reqHeaders["x-deputy-forwarded"]; ok {
```
ChuntaoLu: nit: `ok && hostPost != ""`

examples/example-gateway/build/endpoints/tchannel/baz/baz_simpleservice_method_call_tchannel.go
herainman: Once the peer is added to peer list to the subchannel, if the subchannel.beginCall, it would use round robin to select peer (prod and deputy) based on the peer score to make a outbound call. Hence the prod traffic has a possibility to send request to deputy.

runtime/tchannel_client.go
herainman: client is created at init time with 1 channel for inbound request and 2 subchannel for outbound request. one of the subchannel is used for making request to prod, another one is altsubchannel for making request to staging. peer list only has 127.0.0.1:5437 at the begining

examples/example-gateway/build/endpoints/tchannel/baz/baz_simpleservice_method_call_tchannel.go
ChuntaoLu: For an isolated subchannel, the round robin is among the subchannel's peerlist, not top channel's peerlist.

herainman: Yes, I confirm it is round robin around subchannel's peerlist.

PR #412, opened 2018-07-18
add error details for tchannel system error

cl1337:
we want to capture error details for our client requests, tchannel and http, for better monitoring and debug-ability. Sample metrics we want to capture is request timeout

for http clients, we already emitting metrics for 418 and 504

for tchannel clients, error details are categorized as `systemError`  -> added some known type that we care about to emit extra metrics
- if tchannel error is either of such (like timeout), the `system-error` counter will be tagged with error metrics name. Such subscope is init-ed during the client initialization 

I'm also adding one liner for per host bootstrap since a spike of such metrics was used as indicator of massive instance crash

### DIFF
--- a/runtime/call_metrics.go
+++ b/runtime/call_metrics.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/pkg/errors"
 	"github.com/uber-go/tally"
 	"github.com/uber/tchannel-go"
 )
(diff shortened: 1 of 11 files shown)


runtime/tchannel_client.go
```diff
@@ -268,7 +268,12 @@ func (c *tchannelOutboundCall) finish(err error) {
 
 	// emit metrics
 	if err != nil {
-		c.metrics.SystemErrors.Inc(1)
+		tcause := tchannel.GetSystemErrorCode(errors.Cause(err))
+		if terr, ok := c.metrics.SystemErrors[byte(tcause)]; ok {
```
jacobgreenleaf: nit: you should define `c.metrics.SystemErrors` as a typedef like

```
type SystemErrorMap map[tchannel.SystemErrorCode]tally.Counter
```

Then you can define helper function like

```
func (m SystemErrorMap) Incr(cause tchannel.SystemErrorCode) {
    counter, ok := m[cause]
    if !ok {
        m[0x00].Inc(1)
        return
    }
    counter.Inc(1)
}
```

And this code changes to simply

```
c.metrics.SystemErrors.Incr(tcause)
```

And you can delete the duplicate code from below.

runtime/tchannel_logger.go
```diff
@@ -127,11 +127,19 @@ func (l TChannelLogger) WithFields(fields ...tchannel.LogField) tchannel.Logger
...
 		} else {
 			logger.Error(msg, zap.Error(err))
 		}
 	}
 }
+
+// IsTimeout return true if error caused by timeout
+func IsTimeout(err error) bool {
+	cause := errors.Cause(err)
+	return cause == context.Canceled ||
```
jacobgreenleaf: Context can be cancelled for reasons other than timeout.

cl1337: wrapping previous logic to a function, it probably makes sense under its scope cc @ravirajj 

I can make this more explicit

runtime/tchannel_client.go
cl1337: Do you mind we do a style clean also for http status metrics at once in a separate diff? following the same style here but I think your comment is solid

I'm trying not to over-DRY here, the metrics logic between `tchannel_client.go` and `tchannel_server.go` seems a bit repetitive, but the metric behavior for endpoint & client might not always be the same (tchannel server might not need to emit timeout at all)

jacobgreenleaf: OK

runtime/call_metrics.go
```diff
@@ -132,6 +141,49 @@ var knownStatusCodes = map[int]bool{
...
+	}
+	return ret
+}
+
+// IncrErr will increase according error counter
+func (sem systemErrorMap) IncrErr(err error, count int64) {
+	errCause := tchannel.GetSystemErrorCode(errors.Cause(err))
+	counter, ok := sem[errCause]
+	if !ok {
+		counter = sem[0x00]
```
ChuntaoLu: instead of `0x00`, here the key should be `tchannel.ErrCodeInvalid`.

test/health_test.go
```diff
@@ -211,12 +211,12 @@ func TestRuntimeMetrics(t *testing.T) {
...
 
 	// Expect 9 runtime metrics + 1 logged metric
-	numMetrics := 10
+	numMetrics := 11
 	cgateway.MetricsWaitGroup.Add(numMetrics)
 	cgateway.MetricsWaitGroup.Wait()
 
 	metrics := cgateway.M3Service.GetMetrics()
-	assert.Equal(t, numMetrics, len(metrics), "expected 10 metrics")
+	assert.Equal(t, numMetrics, len(metrics), "expected 10TestCallMetrics metrics")
```
ChuntaoLu: typo?

cl1337: 🤦‍♂️

runtime/call_metrics.go
cl1337: `ErrCodeInvalid` commented as `"never should be used"`, show my respect for the author's will ;)

PR #413, opened 2018-07-20
Create sub-channel with dedicated hostport for deputy request

herainman:
- Create a subchannel with dedicated hostport per deputy request.
- Remove callToHostPort method from PR#408 at Tchannel_Client


ChuntaoLu:
@herainman lgtm, let's squash the commits and give it some more meaningful message.

PR #414, opened 2018-07-20
bug fixes for type converter

cl1337:
```
when converting (*float)&Param

go formatting formats such line wrong if it's not (*float)(&(Param))
```


codegen/type_converter_test.go
```diff
@@ -229,7 +229,7 @@ func TestConvertBoolsOptionalToRequired(t *testing.T) {
 
 	assert.NoError(t, err)
 	assertPrettyEqual(t, trim(`
-		out.One = (*bool)&(in.One)
```
ChuntaoLu: Ha, this is fun, bitwise and.

cl1337: ;)

PR #417, opened 2018-07-27
inject jaeger span before send http client request

herainman:
inject jaeger span before send http client request, it is the [same](https://github.com/uber/zanzibar/blob/master/runtime/tchannel_client.go#L227) as we injecting span before sending out tchannel client request

### DIFF
--- a/runtime/client_http_request.go
+++ b/runtime/client_http_request.go
@@ -158,7 +158,8 @@ func (req *ClientHTTPRequest) Do(
 	methodTag := opentracing.Tag{Key: "Method", Value: req.httpReq.Method}
 	span, ctx := opentracing.StartSpanFromContext(ctx, opName, urlTag, methodTag)
 	carrier := opentracing.HTTPHeadersCarrier(req.httpReq.Header)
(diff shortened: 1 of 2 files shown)


herainman:
Update just for 100% UT coverage.....

Have to expose it as a public method since source code and test code is not under the same pkg name.....

herainman:
Add comments on the method which is exposed as public.

I tried to change source and test code as the same pkg name, however, I got the following errors:
# github.com/uber/zanzibar/runtime
import cycle not allowed in test
package github.com/uber/zanzibar/runtime (test)
	imports github.com/uber/zanzibar/examples/example-gateway/build/services/example-gateway
	imports github.com/uber/zanzibar/examples/example-gateway/build/services/example-gateway/module
	imports github.com/uber/zanzibar/examples/example-gateway/build/clients/bar
	imports github.com/uber/zanzibar/examples/example-gateway/build/clients/bar/module
	imports github.com/uber/zanzibar/runtime

FAIL	github.com/uber/zanzibar/runtime [setup failed]

It seems there is some cycle dependencies there....

PR #425, opened 2018-08-06
add req uuid in logger

cl1337:


ChuntaoLu:
One important thing we need to keep in mind is that per request logger does not necessarily mean only one child logger per request, depending on the scope, many more child logger may be in need, e.g. a task group may create a child logger and each individual tasks might also have a scope for child logger, etc. So scale is nondeterministic.

@jacobgreenleaf and @srivastavankit pointed out the logging approach that Google App Engine uses, which we could leverage. https://cloud.google.com/appengine/docs/standard/go/logs/

Issue #337, opened 2018-02-22
Upgrade httprouter package


jacobgreenleaf:
This repo depends on `julienschmidt/httprouter` who made a [breaking change in 2015](https://github.com/julienschmidt/httprouter/commit/70708e46004c7bcb09b70e685a8b74a690135387) by changing its API from `http.HandleFunc` to `http.Handler`. We are still using a prior version which means attempting to update past this change on `httprouter` triggers a build failure.

PR #427, opened 2018-08-08
Update httprouter to master

ChuntaoLu:
Fixes https://github.com/uber/zanzibar/issues/337

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -28,8 +28,9 @@ import:
   - lib/go/thrift
 - package: github.com/uber-go/tally
   version: ^3
+# update to master since the last semver release is 2 years old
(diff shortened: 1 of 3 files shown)


glide.yaml
```diff
@@ -29,7 +29,7 @@ import:
 - package: github.com/uber-go/tally
   version: ^3
 - package: github.com/julienschmidt/httprouter
-  version: v1.1
+  version: ^1.1
```
jacobgreenleaf: This is a noop.

ChuntaoLu: I was hoping there would be a semver release in the future. Updated to master anyway.

Issue #415, opened 2018-07-20
support typedef in query params


cl1337:
currently zanzibar / edge gateway always assume `?param=x1` is string, but we sometimes assign such value to typedef string uuid

in build 

when thrift schema specify field come from query params, e.g.
```
typedef string uuid

struct Bar {
   1: required uuid field1 ("zanzibar.http.ref = "query.myuuid"")
}
```

this will cause build failure

PR #431, opened 2018-08-16
Support typedef in query params

ChuntaoLu:
This PR adds typedef support thrift method parameters that decodes from (or encodes to) http query params.
Fixes https://github.com/uber/zanzibar/issues/415.


codegen/method.go
```diff
@@ -1301,15 +1354,42 @@ func (ms *MethodSpec) setParseQueryParamStatements(
 		statements.append("\treturn")
 		statements.append("}")
 
-		if field.Required {
-			statements.appendf("requestBody%s = %s", longFieldName, identifierName)
+		target := identifierName
+
+		indent := ""
```
cl1337: nit: I think without proper indent `go fmt` will also take care of it

ChuntaoLu: `go fmt` formats the generated code, but not the golden files.

PR #439, opened 2018-08-21
Json neutral

l-w-2017:
Refactor static_config in preparation for YAML support.
1. Add more test cases for static_config
2. Use Unmarshal to parse JSON config file at once instead of doing a lazy parsing. This simplifies logic and avoid JSON related logic scattered everywhere. After this change, adding the YAML support is easier and cleaner. The config file is not large. Parsing file at once should not have performance issue.
3. Update a bunch of auto generated files. Those files are in `examples/...` folder

### DIFF
--- a/runtime/static_config.go
+++ b/runtime/static_config.go
@@ -25,8 +25,9 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
+	"strconv"
 
-	"github.com/buger/jsonparser"
+	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 )
 
@@ -37,17 +38,11 @@ const (
 	fileContentsConfigType configType = 2
 )
 
-// StaticConfigValue represents a json serialized string.
-type StaticConfigValue struct {
-	bytes    []byte
-	dataType jsonparser.ValueType
-}
-
 // StaticConfig allows accessing values out of json config files
 type StaticConfig struct {
 	seedConfig    map[string]interface{}
 	configOptions []*ConfigOption
-	configValues  map[string]StaticConfigValue
+	configValues  map[string]interface{}
 	frozen        bool
 	destroyed     bool
 }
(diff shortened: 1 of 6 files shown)


examples/example-gateway/build/gen-code/endpoints/multi/multi/versioncheck.go
```diff
@@ -0,0 +1,10 @@
+// Code generated by thriftrw v1.8.0. DO NOT EDIT.
```
ChuntaoLu: Your vendor packages (specifically thriftrw) are outdated, therefore the noisy diffs in this PR. Please reinstall the packages and then update the PR.

runtime/static_config.go
```diff
@@ -283,27 +243,29 @@ func (conf *StaticConfig) MustGetStruct(key string, ptr interface{}) {
 // SetConfigValueOrDie sets the static config value.
 // dataType can be a boolean, number or string.
 // SetConfigValueOrDie will panic if the config is frozen.
-func (conf *StaticConfig) SetConfigValueOrDie(key string, bytes []byte, dataType string) {
+func (conf *StaticConfig) SetConfigValueOrDie(key string, str string, dataType string) {
```
ChuntaoLu: Since you are at it, the `dataType` should really be an enum type, so that there would be possible to set a invalid dataType.

l-w-2017: Unfortunately, there is a use case, where the type is from a config file. In that case the type is a string. If we change the type into an enum, then our customer need to parse the type. I would say let's keep this as it is for now. SetConfigValueOrDie now only accepts string "boolean", "number" and "string" as type.

ChuntaoLu: Do we have the uses case in /examples/example-gateway?

ChuntaoLu: ok, I see. We can probably deal with that in a separate PR.

ChuntaoLu:
Your build is failing because the coverage on runtime/static_config.go has dropped
master: https://travis-ci.org/uber/zanzibar/jobs/419860553#L1129
your branch: https://travis-ci.org/uber/zanzibar/jobs/419444106#L1137

runtime/static_config.go
```diff
@@ -283,27 +253,29 @@ func (conf *StaticConfig) MustGetStruct(key string, ptr interface{}) {
...
```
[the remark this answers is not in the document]
ChuntaoLu: > It seems like you want this method to work for both JSON and YAML.

Is that the reason? because `[]byte` would work as good as `string`.

PR #444, opened 2018-08-23
Switch to Logger(ctx) for success HTTP client log

jacobgreenleaf:


runtime/context.go
```diff
@@ -81,3 +91,27 @@ func GetRoutingDelegateFromCtx(ctx context.Context) string {
...
+		return nil
+	}
+	logger := val.(*zap.Logger)
+
+	return logger
+}
+
+// WithLogFields is a helper function to add the log field to the context
+func WithLogFields(ctx context.Context, fields ...zap.Field) context.Context {
+	lg := Logger(ctx)
```
cl1337: nit: just use log :)

runtime/context.go
```diff
@@ -34,6 +37,13 @@ const (
 	routingDelegateKey = contextFieldKey("rd")
 )
 
+const (
+	logFieldRequestMethod       = "method"
+	logFieldRequestURL          = "url"
+	logFieldRequestStartTime    = "timestamp-started"
+	logFieldRequestHeaderPrefix = "Request-Header"
```
cl1337: nit: seems like you're implicating case sensitive, is this intended ?

runtime/client_http_request.go
```diff
@@ -149,24 +149,40 @@ func (req *ClientHTTPRequest) WriteJSON(
...
+	)
+
 	return nil
 }
 
 // Do will send the request out.
-func (req *ClientHTTPRequest) Do(
-	ctx context.Context,
-) (*ClientHTTPResponse, error) {
+func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
```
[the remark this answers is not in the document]
jacobgreenleaf: I've updated the comments, but it was sort of implied before (since `started` and `startTime` make this class not reusable).

runtime/client_http_request.go
```diff
@@ -149,24 +149,40 @@ func (req *ClientHTTPRequest) WriteJSON(
 	httpReq.Header.Set("Content-Type", "application/json")
 
 	req.httpReq = httpReq
+	req.ctx = WithLogFields(req.ctx,
```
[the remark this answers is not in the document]
srivastavankit: +1. The context child logger should be created once for the client and we register all the context fields that we need to fetch from context.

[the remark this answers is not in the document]
jacobgreenleaf: Lu's comment refers to an earlier diff where we were creating a new `*zap.Logger` each time. Now instead we are just mutating the `[]zap.Fields` each time.

runtime/client_http_request.go
```diff
@@ -61,6 +61,7 @@ func NewClientHTTPRequest(
 		MethodName:     methodName,
 		client:         client,
 		Logger:         client.loggers[methodName],
+		ContextLogger:  NewContextLogger(client.loggers[methodName]),
```
ChuntaoLu: Now that we have a proper context logger, we will no longer need to create a child logger per method as in https://github.com/uber/zanzibar/blob/master/runtime/http_client.go#L62-L68, we only need to include the method name in the context log fields. That could be in different PR tho (please create an issue to remind us if you decide to have that done separately).

runtime/context.go
```diff
@@ -32,6 +33,16 @@ const (
 	endpointKey        = contextFieldKey("endpoint")
 	requestUUIDKey     = contextFieldKey("requestUUID")
 	routingDelegateKey = contextFieldKey("rd")
+	requestLogFields   = contextFieldKey("requestLogFields")
+)
+
+const (
+	logFieldRequestMethod       = "method"
```
ChuntaoLu: How should we differentiate the field between client and endpoint?

runtime/client_http_request.go
jacobgreenleaf: Yes, it'll be done in a separate diff.

runtime/client_http_response_test.go
```diff
@@ -76,11 +77,13 @@ func TestReadAndUnmarshalNonStructBody(t *testing.T) {
...
 	)
 	ctx := context.Background()
+	ctx = runtime.WithLogger(ctx, bgateway.ActualGateway.Logger)
+
 	req := zanzibar.NewClientHTTPRequest(ctx, "bar", "echo", client)
 
 	err = req.WriteJSON("POST", baseURL+"/bar/echo", nil, myJson{})
 	assert.NoError(t, err)
-	res, err := req.Do(context.Background())
+	res, err := req.Do()
```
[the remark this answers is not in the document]
jacobgreenleaf: I don't see that there is a difference for deadline propagation. It was redundant before since there were no mutations done to the context, but now `WriteJSON` modifies the context so it's actually confusing / misleading to pass context into `Do()`.

runtime/context.go
jacobgreenleaf: I'm preserving these log fields as they were before, open to better names. `endpoint-http-method`, `endpoint-url`, etc. seem better to me. Thoughts?

ChuntaoLu: That works, we should also think about how these can be customizable so that user can define log fields.

PR #452, opened 2018-08-30
Add support for HTTP delete with body

### LINKED ISSUE #446: Zanzibar does NOT send HTTP body on DELETE requests
HTTP spec allows the body on DELETE request but zanzibar does NOT send the body for any DELETE request.

https://github.com/uber/zanzibar/blob/master/codegen/templates/http_client.tmpl#L150

jacobgreenleaf:
Fixes #446


ChuntaoLu:
lgtm, `make check-generate` and checking in the changes should fix the build.

PR #458, opened 2018-09-06
Handle panic at endpoint

herainman:
This commits 

- add defer function and endpoint.tmpl and tchannel_endpoint.tmpl which handles all the panic at endpoint level
- add UTs both for HTTP and Tchannel request for this change


herainman:
I finally added UTs...

examples/example-gateway/endpoints/panic/endpoint_panic.go
```diff
@@ -0,0 +1,35 @@
...
+type EndpointPanicEndpoint struct {
+	Clients *module.ClientDependencies
+}
+
+// Handle ...
+func (w EndpointPanicEndpoint) Handle(
+	ctx context.Context,
+	headers zanzibar.Header,
+) (string, zanzibar.Header, error) {
+	defer func() {
```
jacobgreenleaf: You don't need to `defer` this. Just call `panic()` directly.

codegen/templates/endpoint.tmpl
```diff
@@ -101,7 +102,22 @@ func (h *{{$handlerName}}) HandleRequest(
 	req *zanzibar.ServerHTTPRequest,
 	res *zanzibar.ServerHTTPResponse,
 ) {
-	{{- if $reqHeaderRequiredKeys -}}
+	defer func() {
+		if r := recover(); r != nil {
+			stacktrace := string(debug.Stack())
+			e := errors.Errorf("enpoint panic: %v, stacktrace: %v", r, stacktrace)
+			h.Dependencies.Default.ContextLogger.Error(
```
jacobgreenleaf: Can you also increment a specific metric for panics so we can add alerting on specifically panics?

codegen/templates/endpoint.tmpl
```diff
@@ -101,7 +102,22 @@ func (h *{{$handlerName}}) HandleRequest(
...
+				ctx,
+				"endpoint panic",
+				zap.Error(e),
+				zap.String("stacktrace", stacktrace),
+				zap.String("endpoint", h.endpoint.EndpointName))
+
+			res.SendError(502, "Unexpected workflow panic, recovered at endpoint.", e)
```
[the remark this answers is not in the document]
herainman: fixed

examples/example-gateway/endpoints/panic/endpoint_panic.go
herainman: fixed

codegen/templates/endpoint.tmpl
herainman: fixed

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -80,7 +86,25 @@ func (h *{{$handlerName}}) Handle(
...
+				ctx,
+				"endpoint panic",
+				zap.Error(e),
+				zap.String("stacktrace", stacktrace),
+				zap.String("endpoint", h.endpoint.EndpointID))
+
+			h.endpointScope.Counter("endpoint.panic").Inc(1)
+			isSuccessful = false
+			response = nil
+			headers = map[string]string{}
```
cl1337: nit: do we want to preserve some header information if any

codegen/test_data/endpoints/bar_bar_method_argnotstruct.gogen
```diff
@@ -54,6 +58,9 @@ func NewBarArgNotStructHandler(deps *module.Dependencies) *BarArgNotStructHandle
 		"bar", "argNotStruct",
 		handler.HandleRequest,
 	)
+	handler.endpointScope = deps.Default.Scope.Tagged(map[string]string{
```
cl1337: let's just reuse this scope: https://github.com/uber/zanzibar/blob/master/runtime/router.go#L70

herainman: fixed

codegen/templates/tchannel_endpoint.tmpl
herainman: return nil instead of empty, this is the same behavior https://github.com/uber/zanzibar/pull/458/files/ef542b0b5925a2d569a7fc6d424379c6d39ff036#diff-7c8f7ec39bf077182fa0b404b3d49d76R124 when other error happens

herainman:
not sure why build and coverage not showing as the cadence.

However, I can see it is already passed the [build](https://travis-ci.org/uber/zanzibar/builds/426849761)

herainman:
thank you for review

PR #467, opened 2018-09-18
ScopeField Extractor

herainman:
This commits:
- Add request headers to scopeField
- Add PackageRoot at PackageInfo
- GetContextScopeExtractors defines the mappings between tags name and request headers


runtime/client_http_request.go
```diff
@@ -150,6 +150,7 @@ func (req *ClientHTTPRequest) WriteJSON(
 	httpReq.Header.Set("Content-Type", "application/json")
 
 	req.httpReq = httpReq
+	req.ctx = WithScopeFields(req.ctx, zap.String(scopeFieldRequestMethod, method))
```
jacobgreenleaf: You shouldn't use `zap.String`; the model of `tally.Scope` is `map[string]string`.

runtime/client_http_request.go
```diff
@@ -174,10 +175,13 @@ func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
 	}
 
 	logFields := make([]zap.Field, 0, len(req.httpReq.Header))
+	scopeFields := make([]zap.Field, 0, len(req.httpReq.Header))
 	for k, v := range req.httpReq.Header {
 		logFields = append(logFields, zap.String(fmt.Sprintf("%s-%s", logFieldRequestHeaderPrefix, k), v[0]))
+		scopeFields = append(logFields, zap.String(k, v[0]))
 	}
 	ctx = WithLogFields(ctx, logFields...)
+	ctx = WithScopeFields(ctx, scopeFields...)
```
jacobgreenleaf: We don't want to use headers as scope fields.

herainman:
For further PRs:
- add logField extractor
- use GetScopeTagFromCtx in http and tchannel request headers
- refactor router.go to pass in scope instead of counter and timer since we could only add tag at scope level.

runtime/context.go
```diff
@@ -124,6 +210,19 @@ func accumulateLogFields(ctx context.Context, newFields []zap.Field) []zap.Field
 	return fields
 }
 
+// ExtractScope extracts scope fields from a context into a tag.
+func ExtractScope(ctx context.Context, key string) map[string]string {
+	tags := map[string]string{}
+	if fn, ok := ContextScopeExtractors[key]; ok {
```
jacobgreenleaf: I thought this would be part of `Gateway` struct like most of the other singletons, that way we could have multiple per application or concurrently.

runtime/context.go
```diff
@@ -64,6 +86,39 @@ func GetRequestEndpointFromCtx(ctx context.Context) string {
...
+func WithEndpointRequestHeaderField(ctx context.Context, key string, value string) context.Context {
+	headers := GetEndpointRequestHeadersFromCtx(ctx)
+	headers[key] = value
+
+	return context.WithValue(ctx, endpointRequestHeader, headers)
+}
+
+// WithEndpointRequestHeadersField adds the endpoint request header information in the
+// request context.
+func WithEndpointRequestHeadersField(ctx context.Context, requestHeaders map[string]string) context.Context {
```
jacobgreenleaf: This function is unused, you are only using the single header version `WithEndpointRequestHeaderField`. You should only have one. I prefer this one since it can be used for single headers as well by passing in a map with a single item.

runtime/context.go
```diff
@@ -124,6 +201,32 @@ func accumulateLogFields(ctx context.Context, newFields []zap.Field) []zap.Field
...
+// NewContextExtractor returns a extractor that extracts log fields a context.
+func NewContextExtractor() ContextExtractor {
+	return &contextExtractor{}
+}
+
+type contextExtractor struct {
+}
+
+// ExtractScope extracts scope fields from a context into a tag.
+func (c *contextExtractor) ExtractScope(ctx context.Context, key string) map[string]string {
```
jacobgreenleaf: I figured all extractors would be used at every call site regardless of whether it's a client or an endpoint. For example, if the user defines a custom type `task` that defines their own context fields, the client metrics in zanzibar should pick up those task fields. As written this would require the user to carefully configure the map they return in the App struct as:

```
var App = &zanzibar.App{
    GetContextScopeExtractors: func() map[string]ContextExtractor {
        return map[string]ContextExtractor{
            "client": GetTaskAndClientTags,
            "task": GetTaskTags,
        }
    },
}
```

That is, they would need to be careful that the extractor for `"client"` returns task tags too, else the client metrics in zanzibar will not pick up the task tags. We don't really need that separation though, instead we can just return a list of functions that's run every time. While it does waste a bit of computation, it's probably a bit easier to understand:

```
var App = &zanzibar.App{
    GetContextScopeExtractors: func() []ContextExtractor {
        return []ContextExtractor{
            GetClientTags,
            GetTaskTags,
            GetRegionHeaderTags,
        }
    },
}
```

runtime/gateway.go
```diff
@@ -125,22 +127,26 @@ func CreateGateway(
...
 	var metricsBackend tally.CachedStatsReporter
 	var logWriter zapcore.WriteSyncer
+	var scopeExtractors []ContextScopeExtractor
 	if opts != nil && opts.MetricsBackend != nil {
 		metricsBackend = opts.MetricsBackend
 	}
 	if opts != nil && opts.LogWriter != nil {
 		logWriter = opts.LogWriter
 	}
+	if opts != nil && opts.ScopeExtractors != nil {
```
jacobgreenleaf: This code is a bit redundant. Can you change it to be something like 

```
if opts != nil {
     opts = &Options{}
}
```

Since the other nil checks and local variable assignments are redundant; if this nil check doesn't pass then `scopeExtractors` is still nil...

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Most built-in logs like `Finished an outgoing client HTTP request` now use the context logger. 
+- **Context Extractor**: Added [`ContextExtractor`](https://godoc.org/github.com/uber/zanzibar/runtime#ContextExtractor) interface. It can be used to define "extractors" or functions to pull out dynamic fields like trace ID, request headers, etc. out of the context to be used in log fields and metric tags. These can be used to pull out fields that are application-specific without adding code to zanzibar.
+- **PackageRoot**: Added PackageRoot for definition of root folder
```
jacobgreenleaf: You don't have to document this. The changelog isn't for exhaustively documenting every minor change. It's for notable changes. If you want to put it in the changelog you should be describing what it is, why one would use it, etc. but it is probably too minor to include here.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### BREAKING CHANGE
```
jacobgreenleaf: nit: Keep this under `Changed` section but just do bold **BREAKING**. 

Needs to be updated to reflect `Option` change.

codegen/templates/main.tmpl
```diff
@@ -42,8 +43,11 @@ func getConfig() *zanzibar.StaticConfig {
 
 func createGateway() (*zanzibar.Gateway, error) {
 	config := getConfig()
-	
-	gateway, _, err := service.CreateGateway(config, nil)
+	opts := &zanzibar.Options{
```
jacobgreenleaf: Since `app.AppOptions` is already an instance of `*zanzibar.Options`, you don't need to copy the fields over, just pass it as-is to `CreateGateway(config, app.AppOptions)`

PR #468, opened 2018-09-19
Replace JSON parser with YAML parser for *-configs

l-w-2017:


codegen/post_gen_hooks.go
```diff
@@ -89,10 +89,10 @@ func ClientMockGenHook(h *PackageHelper, t *Template) (PostGenHook, error) {
 		for _, instance := range clientInstances {
 			key := instance.ClassType + instance.InstanceName
 			var mc moduleConfig
-			if err := json.Unmarshal(instance.JSONFileRaw, &mc); err != nil {
+			if err := yaml.Unmarshal(instance.YAMLFileRaw, &mc); err != nil {
 				return errors.Wrapf(
 					err,
-					"error parsing client-config.json for client %q",
+					"error parsing client-config.yaml for client %q",
```
jacobgreenleaf: nit: Since it could be either `client-config.json` or `client-config.yaml` you should use the import path here rather than hard coding it.

codegen/module.go
```diff
@@ -698,76 +698,76 @@ func (system *ModuleSystem) readInstance(
 	instanceDirectory string,
 ) (*ModuleInstance, error) {
 
-	jsonFileName := className + configSuffix
+	yamlFileName := className + configSuffix
```
jacobgreenleaf: I think we should be looking for both `foo-config.yaml` and `foo-config.json`, preferring the `.yaml` file if it exists, otherwise falling back to `.json` for backwards compatibility.

codegen/template.go
```diff
@@ -60,7 +60,7 @@ var defaultFuncMap = tmpl.FuncMap{
 	"dec":           decrement,
 	"basePath":      filepath.Base,
 	"pascal":        PascalCase,
-	"jsonMarshal":   jsonMarshal,
+	"yamlMarshal":   yamlMarshal,
```
jacobgreenleaf: nit: from what I could tell this function is unused in the template.

codegen/module.go
l-w-2017: "configSuffix" is still "json". In the future diff, we will create a helper function to get the correct fileName for both json and yaml.

codegen/template.go
l-w-2017: function removed

codegen/gateway.go
```diff
@@ -236,14 +236,14 @@ func NewCustomClientSpec(
...
 				"client config %q must have %q field for type custom",
-				instance.JSONFileName,
+				instance.YAMLFileName,
 				f,
 			)
 		}
 	}
 
 	clientSpec := &ClientSpec{
-		JSONFile:           instance.JSONFileName,
```
jacobgreenleaf: There is existing code that relies on `JSONFile` property. Can you keep the existing property and add `YAMLFile` as a new property?

codegen/module.go
```diff
@@ -698,76 +698,77 @@ func (system *ModuleSystem) readInstance(
...
+		ClassType:             yamlConfig.Type,
 		BaseDirectory:         baseDirectory,
 		Directory:             instanceDirectory,
-		InstanceName:          jsonConfig.Name,
+		InstanceName:          yamlConfig.Name,
 		Dependencies:          dependencies,
 		ResolvedDependencies:  map[string][]*ModuleInstance{},
 		RecursiveDependencies: map[string][]*ModuleInstance{},
 		DependencyOrder:       []string{},
-		JSONFileName:          jsonFileName,
```
jacobgreenleaf: Keep `JSONFileRaw`

codegen/module.go
```diff
@@ -1318,10 +1319,12 @@ type ModuleInstance struct {
 	// DependencyOrder is the bottom to top order in which the recursively
 	// resolved dependency class names can depend on each other
 	DependencyOrder []string
-	// The JSONFileName is file name of the instance json file
-	JSONFileName string
-	// JSONFileRaw is the raw JSON file read as bytes used for future parsing
-	JSONFileRaw []byte
+	// The JSONFileName is file name of the instance JSON file
+	JSONFileName string // Deprecated
+	// The YAMLFileName is file name of the instance YAML file
```
jacobgreenleaf: Missing `JSONFileRaw`

codegen/post_gen_hooks.go
```diff
@@ -40,22 +40,22 @@ const (
 
 // helper struct to pull out the fixture config
 type moduleConfig struct {
-	Config *config `json:"config"`
+	Config *config `yaml:"config"`
```
jacobgreenleaf: Missing `json` tags

PR #470, opened 2018-09-26
Expose Scopes for HTTPClient

edwinzg:


runtime/http_client.go
```diff
@@ -35,6 +35,7 @@ type HTTPClient struct {
 	BaseURL        string
 	DefaultHeaders map[string]string
 	loggers        map[string]*zap.Logger
+	Scopes         map[string]tally.Scope
```
jacobgreenleaf: You should expose the singleton scope. These properties are going away in the future. I will probably mark them as deprecated soon.

edwinzg: If I expose the singleton scope how do I have separate tags for different methods?

PR #475, opened 2018-09-29
Expose an API to unmarshal raw ClassConfig

l-w-2017:


ChuntaoLu:
I think the refactor makes sense, but i am curious what's the intended use for the exposed `ConstructClassConfig`? With the given struct definition, unmarshaling is literally just `yaml.Unmarshal`, and `ConstructClassConfig` only adds some validations on top, are the validations the reason for the abstraction?

l-w-2017:
> I think the refactor makes sense, but i am curious what's the intended use for the exposed `ConstructClassConfig`? With the given struct definition, unmarshaling is literally just `yaml.Unmarshal`, and `ConstructClassConfig` only adds some validations on top, are the validations the reason for the abstraction?

Yeah. Now, the constructor only does some validation work. I don't know whether in the future we want to add any other logic into the constructor or not. I'm OK, if you think we should not expose the constructor.

codegen/module.go
```diff
@@ -1362,49 +1367,35 @@ type yamlClassConfig struct {
...
-	if readErr != nil {
-		return nil, errors.Wrapf(
-			readErr,
-			"Error reading class config %q",
-			classConfigPath,
-		)
-	}
-
-	parseErr := yaml.Unmarshal(configFile, &yamlConfig)
+func ConstructClassConfig(raw []byte, config *ClassConfig) error {
```
jacobgreenleaf: nit: this constructor is awkward, shouldn't it construct `*ClassConfig` for you?

```
func NewClassConfig(bytes []byte) (*ClassConfig, error)
```

ChuntaoLu:
> > I think the refactor makes sense, but i am curious what's the intended use for the exposed `ConstructClassConfig`? With the given struct definition, unmarshaling is literally just `yaml.Unmarshal`, and `ConstructClassConfig` only adds some validations on top, are the validations the reason for the abstraction?
> 
> Yeah. Now, the constructor only does some validation work. I don't know whether in the future we want to add any other logic into the constructor or not. I'm OK, if you think we should not expose the constructor.

I do like the refactor part, the constructor doesn't seem to be necessarily public unless you have a future use case in mind.

l-w-2017:
> > > I think the refactor makes sense, but i am curious what's the intended use for the exposed `ConstructClassConfig`? With the given struct definition, unmarshaling is literally just `yaml.Unmarshal`, and `ConstructClassConfig` only adds some validations on top, are the validations the reason for the abstraction?
> > 
> > 
> > Yeah. Now, the constructor only does some validation work. I don't know whether in the future we want to add any other logic into the constructor or not. I'm OK, if you think we should not expose the constructor.
> 
> I do like the refactor part, the constructor doesn't seem to be necessarily public unless you have a future use case in mind.

Yeah, you are right. External user will have access to "ClassConfig" soon. We are going to merge "tasks.yaml" into the "Config" field of "task-config.yaml". "ClassConfig" here is the internal representation of "tasks-config.yaml". In the future, any user of "tasks.yaml" will need to access "ClassConfig" to get the data captured in the "Config" field. In this diff we expose "ClassConfig".

PR #476, opened 2018-10-01
Parse either json or yaml

l-w-2017:
Make it possible to parse either YAML or JSON config files


scripts/generate.sh
```diff
@@ -13,7 +13,7 @@ ANNOPREFIX=${1:-zanzibar}
 bash ./codegen/runner/pre-steps.sh "$PREFIX/build" "$PREFIX" "$ANNOPREFIX"
 
 start=$(cat .TMP_ZANZIBAR_TIMESTAMP_FILE.txt)
-go run codegen/runner/runner.go -config="$PREFIX/build.json"
```
jacobgreenleaf: This is a breaking change; you should check for `build.yaml` and fall back to `build.json`

codegen/runner/pre-steps.sh
```diff
@@ -83,17 +83,28 @@ ABS_IDL_DIR="$(cd "$CONFIG_DIR" && pwd)/$(basename "$CONFIG_DIR/idl")"
...
+
+	module_type=$($processor -r .type "$config_file")
 	[[ ${module_type} != "http" ]] && continue
 	dir=$(dirname "$config_file")
-	json_files=$(find "$dir" -name "*.json")
-	for json_file in ${json_files}; do
-		thrift_file=$(jq -r '.. | .thriftFile? | select(strings | endswith(".thrift"))' "$json_file")
+	yaml_files=$(find "$dir" -name "*.json" -o -name "*.yaml")
+	for yaml_file in ${yaml_files}; do
+		processor="yq"
```
jacobgreenleaf: Note for myself, we now require user to have installed `yq`. I personally didn't have this on my laptop.

PR #478, opened 2018-10-04
Use ./vendor as path to vendor directory for finding mockgen.

jacobgreenleaf:
With pending customizations for config directory, the assumption that we can derive `vendor` folder absolute directory from GOPATH + config_dir + `"vendor"` is no longer true. We don't really need the flexibility to run mockgen from any directory, so just assume it is in `./vendor` for now.

### DIFF
--- a/codegen/mockgen.go
+++ b/codegen/mockgen.go
@@ -22,24 +22,20 @@ package codegen
 
 import (
 	"bytes"
+	"github.com/golang/mock/mockgen/model"
+	"github.com/pkg/errors"
 	"go/token"
 	"os"
 	"os/exec"
 	"path"
-	"path/filepath"
 	"runtime"
 	"sort"
 	"strconv"
 	"strings"
-
-	"github.com/golang/mock/mockgen/model"
-	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


codegen/mockgen.go
```diff
@@ -22,48 +22,35 @@ package codegen
...
-	pkgRoot := h.PackageRoot()
-	// we would not need this if example gateway is a fully standalone application
-	if pkgRoot == exampleGatewayPkg {
-		pkgRoot = zanzibarPkg
-	}
-	projRoot := filepath.Join(os.Getenv("GOPATH"), "src", pkgRoot)
-
 	// we assume that the vendor directory is flattened as Glide does
-	mockgenDir := path.Join(projRoot, "vendor", mockgenPkg)
+	mockgenDir := path.Join("vendor", mockgenPkg)
```
ChuntaoLu: This will end up with a path under `vendor/` which does not exist.

jacobgreenleaf: That's the intention, I think it's incorrect that it doesn't exist (see jenkins).

PR #479, opened 2018-10-04
Run easyjson for any module type with http

edwinzg:


README.md
```diff
@@ -384,6 +384,11 @@ git clone git@github.com:uber/zanzibar $GOPATH/src/github.com/uber/zanzibar
 cd $GOPATH/src/github.com/uber/zanzibar
 make install
 ```
+### Running make generate
+```
+make generate
+```
+If this fails due to `unknown command "examples/example-gateway/clients/baz/client-config.yaml" for "yq"` or `./codegen/runner/pre-steps.sh: line 97: yq: command not found` run `pip install yq` or `pip install --force-reinstall yq`
```
l-w-2017: I would say:
If code generation fails due to `yq: command not found`, then install `yq` by running `pip install yq`.
If you already installed `yq` and the build fails due to some error like `unknown command "examples/example-gateway/clients/baz/client-config.yaml" for "yq"`, then delete your installed `yq` and reinstall `yq` by running `pip install --force-reinstall yq`. This is because that not all distributions of `yq` works for Zanzibar. Zanzibar uses the python distribution.

PR #481, opened 2018-10-05
allow resolve i64 to be configured for path and annotation

cl1337:
application need to configure resolve i64 build path and annotation


scripts/resolve_i64/main.go
```diff
@@ -85,15 +85,17 @@ func getDirName() string {
 
 func main() {
 	thriftFile := os.Args[1]
+	stripPath := os.Args[2]
```
ChuntaoLu: The name of `stripPath` seems less descriptive.

PR #490, opened 2018-10-11
Make yq installation invisible for user

l-w-2017:


ChuntaoLu:
I thought you were going to vendor yq and use it the same way as thriftrw, what has changed?

jacobgreenleaf:
> I thought you were going to vendor yq and use it the same way as thriftrw, what has changed?

`yq` is a Python library, the only way to vendor is would be to do something like `virtualenv`. In general we should revisit this at a later time. The way we have usually solved this is by storing the binaries on some storage like GCS/S3 and downloading them to developer's laptop as needed.

PR #488, opened 2018-10-09
ContextMetrics

herainman:
This commit 
- implement ContextMetrics
- use ContextExtractor to extract http and tchannel request headers from Context and add it as ScoptTags to Context
- use ContextMetrics to emit http and TChannel metrics at endpoint and client level

PR #491, opened 2018-10-15
Add tags to http and tchannel client

herainman:
This commits add tags to http and Tchannel client

### DIFF
--- a/runtime/client_http_request.go
+++ b/runtime/client_http_request.go
@@ -56,13 +56,20 @@ func NewClientHTTPRequest(
 	clientID, methodName string,
 	client *HTTPClient,
 ) *ClientHTTPRequest {
+	scope := client.Scope.Tagged(map[string]string{
+		"client": clientID,
+		"method": methodName,
+	})
+
+	scopeFields := GetScopeFieldsFromCtx(ctx)
+	scope = scope.Tagged(scopeFields)
 	req := &ClientHTTPRequest{
 		ClientID:       clientID,
 		MethodName:     methodName,
 		client:         client,
 		Logger:         client.loggers[methodName],
 		ContextLogger:  NewContextLogger(client.loggers[methodName]),
-		metrics:        client.metrics[methodName],
+		metrics:        NewOutboundHTTPMetrics(scope),
 		defaultHeaders: client.DefaultHeaders,
 		ctx:            ctx,
 	}
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -68,7 +68,6 @@ type TChannelClient struct {
 	routingKey        *string
 	Loggers           map[string]*zap.Logger
 	Scopes            map[string]tally.Scope
-	metrics           map[string]*OutboundTChannelMetrics
 }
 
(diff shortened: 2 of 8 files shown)


herainman:
I will address the same issue at https://github.com/uber/zanzibar/pull/488 and land that PR first

runtime/client_http_request.go
```diff
@@ -56,13 +56,20 @@ func NewClientHTTPRequest(
 	clientID, methodName string,
 	client *HTTPClient,
 ) *ClientHTTPRequest {
+	scope := client.Scope.Tagged(map[string]string{
+		"client": clientID,
+		"method": methodName,
+	})
+
+	scopeFields := GetScopeFieldsFromCtx(ctx)
```
ChuntaoLu: It is not directly related to what this pr does, but we should refactor the function/variable names from `scopeFields` to `scopeTags` since `tags` is the standard terminology that tally uses. We can do it in a separate PR tho.

runtime/client_http_request.go
```diff
@@ -56,13 +56,20 @@ func NewClientHTTPRequest(
...
+	scope = scope.Tagged(scopeFields)
 	req := &ClientHTTPRequest{
 		ClientID:       clientID,
 		MethodName:     methodName,
 		client:         client,
 		Logger:         client.loggers[methodName],
 		ContextLogger:  NewContextLogger(client.loggers[methodName]),
-		metrics:        client.metrics[methodName],
+		metrics:        NewOutboundHTTPMetrics(scope),
```
ChuntaoLu: This probably will make a performance hit because now it creates sub scopes and metric objects per request while it is previously per client method. It does make sense because we are extracting scope tags dynamically at runtime. And I can't think of an easy way to get around it as the tags are unknown at compile time. For now let's put a todo comment here to document the impact on performance so that we know where to revisit in the future.

runtime/tchannel_client.go
```diff
@@ -140,14 +137,16 @@ func (c *TChannelClient) Call(
...
+	scope := c.Scopes[serviceMethod].Tagged(scopeFields)
 
 	call := &tchannelOutboundCall{
 		client:        c,
 		methodName:    c.methodNames[serviceMethod],
 		serviceMethod: serviceMethod,
 		reqHeaders:    reqHeaders,
 		logger:        c.Loggers[serviceMethod],
-		metrics:       c.metrics[serviceMethod],
+		metrics:       NewOutboundTChannelMetrics(scope),
```
ChuntaoLu: Let's put the same todo comment as in http client request.

runtime/tchannel_client.go
```diff
@@ -161,14 +160,16 @@ func (c *TChannelClient) CallThruAltChannel(
...
```
ChuntaoLu: same as above

runtime/tchannel_client_raw.go
```diff
@@ -22,7 +22,6 @@ package zanzibar
 
 import (
 	"context"
-
```
ChuntaoLu: We should keep this line for import grouping.

PR #492, opened 2018-10-16
Various bug fixes

ChuntaoLu:
This PR fixes a few bugs:
- Use correct trim fuction (`strings.TrimSuffix` instead of `strings.TrimRight`) to construct workflow file path;
- Assign value of refernce type to `Success` field of thriftrw generated result struct;
- Avoid declaring dependency nodes type multiple times in generated mock workflow.

I also have to include a bunch of noisy go import grouping changes since the latest goimports behaves as described in https://github.com/golang/go/issues/20818.


codegen/test_data/clients/bar.gogen
```diff
@@ -31,11 +31,10 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	zanzibar "github.com/uber/zanzibar/runtime"
```
cl1337: gogen lost its purpose when we moved ochestrator out zanzibar, I would actually prefer to have unit tests / integration tests verify I/O instead of the content of generated go code

codegen/test_data/endpoints/bar_bar_method_argnotstruct.gogen
```diff
@@ -30,14 +30,12 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
+	module "github.com/uber/zanzibar/.tmp_gen/endpoints/bar/module"
```
cl1337: lol the linter change

codegen/test_data/clients/bar.gogen
ChuntaoLu: Good point, that's on the todo list.

codegen/test_data/endpoints/bar_bar_method_argnotstruct.gogen
ChuntaoLu: I hope this import ordering/grouping thing is final.

PR #495, opened 2018-10-22
Add default tags to metrics

herainman:
This commits add host, type and dc as default tags at gateway.go

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -452,7 +452,6 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 	defaultTags := map[string]string{
 		"env":     env,
 		"service": service,
-		"type":    metricsType,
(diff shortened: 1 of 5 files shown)


runtime/gateway.go
```diff
@@ -452,6 +452,9 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 	defaultTags := map[string]string{
 		"env":     env,
 		"service": service,
+		"type":    metricsType,
```
jacobgreenleaf: This tag is redundant. Usually only M3 metrics are emitted with tags. Would you ever query on `type` field?

herainman: fixed

ChuntaoLu:
lgtm, but let's wait for merge before travis is back https://www.traviscistatus.com/

PR #497, opened 2018-10-22
Fix tchannel server error propagation

ChuntaoLu:
This PR fixes the tchannel server error propagation when unexpected error happens.
https://github.com/uber/zanzibar/blob/master/runtime/tchannel_server.go#L399 overwrites the error to nil when the `SendSystemError` call succeeds, which means https://github.com/uber/zanzibar/blob/master/runtime/tchannel_server.go#L237 gets a nil error back even when there is actually an error. https://github.com/uber/zanzibar/blob/master/runtime/tchannel_server.go#L243 which should have been bypassed in error case now is called and it naturally fails. This leads to that the `Unexpected tchannel system error` error is always followed by `Could not create arg2writer for inbound response` error.

### DIFF
--- a/runtime/tchannel_server.go
+++ b/runtime/tchannel_server.go
@@ -396,7 +396,9 @@ func (c *tchannelInboundCall) handle(ctx context.Context, wireValue *wire.Value)
 	}
 	if err != nil {
 		LogErrorWarnTimeoutContext(c.ctx, c.endpoint.contextLogger, err, "Unexpected tchannel system error")
-		err = c.call.Response().SendSystemError(errors.New("Server Error"))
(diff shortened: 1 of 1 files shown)


jacobgreenleaf:
Can you add a test similar to https://github.com/uber/zanzibar/blob/master/test/endpoints/tchannel/baz/baz_simpleservice_method_call_test.go#L259

ChuntaoLu:
> Can you add a test similar to https://github.com/uber/zanzibar/blob/master/test/endpoints/tchannel/baz/baz_simpleservice_method_call_test.go#L259

Done in https://github.com/uber/zanzibar/pull/498

PR #498, opened 2018-10-23
Clean up tchannel client/endpoint/server logs

ChuntaoLu:
This PR cleans up tchannel client/endpoint/server logs in the following way:
- remove redundent tchannel client/server runtime logs since errors are warpped and returned to callers where logging happens;
- generated endpoint handler log errors with `error` level, everywhere else errrors are logged with `warn` level;

There will be a another PR that takes cares of http client/endpoint/server logs.

### DIFF
--- a/runtime/tchannel_client.go
+++ b/runtime/tchannel_client.go
@@ -216,7 +216,6 @@ func (c *TChannelClient) call(
 			RequestState: rs,
 		})
 		if cerr != nil {
-			LogErrorWarnTimeout(call.logger, err, "Could not begin outbound request")
 			return errors.Wrapf(
 				err, "Could not begin outbound %s.%s (%s %s) request",
 				call.client.ClientID, call.methodName, call.client.serviceName, call.serviceMethod,
@@ -245,7 +244,6 @@ func (c *TChannelClient) call(
 	})
 
 	if err != nil {
-		LogErrorWarnTimeout(call.logger, err, "Could not make outbound request")
 		// Do not wrap system errors.
 		if _, ok := err.(tchannel.SystemError); ok {
 			return call.success, nil, err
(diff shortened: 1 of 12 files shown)


runtime/tchannel_client.go
```diff
@@ -277,7 +275,11 @@ func (c *tchannelOutboundCall) finish(err error) {
 	c.metrics.Latency.Record(c.finishTime.Sub(c.startTime))
 
 	// write logs
-	c.logger.Info("Finished an outgoing client TChannel request", c.logFields()...)
+	if err == nil {
+		c.logger.Info("Finished an outgoing client TChannel request", c.logFields()...)
+	} else {
+		c.logger.Warn("Failed to send outgoing client TChannel request", zap.Error(err))
```
herainman: should this be logger.error

runtime/tchannel_server.go
```diff
@@ -269,8 +269,11 @@ func (c *tchannelInboundCall) finish(err error) {
 	c.endpoint.Metrics.Latency.Record(c.finishTime.Sub(c.startTime))
 
 	// write logs
-	LogErrorWarnTimeoutContext(c.ctx, c.endpoint.contextLogger, err, "Thrift server error")
-	c.endpoint.contextLogger.Info(c.ctx, "Finished an incoming server TChannel request", c.logFields()...)
+	if err == nil {
+		c.endpoint.contextLogger.Info(c.ctx, "Finished an incoming server TChannel request", c.logFields()...)
+	} else {
+		c.endpoint.contextLogger.Warn(c.ctx, "Failed to serve incoming TChannel request", zap.Error(err))
```
herainman: is this contextLogger.Error

runtime/tchannel_client.go
ChuntaoLu: This is open to discussion. My reasoning is that when error happens, we will look at one error message (which is logged at endpoint handler and anywhere else is warn) and could understand what happened, if that is not sufficient we then look further for the warn logs associated with the request. Based on what I have experienced, multiple error log sites that describe the same cause create distraction and make logs less effective.

herainman: We could have a zap.field GUID associated with each warn, and let the error catch that GUID which is the root cause of this Error. We could have a runbook/wiki document all the mitigation step for each GUID and handle the oncall to SRE

ChuntaoLu: Once we use context.logger everywhere, we will have the request uuid in the context and all the logs that associated with the request will have the same request uuid.

PR #499, opened 2018-10-23
Metrics namespace change

herainman:
This commits:

- Use RootScope instead of PerHostScope
- Rename counter and timer
- Add protocol tag for HTTP and TChannel


runtime/gateway.go
```diff
@@ -476,11 +473,11 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 	gateway.AllHostScope = gateway.RootScope.SubScope(
```
ChuntaoLu: Is `AllHostScope` needed anymore?

ChuntaoLu:
This is a breaking change in a sense that dashboards built on previous metrics will no longer work. We need to put this in the changelog and communicate.

runtime/gateway.go
herainman: AllHostScope is needed since it is related to zap namespace, however, metrics would only have one scope which is RootScope

runtime/call_metrics.go
```diff
@@ -30,28 +30,36 @@ import (
...
+	statusTag        = "status"
+	protocolTag      = "protocal"
+	protocalHTTP     = "HTTP"
+	protocalTChannel = "TChannel"
+
+	clientScope   = "client"
+	endpointScope = "endpoint"
+
+	inboundCallsRecvd        = "request"
+	inboundCallsLatency      = "totalhandlertime"
```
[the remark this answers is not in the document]
herainman: Agreed. Totalhandlertime is to align with RTAPI

runtime/call_metrics.go
```diff
@@ -30,28 +30,36 @@ import (
...
 
-	outboundCallsSent         = "outbound.calls.sent"
-	outboundCallsLatency      = "outbound.calls.latency"
-	outboundCallsSuccess      = "outbound.calls.success"
-	outboundCallsAppErrors    = "outbound.calls.app-errors"
-	outboundCallsSystemErrors = "outbound.calls.system-errors"
-	outboundCallsErrors       = "outbound.calls.errors"
-	outboundCallsStatus       = "outbound.calls.status"
+	outboundCallsSent         = "request"
+	outboundCallsLatency      = "routerhandlertime"
```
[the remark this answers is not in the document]
herainman: Agreed. routerhandlertime is to align with RTAPI

runtime/call_metrics.go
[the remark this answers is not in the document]
ChuntaoLu: We don't have to align just for the sake of alignment. The name `totalhandlertime` is historical and shouldn't be inherited moving forward.

runtime/gateway.go
ChuntaoLu: > AllHostScope is needed since it is related to zap namespace and tchannel metrics namespace

Related in what way? Can you elaborate?

herainman: https://github.com/uber/zanzibar/blob/master/runtime/gateway.go#L556-L560
https://github.com/uber/zanzibar/blob/master/runtime/gateway.go#L619-L622
https://github.com/uber/zanzibar/blob/master/runtime/gateway.go#L580-L596

> > AllHostScope is needed since it is related to zap namespace and tchannel metrics namespace
> 
> Related in what way? Can you elaborate?

https://github.com/uber/zanzibar/blob/master/runtime/gateway.go#L556-L560
https://github.com/uber/zanzibar/blob/master/runtime/gateway.go#L619-L622
https://github.com/uber/zanzibar/blob/master/runtime/gateway.go#L580-L596

runtime/call_metrics.go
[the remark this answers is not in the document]
herainman: > We don't have to align just for the sake of alignment. The name `totalhandlertime` is historical and shouldn't be inherited moving forward.

fixed

runtime/call_metrics.go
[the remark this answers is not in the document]
herainman: fixed

runtime/call_metrics.go
```diff
@@ -30,28 +30,36 @@ import (
...
+	outboundCallsSuccess      = "success"
+	outboundCallsAppErrors    = "app-errors"
+	outboundCallsSystemErrors = "system-errors"
+	outboundCallsErrors       = "errors"
+	outboundCallsStatus       = "status"
```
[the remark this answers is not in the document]
herainman: fixed

runtime/gateway.go
ChuntaoLu: All these references can be replaced by the `RootScope`, the `AllHostScope` metric name `service + "." + env + ".all-workers"` is redundant and verbose, it is already tagged with `env` and `service`. I believe the original purpose for its existence is to reduce the cardinality introduced by tags `host` and `dc`, which I don't know if it is still a valid concern. If we want to have the same effect, we can replace it the `RootScope` before it is tagged with perHostTags, not the one that is overwritten here https://github.com/uber/zanzibar/pull/499/files#diff-ee4b13d64c39fd8e3d0483c1076bb31cR480.

runtime/call_metrics.go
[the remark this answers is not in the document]
ChuntaoLu: Did you forget to push to remote? Still showing `totalhandlertime` to me.

runtime/call_metrics.go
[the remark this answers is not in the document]
ChuntaoLu: same here

runtime/gateway.go
herainman: fixed

herainman:
Merged master, please review, thank you

runtime/metrics_namespace.go
```diff
@@ -20,63 +20,49 @@
...
-	outboundCallsSuccess      = "outbound.calls.success"
-	outboundCallsAppErrors    = "outbound.calls.app-errors"
-	outboundCallsSystemErrors = "outbound.calls.system-errors"
-	outboundCallsErrors       = "outbound.calls.errors"
-	outboundCallsStatus       = "outbound.calls.status"
+	clientRequest   = "client.request"
+	clientSuccess   = "client.success"
+	clientStatus    = "client.status"
+	clientErrors    = "client.errors"
+	clientAppErrors = "client.app-errors"
```
ChuntaoLu: Shouldn't there be a `clientSystemErrors = "client.system-errors"`?

runtime/tchannel_client.go
```diff
@@ -271,13 +272,13 @@ func (c *tchannelOutboundCall) finish(ctx context.Context, err error) {
 		errCause := tchannel.GetSystemErrorCode(errors.Cause(err))
 		scopeTags := map[string]string{scopeTagError: errCause.MetricsKey()}
 		ctx = WithScopeTags(ctx, scopeTags)
-		c.metrics.IncCounter(ctx, outboundCallsSystemErrors, 1)
+		c.metrics.IncCounter(ctx, clientErrors, 1)
```
ChuntaoLu: This should be `clientSystemErrors`

runtime/metrics_namespace.go
herainman: fixed

runtime/tchannel_client.go
herainman: fixed

PR #505, opened 2018-10-29
Add the ability to recognize (and not error) in the event of "transfo…

gjtrowbridge:
…rmError" middlewares.

Note that this diff does not add full support for this middleware, only the ability to recognize the middleware.


jacobgreenleaf:
Build is failing due to #506 I think.

PR #511, opened 2018-10-31
Add resolveHeaders tag to tchannel endpoint.request metrics

herainman:
This commits 

- Add resolveHeaders tag to tchannel endpoint.request metrics
- Correct spelling error of protocol tag

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -64,11 +64,12 @@ const (
 	scopeTagHandler         = "handlerid"
 	scopeTagError           = "error"
 	scopeTagStatus          = "status"
-	scopeTagProtocal        = "protocal"
+	scopeTagProtocol        = "protocol"
 	scopeTagHTTP            = "HTTP"
 	scopeTagTChannel        = "TChannel"
 	scopeTagsTargetService  = "targetservice"
 	scopeTagsTargetEndpoint = "targetendpoint"
+	scopeTagResolvedHeaders = "resolvedheaders"
 )
 
 // WithEndpointField adds the endpoint information in the
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -124,7 +124,7 @@ func (endpoint *RouterEndpoint) HandleRequest(
 	scopeTags := map[string]string{
 		scopeTagEndpoint: endpoint.EndpointName,
 		scopeTagHandler:  endpoint.HandlerName,
-		scopeTagProtocal: scopeTagHTTP,
+		scopeTagProtocol: scopeTagHTTP,
 	}
 
 	headers := map[string]string{}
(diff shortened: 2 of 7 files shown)


runtime/context.go
```diff
@@ -64,11 +64,12 @@ const (
 	scopeTagHandler         = "handlerid"
 	scopeTagError           = "error"
 	scopeTagStatus          = "status"
-	scopeTagProtocal        = "protocal"
+	scopeTagProtocol        = "protocol"
 	scopeTagHTTP            = "HTTP"
 	scopeTagTChannel        = "TChannel"
 	scopeTagsTargetService  = "targetservice"
 	scopeTagsTargetEndpoint = "targetendpoint"
+	scopeTagResolvedHeaders = "resolvedheaders"
```
ChuntaoLu: What is the semantics of this tag? How to decide when it is true or false?

herainman: This is used for tchannel endpoint.request metrics, currently, if we do `fetch endpoint.request`, it would return http request count + [tchannel request count](https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L270) + [tchannel request count with resolved headers](https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L239), this commits corrects the query request count result by query `fetch endpoint.request diff (fetch endpoint.request resolvedheaders:true)`

ChuntaoLu: Do we need this counter https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L270 at all? It seems to be counting the same request as https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L239, but with less tags.

herainman: > Do we need this counter https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L270 at all? It seems to be counting the same request as https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L239, but with potenitally fewer tags.

If we want to record all endpoint.request(including the errors one) for tchannel, then we need it.

ChuntaoLu: The only error we could miss is when it errors here https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L228. We can emit the request count metrics in the `finish` func https://github.com/uber/zanzibar/blob/1cbe93ed5b0781fc0bb6fe29c684c375a396434b/runtime/tchannel_server.go#L273, because it is always called, no matter there is error or not. I think it is not good to have 2 almost identical counters and a tag whose semantics is hard to reason about without knowing the internal implementation.

PR #523, opened 2018-11-16
Refactor HTTP router tests to be unit tests

jacobgreenleaf:


runtime/router_test.go
```diff
@@ -18,411 +18,201 @@
...
-		defaultTestOptions,
-		exampleGateway.CreateGateway,
-	)
-	if !assert.NoError(t, err) {
-		return
-	}
-	defer gateway.Close()
 
-	bgateway := gateway.(*benchGateway.BenchGateway)
+/*
```
shirchen: git is good at keeping history if you want to keep this one around. otherwise, let's delete.

runtime/router_test.go
```diff
@@ -18,411 +18,201 @@
...
-
-	assert.Equal(t, bytes, []byte("Internal Server Error\n"))
-
-	allLogs := gateway.AllLogs()
-	assert.Equal(t, 1, len(allLogs))
+func (s *routerSuite) TestExtractParameters() {
+	err := s.router.Handle("GET", "/foo/:a", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		params := ParamsFromContext(r.Context())
+		s.Equal("x", params.Get("a"))
+		w.Write([]byte("foo\n"))
```
shirchen: want to assert on output too if we are returning it?

runtime/router_test.go
```diff
@@ -18,411 +18,201 @@
...
-	assert.Equal(t, "an error", line["error"])
-	assert.Equal(t, "/panic", line["pathname"])
-	assert.Contains(
-		t,
-		line["errorVerbose"],
-		"runtime_test.TestRouterPanicObject.func1",
-	)
+func (s *routerSuite) TestPanicNonerror() {
+	err := s.router.Handle("GET", "/panic", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		panic("string")
```
shirchen: assert on response if returning?

runtime/router_test.go
```diff
@@ -18,411 +18,201 @@
...
-
-	line := logLines[0]
-	assert.Equal(t, "an error", line["error"])
-	assert.Equal(t, "/panic", line["pathname"])
-	assert.Contains(
-		t,
-		line["errorVerbose"],
-		"runtime_test.TestRouterPanicObject.func1",
-	)
+func (s *routerSuite) TestPanicNonerror() {
```
shirchen: name implies there is no error, but we check if there is one?

runtime/router_test.go
```diff
@@ -18,411 +18,201 @@
...
-	bytes, err := ioutil.ReadAll(resp.Body)
-	if !assert.NoError(t, err) {
-		return
-	}
-
-	assert.Equal(t, bytes, []byte("Method Not Allowed\n"))
-	assert.Equal(t, 1, len(
-		gateway.Logs("info", "Finished an incoming server HTTP request"),
-	))
+func (s *routerSuite) TestInvalidMethod() {
```
shirchen: can these error checks be folded into a single table test?

PR #531, opened 2018-12-11
Add more info to error messages

herainman:
This commits add prefix

- "Endpoint failure:"

- "Client failure:"

to existing error and warn messages at contextLogger which would help[ to indicate where(endpoint or client) this error happens


herainman:
Thank you for review

PR #539, opened 2019-01-08
Added a circuit breaker for http and tchannel client calls

abhishekparwal:
Added a circuit breaker for http and tchannel client calls for preventing downstream from failures/outages/high latency


runtime/plugins/m3_aggregator.go
```diff
@@ -0,0 +1,117 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package plugins allows users to operate on statistics recorded for each circuit operation.
+// Plugins should be careful to be lightweight as they will be called frequently.
+package plugins
+
+import (
+	"github.com/uber-go/tally"
```
ChuntaoLu: import grouping

codegen/templates/http_client.tmpl
```diff
@@ -167,7 +176,11 @@ func (c *{{$clientName}}) {{$methodName}}(
 	}
 	{{- end}}
 
-	res, err := req.Do()
+	var res *zanzibar.ClientHTTPResponse
+	err = hystrix.Do("{{$clientID}}", func() error {
```
ChuntaoLu: We should call `hystrix.DoC` with the proper context here, so the context is respected.

codegen/templates/tchannel_client.tmpl
```diff
@@ -5,6 +5,7 @@ package {{$instance.PackageInfo.PackageName}}
 import (
 	"context"
 	"errors"
+	"github.com/afex/hystrix-go/hystrix"
```
ChuntaoLu: import grouping

codegen/templates/http_client.tmpl
```diff
@@ -5,6 +5,7 @@ package {{$instance.PackageInfo.PackageName}}
 import (
 	"context"
 	"fmt"
+	"github.com/afex/hystrix-go/hystrix"
```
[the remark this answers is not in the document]
abhishekparwal: ok, while doing codegen we are already removing , reorganizing imports along with code fmt.

runtime/gateway.go
```diff
@@ -23,6 +23,8 @@ package zanzibar
 import (
 	"context"
 	"fmt"
+	"github.com/afex/hystrix-go/hystrix/metric_collector"
```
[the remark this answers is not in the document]
abhishekparwal: same as above

codegen/templates/http_client.tmpl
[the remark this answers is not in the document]
ChuntaoLu: That is correct, but consistency is good.

codegen/templates/tchannel_client.tmpl
abhishekparwal: same as above

ChuntaoLu:
Copyright change is landed on master, rebase would get rid of the noises.

abhishekparwal:
#
#   benchmark output circuit_breaker
#

    Latency    32.00ms   33.25ms   1.08s    88.01%
    Req/Sec     1.20k   203.56     2.12k    70.97%
  429092 requests in 30.04s, 47.06MB read
  Socket errors: connect 0, read 235, write 0, timeout 0
  Non-2xx or 3xx responses: 14
Requests/sec:  14284.97
Transfer/sec:      1.57MB

#
#   benchmark output master
#

    Latency    35.75ms   32.25ms 392.37ms   65.52%
    Req/Sec     1.05k   152.10     2.02k    77.17%
  376117 requests in 30.04s, 41.25MB read
  Socket errors: connect 0, read 253, write 0, timeout 0
Requests/sec:  12520.04
Transfer/sec:      1.37MB
Switched to branch 'circuit_breaker'

README.md
```diff
@@ -140,6 +140,23 @@ func (c *quux) Echo(s string) string { return s }
 ```
 Note the type of `deps` parameter passed to `NewClient` constructor function is generated by Zanzibar, as indicated by the import path. Zanzibar takes care of initializing and passing in the acutal `deps` argument, as mentioned in [Dependency Injection](#dependency-injection).
 
+###### Circuit Breaker
+For increasing overall system resiliency, zanzibar uses a [Circuit Breaker](https://msdn.microsoft.com/en-us/library/dn589784.aspx) to avoid calling clients when there is a increase in failure rate beyond a set
```
ChuntaoLu: typo: an increase

README.md
```diff
@@ -140,6 +140,23 @@ func (c *quux) Echo(s string) string { return s }
 ```
 Note the type of `deps` parameter passed to `NewClient` constructor function is generated by Zanzibar, as indicated by the import path. Zanzibar takes care of initializing and passing in the acutal `deps` argument, as mentioned in [Dependency Injection](#dependency-injection).
 
+###### Circuit Breaker
+For increasing overall system resiliency, zanzibar uses a [Circuit Breaker](https://msdn.microsoft.com/en-us/library/dn589784.aspx) to avoid calling clients when there is a increase in failure rate beyond a set 
+threshold. It recovers it after elasped time to again fo in half-open and close state.
```
ChuntaoLu: > It recovers it after elasped time to again fo in half-open and close state.

Can you rework this sentence? a few typos and the information it tries to convey doesn't seem clear to me.

README.md
```diff
@@ -140,6 +140,23 @@ func (c *quux) Echo(s string) string { return s }
...
+
+To set max concurrent requests to client, set: 
+```
+	"clients.<clientID>.maxConcurrentRequests": 1000
+```
+To set error percent threshold beyond which to open the circuit to the client, set:
+```
+    "clients.<clientID>.errorPercentThreshold": 20
+```
+By default circuit breaker is enabled, if you want to disable it , set :
```
ChuntaoLu: redundant white space before last `,`

examples/example-gateway/config/production.yaml
```diff
@@ -3,28 +3,44 @@ clients.bar.defaultHeaders:
...
+clients.corge-http.errorPercentThreshold: 20
 clients.corge.serviceName: Corge
 clients.corge.timeout: 10000
 clients.corge.timeoutPerAttempt: 2000
+clients.corge.maxConcurrentRequests: 1000
+clients.corge.errorPercentThreshold: 20
 clients.google-now.ip: 127.0.0.1
 clients.google-now.port: 14120
 clients.google-now.timeout: 10000
+clients.google-now.maxConcurrentRequests: 1000
```
ChuntaoLu: Let's put the disabled config here as well, to be consistent with the json config. (yaml is prefered, but zanzibar is backward compatible with json config)

examples/example-gateway/config/test.yaml
```diff
@@ -3,27 +3,41 @@ clients.bar.defaultHeaders:
 clients.bar.ip: 127.0.0.1
 clients.bar.port: 4001
 clients.bar.timeout: 10000
+clients.bar.maxConcurrentRequests: 1000
```
ChuntaoLu: same as above

README.md
```diff
@@ -140,6 +140,40 @@ func (c *quux) Echo(s string) string { return s }
 ```
 Note the type of `deps` parameter passed to `NewClient` constructor function is generated by Zanzibar, as indicated by the import path. Zanzibar takes care of initializing and passing in the acutal `deps` argument, as mentioned in [Dependency Injection](#dependency-injection).
 
+###### Circuit Breaker
+For increasing overall system resiliency, zanzibar uses a [Circuit Breaker](https://msdn.microsoft.com/en-us/library/dn589784.aspx) which avoids calling client when there is an increase in failure rate beyond a set 
+threshold. After a sleepWindowInMilliseconds, client calls are attempted recovery by going in half-open and then close state. 
+
+circuitBreakerDisabled: Default false. Disables the circuit-breaker:
```
ChuntaoLu: Disables -> To disable

README.md
```diff
@@ -140,6 +140,40 @@ func (c *quux) Echo(s string) string { return s }
...
+###### Circuit Breaker
+For increasing overall system resiliency, zanzibar uses a [Circuit Breaker](https://msdn.microsoft.com/en-us/library/dn589784.aspx) which avoids calling client when there is an increase in failure rate beyond a set 
+threshold. After a sleepWindowInMilliseconds, client calls are attempted recovery by going in half-open and then close state. 
+
+circuitBreakerDisabled: Default false. Disables the circuit-breaker: 
+```
+    "clients.<clientID>.circuitBreakerDisabled" : true
+```
+
+maxConcurrentRequests: Default 50. Sets how many requests can be run at the same time, beyond which requests are
```
ChuntaoLu: Sets -> To set

Same for the following 3 configurations

PR #540, opened 2019-01-08
Fix staticcheck errors in generated code

gjtrowbridge:
This diff addresses a few observed static check errors in the generated code for the service.

### DIFF
--- a/codegen/template_bundle/template_files.go
+++ b/codegen/template_bundle/template_files.go
@@ -1235,7 +1235,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"runtime"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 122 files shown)


codegen/template_bundle/template_files.go
```diff
@@ -1388,7 +1382,7 @@ func listenOnSignals() {
 	signal.Notify(sigs, syscall.SIGUSR2)
 
 	go func() {
-		_ = <-sigs
+		<-sigs
```
gjtrowbridge: The `_` is unnecessary here and triggers a static check error.

codegen/template_bundle/template_files.go
```diff
@@ -1253,11 +1252,6 @@ import (
 
 var configFiles *string
 
-func getDirName() string {
```
gjtrowbridge: This function is not used.

codegen/template_bundle/template_files.go
```diff
@@ -1820,6 +1814,7 @@ func CreateGateway(
 
 // RegisterDeps registers direct dependencies of the service
 func RegisterDeps(g *zanzibar.Gateway, deps *module.Dependencies) error {
+	//lint:ignore S1021 allow less concise variable declaration for ease of code generation
```
gjtrowbridge: Adding a lint exception for this because declaring `var err error` makes the code generation easier, so we just need to make sure static check does not complain.

codegen/templates/main.tmpl
```diff
@@ -26,11 +25,6 @@ import (
 
 var configFiles *string
 
-func getDirName() string {
```
gjtrowbridge: function is not used.

PR #542, opened 2019-01-09
Support flexible folder structures

jacobgreenleaf:
An assumption of the base module system was that the module classes (clients, endpoints, etc.) would have a single folder structure that would be extended (but not modified) by consumer applications. Specifically, the code generation system is given a base directory (`baseDirectory`) and it assumes the built-in modules would be located in exactly the subdirectory of that base directory (e.g. if `baseDirectory = "examples/example-gateway"` then `endpoints` will always be located in `examples/example-gateway/endpoints`). That assumption is encoded in the `Directories` property, which enumerates all the places where Zanzibar will look for module instances.  

This poses a problem for supporting more flexible directory structures, such as keeping clients at `examples/example-gateway/clients` but allowing instances of `endpoints` to exist in `examples/example-gateway/apps/foo/endpoints`; applications could not customize the place where Zanzibar would look for module instances. 

This patch changes the module search logic to be more flexible. Instead of the assumption that the place where Zanzibar configuration files like `clients` will be found is known ahead of time, we instead assume that we will be provided a list of globbing patterns to enumerate all the configuration files that exist:

```
moduleSearchPaths:
- clients/*
- middlewares/*
- endpoints/*
- endpoints/tchannel/*
- services/*
- app/*/clients/*
- app/*/endpoints/*
- app/*/services/*
```

The list will be provided in the `build.yaml` file which is provided by the application and can be customized without modifying Zanzibar. Zanzibar will then look for **every** instance type  it is aware of in every folder. For example, you *could* have:

```
clients/
  foo/
    client-config.yaml
  bar/
    endpoint-config.yaml
```

There's nothing special about the relationship between the module classes and their paths. 

Dependency names as stored in the `map[string][]*ModuleInstance` "tree" are now first extracted from the path name:

| Path | Qualified Instance Name |
| :-- | :-- |
| `clients/foo/client-config.yaml` | `foo` |
| `nested/clients/foo/client-config.yaml` | `nested/foo` |
| `nested/bar/foo/baz/foo-config.yaml` | `nested/bar/baz` |
| `app/demo/services/xyz/service-config.yaml` | `app/demo/xyz` |

As a convenience, the second to last part of the qualified name is removed (`clients/foo` is transformed to `foo`). To depend on a so-called nested dependency, you would use the qualified instance name in the dependency list:

```
dependencies:
  endpoint:
  - app/demo/abc
  - health
```

TODO:
- [x] Create an example case for fully qualified dependency name
- [x] Verify nested service starts up
- [x] Fix mock generation
- [x] Fix lint
- ~~[ ] (?) Remove `name` field from `*-config.yaml` since it's now implied by the path~~


codegen/module.go
```diff
@@ -550,64 +568,93 @@ func (system *ModuleSystem) ResolveModules(
...
+				targetGenDir,
+				moduleDir,
+			)
+			if instanceErr != nil {
+				if _, ok := instanceErr.(inferError); ok {
+					// It could be the case that a glob pattern like "endpoints/*" matches a folder that is later
+					// matched by a longer glob pattern like "endpoints/tchannel/*", so merely print a warning here
+					// rather than failing to generate. It will fail at a later step if no globbing pattern matches
+					// the wanted instance.
+					fmt.Printf("warn: globbing pattern %q: %s\n", moduleDirectoryGlob, instanceErr.Error())
```
ChuntaoLu: This warning message seems unnecessary to me, it warns for each dir that does not contain a module config file, which could become quite a clutter when you have many modules nested.

codegen/module.go
```diff
@@ -616,113 +663,56 @@ func (system *ModuleSystem) ResolveModules(
...
-		if !file.IsDir() {
-			continue
+	var (
+		className, jsonFileName, yamlFileName, classConfigPath string
+	)
+	for _, instanceFile := range instanceFiles {
+		if strings.HasSuffix(instanceFile.Name(), yamlConfigSuffix) {
+			className = strings.TrimSuffix(instanceFile.Name(), yamlConfigSuffix)
+			yamlFileName = instanceFile.Name()
+			// TODO(jacobg): Is this necessary?
```
ChuntaoLu: I don't think it's necessary, should just delete it to avoid future confusion.

codegen/module.go
```diff
@@ -1215,8 +1212,11 @@ func FormatGoFile(filePath string) error {
 // THis could be something like an Endpoint class which contains multiple
 // endpoint configurations, or a Lib class, that is itself a module instance
 type ModuleClass struct {
-	Name        string
-	ClassType   moduleClassType
+	Name       string
+	NamePlural string
+	ClassType  moduleClassType
+
+	// Deprecated: Hard coded list of directory names is replaced with dynamic module search path in build.yaml.
```
ChuntaoLu: same as earlier, deletion instead of deprecation.

examples/example-gateway/build.yaml
```diff
@@ -9,3 +9,12 @@ packageRoot: github.com/uber/zanzibar/examples/example-gateway
 targetGenDir: ./build
 thriftRootDir: ./idl
 traceKey: x-trace-id
+moduleSearchPaths:
+- clients/*
+- middlewares/*
+- endpoints/*
```
ChuntaoLu: Since this is glob, I would not be surprised to see patterns like "endpoints/*/**" in the yaml file, can we have such a case to make sure things work as intended?

jacobgreenleaf: `**` is not supported, see https://golang.org/pkg/path/filepath/#Match

ChuntaoLu: okay, does that mean `endpoint/*` does not match `endpoint/foo/bar/endpoint-config.yaml`?

jacobgreenleaf: Yes, you'd need to write `endpoint/*/*` to capture that, we might want a more flexible globbing pattern library but I just used the standard library one for now.

ChuntaoLu: Okay, got it, this is definitely something worth to document.

PR #545, opened 2019-01-12
Add tagged metrics for Thrift app errors

jacobgreenleaf:


runtime/constants.go
```diff
@@ -34,6 +34,9 @@ const (
 	// MetricEndpointPanics is endpoint level panic counter
 	MetricEndpointPanics = "endpoint.panic"
 
+	// MetricEndpointAppErrors is endpoint level panic counter
```
ChuntaoLu: copypasta

runtime/constants.go
```diff
@@ -34,6 +34,9 @@ const (
...
+	MetricEndpointAppErrors = "endpoint.app-errors-err"
```
ChuntaoLu: Maybe just `endpoint.app-errors`?

jacobgreenleaf: `endpoint.app-errors` already exists. There is some generic code outside the template that processes the TChannel server result. Because it's outside the template it isn't straightforward to add a tag per error type. It would be possible to just remove the other stat and just emit the stat once with the caveat that it makes the generic code a bit more complex by omission.

edwinzg: +1

jacobgreenleaf:
> are we going to have a system-error tag too? or is this only for business errors?

Only business errors; system errors won't have a Thrift type associated with them so there is no additional tag to add.

PR #546, opened 2019-01-14
sorting is non-deterministic for js.type annotations

rpatali:
the sorting is dependent on the type currently. If there are 2 fields
with `js.type="Date"`, the order of the generated code for
Marshal/Unmarshal is underterministic.

This diff fixes that.


scripts/resolve_i64/main.go
```diff
@@ -25,22 +25,22 @@ import (
...
 	"html/template"
-	"bytes"
 	"io/ioutil"
-	"github.com/uber/zanzibar/runtime"
-	"runtime"
 	"path/filepath"
-	"strings"
-	"github.com/uber/zanzibar/codegen"
-	"github.com/pkg/errors"
+	"runtime"
```
ChuntaoLu: since you are at it, let's group all stdlibs to the first import group

rpatali: will do.

PR #551, opened 2019-01-18
Removed adding multiple uber source request headers

sandeepboys:


runtime/client_http_request.go
```diff
@@ -143,11 +143,11 @@ func (req *ClientHTTPRequest) WriteJSON(
 	// Using `Add` over `Set` intentionally, allowing us to create a list
 	// of headerValues for a given key.
 	for headerKey, headerValue := range req.defaultHeaders {
-		httpReq.Header.Add(headerKey, headerValue)
+		httpReq.Header.Set(headerKey, headerValue)
```
ChuntaoLu: I don't think this is the right fix, per [rfc2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) http header can have multiple values.

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

This pr changes the header semantics to be single valued per header key, which is not correct by spec.

PR #552, opened 2019-01-29
Add support for Adapters

Ryanfsdf:
Support for adapters


README.md
```diff
@@ -121,9 +125,12 @@ Configurations are the interface that developers interact with when using the Za
 Because configurations are the core of a Zanzibar application, we create a root directory to host configuration files when starting a Zanzibar application. There are a few typical directories and files under the root directory. Take [example-gateway](https://github.com/uber/zanzibar/tree/master/examples/example-gateway) for example:
 ```
 example-gateway                 # root directory
+├── adapters                    # config directory for modules of adapter module class
+│   └── rate-limiter            # config directory for an adapter named 'rate-limiter'
```
rpatali: use a different example, on line 75 you are using rate-limiter as a middleware.

runtime/adapters.go
```diff
@@ -0,0 +1,47 @@
...
+		shared SharedState,
+	) bool
+	// implement HandleResponse for your adapter. Return false
+	// if the handler writes to the response body.
+	HandleResponse(
+		ctx context.Context,
+		res *ServerHTTPResponse,
+		shared SharedState,
+	)
+	JSONSchema() *jsonschema.Document
```
rpatali: how is this going to be used?

rpatali: middlewares have configurations that can differ on per endpoint basis. I don't think we want that to be the case for Adapter. My understanding is that adapter's don't take configurations per endpoint and their behavior is always the same.

runtime/adapters.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
rpatali: I understand that you are following the existing convention but I did say call this adapters_http. perhaps you can do the same for middlewares too.

runtime/execution_stack.go
```diff
@@ -0,0 +1,138 @@
...
+	res *ServerHTTPResponse) {
+
+	shared := NewSharedState(m.adapters, m.middlewares)
+
+	for i := 0; i < len(m.adapters); i++ {
+		ok := m.adapters[i].HandleRequest(ctx, req, res, shared)
+		// If a adapter errors and writes to the response header
+		// then abort the rest of the stack and evaluate the response
+		// handlers for the adapters seen so far.
+		if ok == false {
```
rpatali: `if !ok {`

README.md
Ryanfsdf: Will change to log-publisher

runtime/adapters.go
Ryanfsdf: For consistency, this will require further changes to method names. For example, in `middlewares.go`, the handle is named `MiddlewareHandle`. In `middlewares_tchannel.go`, the handle is named `MiddlewareTchannelHandle`. For consistency, the handle in `middlewares.go` should be renamed to `MiddlewareHttpHandle`. There are other examples like this where http is left out in the name even though it is explicit for tchannel. Should I rename all these methods as well?

codegen/module.go
```diff
@@ -1222,6 +1260,9 @@ type ModuleClass struct {
 	DependedBy  []string
 	types       map[string]BuildGenerator
 
+	// Every module instance of this module class imports all instances of a module class that is implicitly depended on
+	ImplicitDependsOn []string
```
ChuntaoLu: This approach is interesting, I can see it would make sense for certain module class types, however I don't think it is the way to solve what we want to achieve with this PR. `adapter` and `middleware` are fundamentally the same abstraction, the fact that adapters are mandatory for every endpoint does not add more meaning to the middleware abstraction (at least not enough for a spin off), what it introduces is more a rule for a subset of middlewares. 

If the goal here is no more than "to have a list of default middlewares that every endpoint implicitly depends on", then I think it is an implementation detail of the endpoint generation and needs not level up to an approach that alters high level abstractions (which should be kept simple). I am a fan of explicit, but I admit having each endpoint explicitly declare dependency on the set of default middlewares hurts user experience and sometimes is error prone. Maintaining a similar file as you did with https://github.com/uber/zanzibar/pull/552/files#diff-223333e63915149a0b7e1a4ad57fc256 in middelwares directory(I can also see the file in endpoints dir making sense in a way) is probably a reasonable trade off.

PR #553, opened 2019-02-04
Add support for mandatory middleware

Ryanfsdf:
Introduced support for mandatory middlewares and also the concept of "default dependencies"


codegen/gateway.go
```diff
@@ -420,7 +420,71 @@ func NewEndpointSpec(
...
+	middlewareSpecs map[string]*MiddlewareSpec,
+	classType string,
+) ([]MiddlewareSpec, error) {
+	middlewareObj := map[string][]string{}
+
+	const mandatoryMiddlewareFile = "middlewares/mandatory"
+	middlewareOrderingFile := filepath.Join(cfgDir, mandatoryMiddlewareFile+".yaml")
+	if _, err := os.Stat(middlewareOrderingFile); os.IsNotExist(err) {
+		// Cannot find yaml file, use json file instead
+		middlewareOrderingFile = filepath.Join(cfgDir, mandatoryMiddlewareFile+".json")
```
ChuntaoLu: The reason we check for .json file is that we previously used JSON files for config, and we need to have backward compatibility when we moved to yaml. Since this is a new feature and there is no backward compatibility requirement, I would say let's not support json at all. We will probably ditch support for json in the future.

codegen/gateway.go
```diff
@@ -420,7 +420,71 @@ func NewEndpointSpec(
...
 	}
 
-	return augmentEndpointSpec(espec, endpointConfigObj, midSpecs)
+	mandatoryMidSpecs, err := getOrderedMandatoryMiddlewareSpecs(
+		h.ConfigRoot(),
+		h.MandatoryMiddlewareSpecs(),
+		endpointType.(string))
+	if err != nil {
+		return nil, errors.Wrap(
+			err, "Error getting ordered mandatory middleware specs",
```
ChuntaoLu: Error message usually should not start with capital letter because it's often wrapped by another error or prefixed by another message. https://github.com/golang/go/wiki/Errors#errors. (sadly the existing error strings in the repo are not the best examples)

README.md
```diff
@@ -142,7 +143,10 @@ example-gateway                 # root directory
 │   ├── clients                 # idl directory for client thrift files
 │   └── endpoints               # idl directory for endpoint thrift files
 ├── middlewares                 # config directory for modules of middleware module class
-│   └── transform-response      # config directory for a middleware named 'transform-response'
+│   ├── transform-response      # config directory for a middleware named 'transform-response'
+│   ├── mandatory               # directory for all mandatory middlewares
+│   │   └── log-publisher       # config directory for a mandatory middleware named 'log-publisher'
+│   └── mandatory.yaml          # config file describing mandatory middlewares and their execution order
```
Ryanfsdf: Abhishek has brought to my attention that the contents of the `mandatory.yaml` file can actually be put inside the config field of `services/example-gateway/service-config.yaml`. For example, right now the `mandatory.yaml` file looks like this:
```
http:
  - mandatory_example2
  - mandatory_example
tchannel:
  - mandatory_example_tchannel
``` 
We can move this into `service-config.yaml` and that will look like this:
```
config: {
  mandatory_http_middleware:
  - mandatory_example2
  - mandatory_example
  mandatory_tchannel_middleware:
  - mandatory_example_tchannel
}
dependencies:
  endpoint:
  - bar
  - baz
  - contacts
  - googlenow
  - bazTChannel
  - multi
  - panic
  - panicTChannel
name: example-gateway
type: gateway
```
This makes sense because the mandatory middleware ordering is a service level configuration. Right now, I am leaning towards putting the middleware ordering in the `service-config.yaml` file. The code changes to implement this would be minimal (just changing where the config is read from). Any thoughts on this? @ChuntaoLu

codegen/module.go
```diff
@@ -1212,15 +1336,16 @@ func FormatGoFile(filePath string) error {
...
-	Directories []string
-	DependsOn   []string
-	DependedBy  []string
-	types       map[string]BuildGenerator
+	Name           string
+	ClassType      moduleClassType
+	Directories    []string
+	DependsOn      []string
+	DependedBy     []string
+	DefaultDepDirs []string // Default Dependencies
```
ChuntaoLu: If you put the comment on top of the line, IDEs will recognize it's a comment for the field and displays it when proper.

codegen/package.go
```diff
@@ -177,22 +188,31 @@ func NewPackageHelper(
 	middlewareSpecs, err := parseMiddlewareConfig(options.relMiddlewareConfigDir(), absConfigRoot)
 	if err != nil {
 		return nil, errors.Wrapf(
-			err, "Cannot load middlewares:")
+			err, "Cannot load middlewares")
```
ChuntaoLu: we can kill the line breaking here.

codegen/package.go
```diff
@@ -212,6 +232,11 @@ func (p PackageHelper) MiddlewareSpecs() map[string]*MiddlewareSpec {
 	return p.middlewareSpecs
 }
 
+// MandatoryMiddlewareSpecs returns a map of middlewares available
```
ChuntaoLu: The comment of this method should be reworded to distinguish from `MiddlewareSpecs` method

README.md
```diff
@@ -128,6 +128,7 @@ example-gateway                 # root directory
 │   ├── endpoints               # generated mocks and module initializers for endpoints
 │   ├── gen-code                # generated structs and (de)serializers by Thrift compiler
 │   ├── middlewares             # generated module initializers for middlewares
+│   │   └── mandatory           # generated module initializers for mandatory middlewares
```
ChuntaoLu: Instead of `mandatory`, I think `default` is a better alternative, since it relates better with `DefaultDepDirs` field in the module class.

Also we should have a section in the doc to explain the concept of default dependencies and how it works.

codegen/runner/runner.go
```diff
@@ -92,15 +92,16 @@ func main() {
...
-		AnnotationPrefix:       config.MustGetString("annotationPrefix"),
-		GenCodePackage:         config.MustGetString("genCodePackage"),
-		CopyrightHeader:        string(copyright),
-		StagingReqHeader:       stagingReqHeader,
-		DeputyReqHeader:        deputyReqHeader,
-		TraceKey:               config.MustGetString("traceKey"),
+		RelThriftRootDir:         config.MustGetString("thriftRootDir"),
+		RelTargetGenDir:          config.MustGetString("targetGenDir"),
+		RelMiddlewareConfigDir:   config.MustGetString("middlewareConfig"),
+		RelMandatoryMidConfigDir: config.MustGetString("mandatoryMiddlewareConfig"),
```
ChuntaoLu: We should make such a list optional, i.e., an application does not necessarily have mandatory middlewares (or middlewares at all).

README.md
ChuntaoLu: The downside of this approach is that we lose the genericness of "default dependencies", it becomes a special case for middlewares/services.  Think about the case that we might want endpoints to have default clients.

Ryanfsdf: I think "default dependencies" can still remain generic because the `mandatory.yaml` file is **not** necessary for "default dependencies".

"Default dependencies" was introduced so that anything in the `DefaultDepDirs` field will be included in the module's `module/dependencies.go` file. The standard "default dependencies" feature does **not** need to know about the ordering of the individual dependencies.

`mandatory.yaml` is only required to add the middlewares to the endpoints' execution stack in a specific order. The `mandatory.yaml` "feature" is built **on top** of "default dependencies". 

We can still have endpoints that have default clients, since those do not need a `mandatory.yaml` file.

Let me know if this isn't clear and we can discuss more

ChuntaoLu: Well said, makes sense.

codegen/module.go
rpatali: either way, the comment doesn't appear to be adding much here.

README.md
Ryanfsdf: On second thought, this might not be a good idea. It's possible for to have more than one `service-config.yaml` file if there's more than one service in the `./services` folder. `example-gateway` has `echo-gateway` and `example-gateway` in its `./services` folder. If both of these services depend on the same endpoint, the ordering of the default middleware for this endpoint would be ambiguous since the list must come from either `example-gateway` or `echo-gateway`.

I'll keep the list as it is in a `default.yaml` file.

ChuntaoLu: Yes, multiple services are indeed a norm.

codegen/runner/runner.go
ChuntaoLu: Correct, we actually see problems with that and have to workout with that by having an empty dir. If you can fixed that along the way, that would be awesome!

Ryanfsdf: Fixed this. Had to change a test along the way (`template_test.go`) because the test relied on the default value of `middlewareConfig` being "./middlewares".

abhishekparwal:
default.yaml could be present in the services directory at a global level, rather than individual service level as the endpoint code is generated at that level.

Ryanfsdf:
> default.yaml could be present in the services directory at a global level, rather than individual service level as the endpoint code is generated at that level.

This means moving `default.yaml` from `./middlewares/default.yaml` to `./services/default.yaml`. I think this is good, I'll make the change

ChuntaoLu:
> > default.yaml could be present in the services directory at a global level, rather than individual service level as the endpoint code is generated at that level.
> 
> This means moving `default.yaml` from `./middlewares/default.yaml` to `./services/default.yaml`. I think this is good, I'll make the change

I don’t think this is a good idea, services don’t have direct dependencies on middlewares, having an arbitrary file in services/ demanding all endpoints depend on a list of default middlewares doesn’t make sense to me. The file is about the relationships between endpoints and middlewares, nothing to do with services directly.

Ryanfsdf:
> > > default.yaml could be present in the services directory at a global level, rather than individual service level as the endpoint code is generated at that level.
> > 
> > 
> > This means moving `default.yaml` from `./middlewares/default.yaml` to `./services/default.yaml`. I think this is good, I'll make the change
> 
> I don’t think this is a good idea, services don’t have direct dependencies on middlewares, having an arbitrary file in services/ demanding all endpoints depend on a list of default middlewares doesn’t make sense to me. The file is about the relationships between endpoints and middlewares, nothing to do with services directly.

I see your point. Also, I think the `default.yaml` file is easier to discover if it's right beside the `default` folder itself. I'll revert the change and fix the new merge conflicts

PR #554, opened 2019-02-06
Bump tchannel and unpin broken thrift

Ryanfsdf:
Updated tchannel-go from 1.5.0 to ^1.6.0. Hard pinning 1.5.0 was problematic because other projects (edge-gateway) require tchannel>=1.10.0 due to other dependencies.

The most logical fix would be bumping 1.5.0 to ^1.5.0 but 1.6.0 introduced a change to the number of logs outputted to `zap.logger.info` which breaks some tests in Zanzibar. Thus I changed 1.5.0 to ^1.6.0 and fixed the tests which were expecting a certain number of `zap.logger.info` output.

Also, I removed the pin for apache/thrift because version 0.12.0 introduces a breaking change and the pin ^0.10.0 imports 0.12.0, overriding tchannel's pin of '0.9.3>=, < 0.11.0' and introducing a breaking change.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -14,10 +14,6 @@ import:
   version: ^0.8.0
 - package: go.uber.org/atomic
   version: ^1.3.0
-- package: github.com/apache/thrift
-  version: ^0.10.0
-  subpackages:
-  - lib/go/thrift
 - package: github.com/uber-go/tally
   version: ^3
 # update to master since the last semver release is 2 years old
@@ -28,7 +24,7 @@ import:
 - package: go.uber.org/thriftrw
   version: v1.14.0
 - package: github.com/uber/tchannel-go
-  version: 1.5.0
+  version: ^1.6.0
 - package: github.com/opentracing/opentracing-go
   version: ^1.0.0
 - package: github.com/uber/jaeger-client-go
--- a/glide.yaml
+++ b/glide.yaml
@@ -24,7 +24,7 @@ import:
 - package: go.uber.org/thriftrw
   version: v1.14.0
 - package: github.com/uber/tchannel-go
-  version: ^1.6.0
+  version: ^1.5.0
 - package: github.com/opentracing/opentracing-go
   version: ^1.0.0
 - package: github.com/uber/jaeger-client-go
--- a/runtime/client_http_request_test.go
+++ b/runtime/client_http_request_test.go
@@ -192,7 +192,7 @@ func TestBarClientWithoutHeaders(t *testing.T) {
 
 	logs := gateway.AllLogs()
 
-	assert.Equal(t, 1, len(logs))
+	assert.Equal(t, 3, len(logs))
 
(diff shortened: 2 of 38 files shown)


glide.yaml
```diff
@@ -28,7 +24,7 @@ import:
 - package: go.uber.org/thriftrw
   version: v1.14.0
 - package: github.com/uber/tchannel-go
-  version: 1.5.0
+  version: ^1.6.0
```
abhishekparwal: comment here, thrift is coming tchannel-go

test/bootstrap_test.go
```diff
@@ -59,18 +57,9 @@ func TestBootstrapError(t *testing.T) {
 
 	switch err := errors.Cause(err).(type) {
 	case *testGateway.MalformedStdoutError:
-		var lineStruct struct {
-			Msg   string
-			Error string
-		}
-		jsonErr := json.Unmarshal([]byte(err.StdoutLine), &lineStruct)
```
Ryanfsdf: This expected log output to be a valid json. ie,
`{"msg":...., "err:..."}.

However, tchannel produces an extra logger.info line and the log output changes to
'{"msg":..., "err:..."}
{"msg":..., "err:..."}`
which is not valid json. I changed the test to just check for substring because that is sufficient for this purpose

runtime/client_http_request_test.go
```diff
@@ -192,7 +192,7 @@ func TestBarClientWithoutHeaders(t *testing.T) {
 
 	logs := gateway.AllLogs()
 
-	assert.Equal(t, 1, len(logs))
+	assert.Equal(t, 3, len(logs))
```
abhishekparwal: we should consider putting a filter in all logs to see only logs by zanzibar. this count doesn't achieve much.

Ryanfsdf: These `assert.Equal(t, 1, len(logs))` tests aren't very good tests because they check for library log output. If any dependent libraries emit extra log lines (which they are allowed to do at any time), these tests would break. Should I just remove them?

ChuntaoLu: I agree with you both. We should remove the number checking tests.

glide.yaml
Ryanfsdf: Tally and Tchannel are the only dependencies which use apache/thrift and both pin apache/thrift at '>=0.9.3, <0.11.0'. Since zanzibar is not directly using apache/thrift, zanzibar has no reason to know about apache/thrift

runtime/client_http_request_test.go
Ryanfsdf: I think removing these tests should be sufficient since they don't achieve much. Creating a filter would be difficult because we currently have no way of knowing where the logs came from

Ryanfsdf: A bonus from removing these tests is that we can pin tchannel at ^1.5.0 instead of ^1.6.0

PR #555, opened 2019-02-06
Fix workflow mock generation in case of nonstandard workflow path

tahmidhasan:
Non standard workflows cannot have mocks because the mock generation uses module package location as import path.

### DIFF
--- a/codegen/templates/workflow_mock.tmpl
+++ b/codegen/templates/workflow_mock.tmpl
@@ -31,7 +31,7 @@ import (
 	{{end -}}
 	{{end -}}
 	{{end -}}
-	{{$instance.PackageInfo.PackageAlias}} "{{$instance.PackageInfo.PackagePath}}"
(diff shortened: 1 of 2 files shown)


codegen/templates/workflow_mock.tmpl
```diff
@@ -31,7 +31,7 @@ import (
 	{{end -}}
 	{{end -}}
 	{{end -}}
-	{{$instance.PackageInfo.PackageAlias}} "{{$instance.PackageInfo.PackagePath}}"
+	{{$instance.PackageInfo.PackageAlias}} "{{$espec.WorkflowImportPath}}"
```
jacobgreenleaf: nit: This would be confusing because `PackageAlias` usually always refers to the import path for `PackagePath`. Maybe call it `{{$instance.PackageInfo.PackageAlias}}workflow` to disambiguate it?

tahmidhasan: i am scared of making a larger change so i will do it.... later

PR #557, opened 2019-02-08
Fix moduleSearchPath dependency ordering

Ryanfsdf:
I was working on updating `moduleSearchPaths` from a string to a dictionary when I discovered a bug with `moduleSearchPaths`.

Switching
```
moduleSearchPaths:
-clients/*
-middlewares/*
```
to 
```
moduleSearchPaths:
-middlewares/*
-clients/*
```
resulted in code generation failure. This is because the new `moduleSearchPaths` code **ignores** class dependency ordering. It was working fine right now because `clients/*` was **coincidentally** put before `middlewares/*`. However, switching the ordering made it so that middlewares, which are dependent on clients, were discovered first. Since middlewares are dependent on clients, clients **must** be discovered before middleware.

Thus, I've labeled this update as a "Fix" rather than an "Update"

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -51,7 +51,7 @@ const jsonConfigSuffix = "-config.json"
 const yamlConfigSuffix = "-config.yaml"
 
 // NewModuleSystem returns a new module system
-func NewModuleSystem(moduleSearchPaths []string, postGenHook ...PostGenHook) *ModuleSystem {
+func NewModuleSystem(moduleSearchPaths map[string][]string, postGenHook ...PostGenHook) *ModuleSystem {
 	return &ModuleSystem{
 		classes:           map[string]*ModuleClass{},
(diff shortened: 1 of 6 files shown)


codegen/module.go
```diff
@@ -557,93 +557,87 @@ func (system *ModuleSystem) ResolveModules(
...
 
-		// Resolve dependencies for all classes
-		resolveErr := system.populateResolvedDependencies(
-			resolvedModules[className],
-			resolvedModules,
-		)
-		if resolveErr != nil {
-			return nil, resolveErr
-		}
+			// Resolve dependencies for all classes
```
ChuntaoLu: Are these tabs? gofmt didn't taken care of it?

Ryanfsdf: Yes they're tabs. They look improperly indented in this pull request but they're properly indented in code.

PR #558, opened 2019-02-09
Support for flexible default middleware

Ryanfsdf:
Introduced support for default middlewares and also the concept of "default dependencies". Default dependencies support glob patterns similar to the recent "flexible directories" feature.


Ryanfsdf:
@ChuntaoLu I know in the last pull request, you mentioned that JSON is no longer supported and we should remove the check for json files in general. However, edge-gateway still uses json files so I've kept the json file checks for compatibility.

ChuntaoLu:
Is this PR supposed to replace https://github.com/uber/zanzibar/pull/553?

Ryanfsdf:
> Is this PR supposed to replace #553?


Yes. Although a large portion of this pull request is the same as that pull request, I've had to rework quite a bit of the logic in this one to support glob patterns. I'll close the old pull request

ChuntaoLu:
> > Is this PR supposed to replace #553?
> 
> Yes. Although a large portion of this pull request is the same as that pull request, I've had to rework quite a bit of the logic in this one to support glob patterns. I'll close the old pull request.
> 
> Hopefully this is the last pull request related to default middlewares (unless there's another big merge conflict, but I don't see any problems with any branches right now)

New PRs makes it hard on review side because the progression/context in the old PR is lost. Ideally we should work on the old PR but I understand in this case because the merge conflict is annoying and hard to manage too.

Ryanfsdf:
> > > Is this PR supposed to replace #553?
> > 
> > 
> > Yes. Although a large portion of this pull request is the same as that pull request, I've had to rework quite a bit of the logic in this one to support glob patterns. I'll close the old pull request.
> > Hopefully this is the last pull request related to default middlewares (unless there's another big merge conflict, but I don't see any problems with any branches right now)
> 
> New PRs makes it hard on review side because the progression/context in the old PR is lost. Ideally we should work on the old PR but I understand in this case because the merge conflict is annoying and hard to manage too.

I'll keep note of that for future PRs. Thanks a lot for the review!

PR #559, opened 2019-02-12
Add field accessors

Ryanfsdf:
Added field accessors to the http server


runtime/server_http_request.go
```diff
@@ -661,6 +661,16 @@ func (req *ServerHTTPRequest) GetSpan() opentracing.Span {
 	return req.span
 }
 
+// GetHTTPRequest returns the underlying http request
+func (req *ServerHTTPRequest) GetHTTPRequest() *http.Request {
```
ChuntaoLu: https://golang.org/doc/effective_go.html#Getters, the Go convention is to not have 'Get' in getters.

Ryanfsdf: Got it, good to know.

Thanks for the review!

PR #561, opened 2019-02-12
http template to be able to handle optional fields in path params

abhishekparwal:
…h params


abhishekparwal:
@jacobgreenleaf / @ChuntaoLu please review

PR #562, opened 2019-02-12
Add url pattern field

Ryanfsdf:
A middleware needs access to rawURLPattern for logging purposes. Included this field in ServerHTTPRequest

### DIFF
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -78,10 +78,10 @@ type HandlerFn func(
 // RouterEndpoint struct represents an endpoint that can be registered
 // into the router itself.
 type RouterEndpoint struct {
-	EndpointName string
-	HandlerName  string
-	RawPattern   string
-	HandlerFn    HandlerFn
+	EndpointName  string
+	HandlerName   string
+	RawURLPattern string
+	HandlerFn     HandlerFn
 
 	contextExtractor ContextExtractor
 	logger           *zap.Logger
@@ -95,13 +95,13 @@ func NewRouterEndpoint(
 	deps *DefaultDependencies,
 	endpointID string,
 	handlerID string,
-	rawPattern string,
+	rawURLPattern string,
 	handler HandlerFn,
 ) *RouterEndpoint {
 	return &RouterEndpoint{
 		EndpointName:     endpointID,
 		HandlerName:      handlerID,
-		RawPattern:       rawPattern,
+		RawURLPattern:    rawURLPattern,
 		HandlerFn:        handler,
 		contextExtractor: extractor,
 		logger:           deps.Logger,
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -80,6 +80,7 @@ type HandlerFn func(
 type RouterEndpoint struct {
 	EndpointName string
 	HandlerName  string
+	RawPattern   string
 	HandlerFn    HandlerFn
 
 	contextExtractor ContextExtractor
@@ -94,11 +95,13 @@ func NewRouterEndpoint(
 	deps *DefaultDependencies,
 	endpointID string,
 	handlerID string,
+	rawPattern string,
 	handler HandlerFn,
 ) *RouterEndpoint {
 	return &RouterEndpoint{
 		EndpointName:     endpointID,
 		HandlerName:      handlerID,
+		RawPattern:       rawPattern,
 		HandlerFn:        handler,
 		contextExtractor: extractor,
 		logger:           deps.Logger,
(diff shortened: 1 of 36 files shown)


Ryanfsdf:
Hi @ChuntaoLu @jacobgreenleaf Could I get this and #559 reviewed? A middleware in edge gateway needs access to these fields for logging purposes. Thank you!

runtime/server_http_request_test.go
```diff
@@ -136,7 +136,7 @@ func TestDoubleParseQueryValues(t *testing.T) {
 		"GET", "/foo", http.HandlerFunc(zanzibar.NewRouterEndpoint(
 			bgateway.ActualGateway.ContextExtractor,
 			deps,
-			"foo", "foo",
+			"foo", "foo", "/foo?%gh&%ij",
```
ChuntaoLu: I understand what this test for the specific url, but it is wired to say "this endpoint handles requests with the url that has the specific query values". Usually we only specify the url path, i.e., without the ? part.

runtime/router.go
```diff
@@ -78,9 +78,10 @@ type HandlerFn func(
 // RouterEndpoint struct represents an endpoint that can be registered
 // into the router itself.
 type RouterEndpoint struct {
-	EndpointName string
-	HandlerName  string
-	HandlerFn    HandlerFn
+	EndpointName  string
+	HandlerName   string
+	RawURLPattern string
```
ChuntaoLu: This means an endpoint is always associated with a url, what would be the answers to following questions:
- what is the expectation for below?
```
router.Handle("GET", "/foo",  http.HandlerFunc(zanzibar.NewRouterEndpoint(a, b, "foo", "foo", "/bar", c)))
```
it uses an endpoint that is for `/bar` to handle `/foo`, seems error-prone and shouldn't be allowed?
- previously I can use the same handler to handle different urls, now I will have to create one endpoint per url?

Overall, I don't think it is a good idea to couple the association between an url and an endpoint with the endpoint itself.

jacobgreenleaf:
> A middleware needs access to rawURLPattern for logging purposes. Included this field in ServerHTTPRequest

Can you describe the purpose more? Would it be sufficient to log the URL path part when you receive a request rather than the pattern that the path matched?

runtime/server_http_request_test.go
Ryanfsdf: This was a mistake on my end. In its place is supposed to be just `"/foo"`

runtime/router.go
Ryanfsdf: I understand your points for concern.

However, I need a way to uniquely identify an endpoint in a form canonical to both edge gateway and rtapi. The two choices are the url or thrift method, since `EndpointName` and `HandlerName` are specific to edge gateway and not to rtapi. However, passing even the thrift method into `NewRouterEndpoint` would be coupling the association between a url and an endpoint with the endpoint itself because the thrift method is mapped 1-to-1 with the url.

I'll try to figure out a different way to uniquely identify an endpoint, but please let me know if you have any suggestions.

Again, thanks a lot for the review!

Ryanfsdf:
> > A middleware needs access to rawURLPattern for logging purposes. Included this field in ServerHTTPRequest
> 
> Can you describe the purpose more? Would it be sufficient to log the URL path part when you receive a request rather than the pattern that the path matched?

URL path part is sufficient. However, I need the path part **before** the path parameters have been filled in. For example,  `/test/:uuid/example` vs `/test/12345/example`. `ServerHTTPRequest.Url` contains `/test/12345/example` but not `/test/:uuid/example`, which is what I need.

In terms of purpose, I need a way to uniquely identify an endpoint in a form which is canonical to both edge gateway and rtapi, and the url seemed like a good candidate.

jacobgreenleaf:
> > > A middleware needs access to rawURLPattern for logging purposes. Included this field in ServerHTTPRequest
> > 
> > 
> > Can you describe the purpose more? Would it be sufficient to log the URL path part when you receive a request rather than the pattern that the path matched?
> 
> URL path part is sufficient. However, I need the path part **before** the path parameters have been filled in. For example, `/test/:uuid/example` vs `/test/12345/example`. `ServerHTTPRequest.Url` contains `/test/12345/example` but not `/test/:uuid/example`, which is what I need.
> 
> In terms of purpose, I need a way to uniquely identify an endpoint in a form which is canonical to both edge gateway and rtapi, and the url seemed like a good candidate.

Shouldn't the URL after the path parameters are filled in also be sufficient? After all when the router receives the URL it will be with the path parameters filled in. The router will only match the URL with one pattern. Would like to hear more about the use case.

PR #563, opened 2019-02-13
Remove incremental codegen dependency resolution

jacobgreenleaf:


ChuntaoLu:
lgtm (I think the linter that dislikes Fprintf because the old test stuff are very picky about stdout/stderr?)

PR #564, opened 2019-02-14
Edwin.interface

edwinzg:
Update templates since the extra wrapper around tasks is unnecessary now


jacobgreenleaf:
I am not sure adding a special case for `"task"` is the best way to do this, would like to sync offline to see the motivation here.

ChuntaoLu:
I agree with @jacobgreenleaf, task isn't a builtin module type and should be visible here.

PR #565, opened 2019-02-14
Add support for multiple exceptions with the same status code

gjtrowbridge:


codegen/module.go
```diff
@@ -610,6 +610,7 @@ func (system *ModuleSystem) ResolveModules(
 
 				classConfigPath, _, _ := getConfigFilePath(moduleDirAbs, className)
 				if classConfigPath == "" {
+					fmt.Printf("  no class config found in %s directory, skipping over it\n", moduleDirAbs)
```
jacobgreenleaf: nit: `"warning: no class config found in %s directory\n"`

runtime/client_http_response.go
```diff
@@ -109,6 +110,35 @@ func (res *ClientHTTPResponse) ReadAndUnmarshalBody(v interface{}) error {
...
+// ReadAndUnmarshalBodyMultipleOptions will try to unmarshal non pointer value to one of the provided types or fail
+// It will return the deserialized struct (if any) that succeeded
+func (res *ClientHTTPResponse) ReadAndUnmarshalBodyMultipleOptions(vs []interface{}) (interface{}, error) {
+	rawBody, err := res.ReadAll()
+	if err != nil {
+		/* coverage ignore next line */
+		return nil, err
+	}
+
+	errMessages := []string{}
```
jacobgreenleaf: You should use `go.uber.org/multierr` for this.

gjtrowbridge: While I can certainly make this change, I'm not sure this is actually desirable, since it gives less control over how the error is collated and displayed.  IMO this leads to an easier-to-understand error message vs. relying on the internal collating / `Error` method of a `multierr` object.

I don't feel *that* strongly, so if you feel strongly I'll change it, but wanted to call this out.

jacobgreenleaf: I don't see that you lose control over the error display since `multierr.Errors()` returns the underlying errors. Is there any reason you chose this particular error format (`a | b | c`) as opposed to the one provided by `multierr`?

gjtrowbridge: I guess my point is that it doesn't feel like `multierr` provides a lot of additional value here, and IMO makes the code slightly harder to read.

Everyone knows what `append` does, not everyone will know all the interface methods on `multierr`.  I think that's fine, but given that the result is mostly the same (and takes the same number of lines) in both cases, I slightly prefer the non-multierr approach bc I consider it a bit simpler / easier to read / slightly more flexible.

I really don't feel strongly though, so I'll go ahead and update it to use `multierr` -- sounds like that's more standard for this sort of handling.

PR #566, opened 2019-02-15
Add accessor and fix linter issue

Ryanfsdf:
Created an accessor for response headers (No other accessors are needed. Found workarounds for them).

Linter was checking for Fprintf and throwing an error so I had to change it to Printf.

### DIFF
--- a/codegen/runner/runner.go
+++ b/codegen/runner/runner.go
@@ -38,11 +38,11 @@ type stackTracer interface {
 
 func checkError(err error, message string) {
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "%s:\n%s\n", message, err)
+		fmt.Printf("%s:\n%s\n", message, err)
 
 		causeErr := errors.Cause(err)
 		if causeErr, ok := causeErr.(stackTracer); ok {
-			fmt.Fprintf(os.Stderr, "%+v \n", causeErr.StackTrace())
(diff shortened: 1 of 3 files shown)


codegen/runner/runner.go
```diff
@@ -38,11 +38,11 @@ type stackTracer interface {
 
 func checkError(err error, message string) {
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "%s:\n%s\n", message, err)
+		fmt.Printf("%s:\n%s\n", message, err)
```
abhishekparwal: I know you did this to pass linter, but what you need is on error print it to stderr not stdout

abhishekparwal:
LGTM

runtime/server_http_response.go
```diff
@@ -281,3 +281,8 @@ func (res *ServerHTTPResponse) writeBytes(bytes []byte) {
 func (res *ServerHTTPResponse) GetPendingResponse() ([]byte, int) {
 	return res.pendingBodyBytes, res.pendingStatusCode
 }
+
+// ResponseHeaders returns the underlying http response's headers
+func (res *ServerHTTPResponse) ResponseHeaders() http.Header {
```
ChuntaoLu: nitpick: `Headers` should be sufficient since the method in defined on response.

PR #574, opened 2019-02-27
Provide solution to filter log fields by user demand

ChuntaoLu:
This PR does a few things:
- log endpoint request headers with user-provided filter function
- inherent request uuid if it is present in upstream request
- propogate endpoint request uuid to client request uuid
- rename a few log fields to avoid ambiguity

### DIFF
--- a/examples/example-gateway/config/production.json
+++ b/examples/example-gateway/config/production.json
@@ -50,6 +50,7 @@
 	"clients.withexceptions.errorPercentThreshold": 20,
 	"envVarsToTagInRootScope": [],
 	"http.port": 7783,
+	"http.clients.requestUUIDHeaderKey": "x-request-uuid",
 	"logger.fileName": "/var/log/example-gateway/example-gateway.log",
 	"logger.output": "disk",
 	"metrics.serviceName": "example-gateway",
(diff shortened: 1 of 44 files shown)


examples/example-gateway/config/production.json
```diff
@@ -65,5 +66,7 @@
 	"tchannel.port": 7784,
 	"tchannel.processName": "example-gateway",
 	"tchannel.serviceName": "example-gateway",
-	"useDatacenter": false
+	"tchannel.clients.requestUUIDHeaderKey": "x-request-uuid",
```
edwinzg: I wonder if we can just use the same requestUUIDHeader for tchannel and http

runtime/context.go
```diff
@@ -114,18 +115,17 @@ func GetEndpointRequestHeadersFromCtx(ctx context.Context) map[string]string {
 }
 
 // withRequestUUID returns a context with request uuid.
-func withRequestUUID(ctx context.Context, reqUUID uuid.UUID) context.Context {
+func withRequestUUID(ctx context.Context, reqUUID string) context.Context {
```
edwinzg: why make this a string instead of a UUID?

ChuntaoLu: This is used for log field, which will be `zap.String`, keeping it `uuid.UUID` means useless conversion between UUID and string.

PR #576, opened 2019-03-05
Log endpoint response headers and client request/response headers

ChuntaoLu:


examples/example-gateway/endpoints/tchannel/baz/baz_call.go
```diff
@@ -59,6 +60,9 @@ func (w Workflow) Handle(
 	clientReqHeaders := make(map[string]string)
 	// passing endpoint reqHeaders to downstream client
 	for _, k := range reqHeaders.Keys() {
+		if textproto.CanonicalMIMEHeaderKey(k) == "X-Token" {
```
jacobgreenleaf: Where does this `X-Token` value come from?

ChuntaoLu: upstream, i.e. endpoint request

PR #579, opened 2019-03-21
Adding support for enum in endpoint and client for http GET in query params , list support backward compatible with rtapi

abhishekparwal:
Adding support for enum in endpoint and client for http GET in query params. List support was not backward compatible with rtapi wire contract, which is needed for backward compat, and hence fix for that.


runtime/client_http_request_test.go
```diff
@@ -321,6 +321,39 @@ func TestMakingClientCallWithThriftException(t *testing.T) {
...
+		func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(200)
+			_, _ = w.Write([]byte(`"test-response"`))
+		},
+	)
+
+	deps := bgateway.Dependencies.(*exampleGateway.DependenciesTree)
+	bClient := deps.Client.Bar
+
+	body, _, err := bClient.ListAndEnum(
```
ChuntaoLu: This test does not make sense:
1. the thrift change covers both endpoint and client, a request should be made to the endpoint not the client
2. there is no http query in this test, the code that does `http query -> enum type` is not tested at all
3. there is no change with http client runtime, it doesn't make sense to add tests here. The right place would be `thrift_test.go` and `examples/example-gateway/endpoints/bar`
4. a good example for writing better and easier integration tests is https://github.com/uber/zanzibar/blob/master/examples/example-gateway/endpoints/contacts/save_contacts_test.go#L15

examples/example-gateway/idl/clients/bar/bar.thrift
```diff
@@ -82,6 +88,18 @@ service Bar {
        zanzibar.http.path = "/bar/hello"
        zanzibar.http.status = "200"
     )
+
+    string listAndEnum (
+        1: required list<string> demoIds (rtapi.http.ref = "query.demoIds")
```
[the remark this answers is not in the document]
abhishekparwal: thanks!

runtime/client_http_request_test.go
abhishekparwal: this test is needed bcs we want to test client is sending the right params to the client backend. have added the params though in the request and checking in client backend. Also, added a endpoint test case which mocks client which is insufficient in testing whether client gen code works correctly. thus, two test cases are needed. Or an integration test which can start a client backend and then test e2e would also be good.

examples/example-gateway/idl/clients/bar/bar.thrift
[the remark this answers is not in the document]
ChuntaoLu: correction?

codegen/method.go
```diff
@@ -1204,9 +1210,6 @@ func (ms *MethodSpec) setWriteQueryParamStatements(
 		longQueryName := ms.getLongQueryName(field, thriftPrefix)
 		identifierName := CamelCase(longQueryName) + "Query"
 		_, isList := realType.(*compile.ListSpec)
-		if isList {
-			longQueryName = longQueryName + "[]"
```
ChuntaoLu: Let's update the doc https://github.com/uber/zanzibar/blob/master/docs/thrift.md#zanzibarhttpref 
> If the annotation is on a field of a list and that list is a method argument, the URL query name will be prefixed with the struct's field name plus "[]".

to reflect the change here.

PR #582, opened 2019-04-10
Add support for physical generated directory

jacobgreenleaf:
Decouple the directory in which the output code is generated from the
targetGenDir which is also used to form the generated package name.

This preserves the existing behavior of the deprecated `GenerateBuild` function to build everything, while moving the logic to decide what folder to output into of the core generation logic. 

Removed `serializedModuleTree[Path]` because it's no longer needed.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -43,8 +43,6 @@ const (
 	SingleModule moduleClassType = iota
 	// MultiModule defines a module class type with multiple nested directories
 	MultiModule moduleClassType = iota
-
-	serializedModuleTreePath = "zanzibar.tree"
 )
 
 const jsonConfigSuffix = "-config.json"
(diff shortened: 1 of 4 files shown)


codegen/module.go
```diff
@@ -1024,215 +1022,80 @@ func (system *ModuleSystem) getSpec(instance *ModuleInstance) (interface{}, bool
...
-
-// GenerateIncrementalBuild is like GenerateBuild but filtered to only the given
-// module instances.
-func (system *ModuleSystem) GenerateIncrementalBuild(
+// Build invokes the generator for a module instance and optionally writes the files to disk
+func (system *ModuleSystem) Build(
 	packageRoot string,
 	baseDirectory string,
+	physicalGenDir string,
 	targetGenDir string,
```
ChuntaoLu: Is this still necessary with the newly added `physicalGenDir`?

jacobgreenleaf: You're right it can be removed here, we used to call `ResolveModules` but now we don't.

PR #585, opened 2019-04-10
Split out post-gen hooks to incrementally buildable pieces

jacobgreenleaf:
Post-gen hooks currently run as a side effect of running the build on a particular module. With other build systems such as buck or bazel, post-gen hooks would need to run independently and be invokable independently. 

* Switch post-gen hooks from `func([]*ModuleInstance)` to `func(*ModuleInstance)` for easier migration to incremental builds. Instead of the generator figuring out what instances it wants to generate, the caller of the hook is responsible for running the hook in the correct circumstance. A helper struct called `SimpleHookFunc` is provided to simplify this. 
* Separate implementation of generator from code to parallelize them. The parallelization code is no longer needed when generating just a single mock. 
* Separate code that writes to disk from the generation code. This is needed for parity with #582 to change the physical directory when running codegen in e.g. buck.

### DIFF
--- a/codegen/post_gen_hooks.go
+++ b/codegen/post_gen_hooks.go
@@ -21,18 +21,14 @@
 package codegen
 
 import (
-	"fmt"
 	"os"
 	"path"
 	"path/filepath"
-	"runtime"
 	"strings"
-	"sync"
-	"sync/atomic"
 
 	"github.com/pkg/errors"
 	"gopkg.in/validator.v2"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


codegen/post_gen_hooks.go
```diff
@@ -64,26 +60,75 @@ func newMockableClient(raw []byte) (*mockableClient, error) {
...
+	files, err := h.Generator(instance)
+	if err != nil {
+		return err
+	}
+
+	outputDir := filepath.Join(h.h.CodeGenTargetPath(), instance.Directory, h.directorySuffix)
+	prettyDir, _ := filepath.Rel(h.h.configRoot, outputDir)
+
+	PrintGenLine(
+		"mock",
```
ChuntaoLu: Shouldn't this be passed in?

PR #587, opened 2019-04-15
Temporary solution to support running tests with bazel

ChuntaoLu:
This PR hard codes a temporary solution to support running tests with bazel where accessing to test config files is restricted.
It is obvious that this approach isn't ideal since it tailors to bazel's specific needs, therefore temporary.


ChuntaoLu:
> LGTM. Maybe worth putting [this url](https://docs.bazel.build/versions/master/test-encyclopedia.html) in the comment in case people wondering where TEST_SRCDIR and TEST_WORKSPACE come from.

Good tip, updated the PR with link in code comments.

PR #590, opened 2019-05-09
Support grpc client

ChuntaoLu:
This PR adds grpc client support via yarpc.

The integration is done by creating a yarpc client dispatcher object that manages all grpc clients and making it part of gateway object. The dispatcher is used as a default dependency to create grpc clients. And the dispatcher lifecycle methods are hooked into the gateway lifecycle.


ChuntaoLu:
Separate this PR into two stack PRs (https://github.com/uber/zanzibar/pull/591 https://github.com/uber/zanzibar/pull/592) to isolate the code gen noise.

PR #592, opened 2019-05-09
Support grpc client

ChuntaoLu:
This PR adds grpc client support via yarpc.

The integration is done by creating a yarpc client dispatcher object that manages all grpc clients and making it part of gateway object. The dispatcher is used as a default dependency to create grpc clients. And the dispatcher lifecycle methods are hooked into the gateway lifecycle.


runtime/gateway.go
```diff
@@ -739,6 +778,44 @@ func (gateway *Gateway) setupTChannel(config *StaticConfig) error {
...
+	clientServiceNameMapping := make(map[string]string)
+	config.MustGetStruct("grpc.clientServiceNameMapping", &clientServiceNameMapping)
+
+	// TODO: no op if there is no YARPC clients
+	outbounds := make(yarpc.Outbounds, len(clientServiceNameMapping))
+	for key, value := range clientServiceNameMapping {
+		outbounds[key] = transport.Outbounds{
+			ServiceName: value,
+			// TODO: this setup opens one TCP connection to the sidecar router per client
+			//       Is it better to share the same TCP connection among the clients?
```
AlexAPB: yup a single transport is intended to support several outbounds, this is what yarpcfx does :)

PR #593, opened 2019-05-09
Add default endpoint log fields to server request logger

mattgode:
This provides better observability for instances where the request's logger is used

### DIFF
--- a/runtime/server_http_request.go
+++ b/runtime/server_http_request.go
@@ -103,6 +103,7 @@ func NewServerHTTPRequest(
 	httpRequest := r.WithContext(ctx)
 
 	scope := endpoint.scope.Tagged(scopeTags)
+	logger := endpoint.logger.With(logFields...)
 
 	req := &ServerHTTPRequest{
 		httpRequest:  httpRequest,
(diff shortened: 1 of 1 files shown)


runtime/server_http_request.go
```diff
@@ -103,6 +103,7 @@ func NewServerHTTPRequest(
 	httpRequest := r.WithContext(ctx)
 
 	scope := endpoint.scope.Tagged(scopeTags)
+	logger := endpoint.logger.With(logFields...)
```
ChuntaoLu: the log fields are already put on the context so that they are preserved  when code exits the endpoint scope, the fields should have been logged at log site, are you observing missing log fields?

mattgode: All of the usages of the logger on the request object are failing to add these log fields. They do not use the context while logging and so we aren't getting the necessary fields. You'll note that this is doing the same thing we are doing for the scope (which is set up properly and has tags)

ChuntaoLu: Just took another look, that is the case indeed. Thanks for the fix!

PR #594, opened 2019-05-15
Handle multiple services in generated http and tchannel clients

mattgode:
(Changed from previous version to reflect changes)

In this diff, we add a new parameter which is the client's target endpoint. For a thrift-based http client, this would be in the form of `ServiceName::methodName`. This is to provide support in metrics and logging for thrift files that define multiple services. Clients for such thrift files have cases where the client method name cannot match the thrift method name. Consider the following service definitions in one thrift file for a single client:

```
service Foo {
    string fn1()
}

service Bar {
    string fn1()
}
```

Client method names cannot simply be the method name or we would have conflicting method names for the generated client struct. Thus, we must concatenate the service and method names to make this work. For example. `FooFn1` and `BarFn1`.


runtime/client_http_request.go
```diff
@@ -168,7 +175,7 @@ func (req *ClientHTTPRequest) WriteJSON(
 
 // Do will send the request out.
 func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
-	opName := fmt.Sprintf("%s.%s", req.ClientID, req.MethodName)
+	opName := fmt.Sprintf("%s.%s", req.ServiceName, req.MethodName)
```
ChuntaoLu: If we do change the client request struct, then `("%s.%s(%s)", req.ClientID, req.MethodName, req.ThriftServiceMethod)` might be a good alternative.

runtime/client_http_request.go
```diff
@@ -53,17 +54,23 @@ type ClientHTTPRequest struct {
 // NewClientHTTPRequest allocates a ClientHTTPRequest. The ctx parameter is the context associated with the outbound requests.
 func NewClientHTTPRequest(
 	ctx context.Context,
-	clientID, methodName string,
```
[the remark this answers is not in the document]
mattgode: I was hoping to get away with not making a breaking change and rely on backwards compatibility. I just went ahead and made a breaking change as you suggested and passed the (thrift) service method names as another argument.

runtime/context.go
```diff
@@ -64,18 +64,19 @@ const (
...
-	scopeTagHandler         = "handlerid"
-	scopeTagError           = "error"
-	scopeTagStatus          = "status"
-	scopeTagProtocol        = "protocol"
-	scopeTagHTTP            = "HTTP"
-	scopeTagTChannel        = "TChannel"
-	scopeTagsTargetService  = "targetservice"
-	scopeTagsTargetEndpoint = "targetendpoint"
+	scopeTagClientMethod         = "clientmethod"
+	scopeTagClientTargetEndpoint = "clienttargetendpoint"
```
ChuntaoLu: It is probably not necessary to introduce this new tag, `scopeTagsTargetEndpoint` ("targetendpoint") should be sufficient because of the existing contexts for the client operation. Also tchannel client and http client sharing this same tas make it easier when we query by tags.

codegen/templates/tchannel_client.tmpl
```diff
@@ -102,7 +102,7 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		{{$serviceMethod := printf "%s::%s" $svc.Name .Name -}}
 		{{$methodName := (title (index $exposedMethods $serviceMethod)) -}}
 			{{if $methodName -}}
-			"{{$serviceMethod}}": "{{$methodName}}",
+			"{{$serviceMethod}}": "{{title .Name}}",
```
ChuntaoLu: This should not be changed, the `$methodName` is supposed to be the Go method name of the client.

mattgode: For the Edge Gateway usage of this template, there was always the assumption that the client and method names were a 1-1 mapping. Multiple services in one thrift disallow this. After talking offline with Chuntao offline, it sounds like we are going to just scrap the tchannel part of this diff and make consumers of metrics and logs switch to other metrics to avoid breaking changes in metrics and logging here.

mattgode: cc @abhishekparwal FYI

ChuntaoLu: > For the Edge Gateway usage of this template, there was always the assumption that the client and method names were a 1-1 mapping. Multiple services in one thrift disallow this. After talking offline with Chuntao offline, it sounds like we are going to just scrap the tchannel part of this diff and make consumers of metrics and logs switch to other metrics to avoid breaking changes in metrics and logging here.

To clarify, consumers of the metrics/logs will still look at the same metrics/logs but different tag/log field for thrift service and method. The client method name has always meant to be the Go method name of the client.

runtime/http_client.go
```diff
@@ -75,17 +75,18 @@ func NewHTTPClientContext(
...
 ) *HTTPClient {
-	loggers := make(map[string]*zap.Logger, len(methodNames))
+	loggers := make(map[string]*zap.Logger, len(methodToTargetEndpoint))
 
-	for _, methodName := range methodNames {
+	for methodName, targetEndpointName := range methodToTargetEndpoint {
 		loggers[methodName] = logger.With(
 			zap.String("clientID", clientID),
 			zap.String("clientMethod", methodName),
+			zap.String("targetEndpoint", targetEndpointName),
```
ChuntaoLu:
```suggestion
			zap.String(logFieldClientThriftMethod, targetEndpointName),
```

runtime/http_client.go
```diff
@@ -75,17 +75,18 @@ func NewHTTPClientContext(
...
 	defaultHeaders map[string]string,
 	timeout time.Duration,
 ) *HTTPClient {
-	loggers := make(map[string]*zap.Logger, len(methodNames))
+	loggers := make(map[string]*zap.Logger, len(methodToTargetEndpoint))
 
-	for _, methodName := range methodNames {
+	for methodName, targetEndpointName := range methodToTargetEndpoint {
 		loggers[methodName] = logger.With(
 			zap.String("clientID", clientID),
```
ChuntaoLu:
```suggestion
			zap.String(logFieldClientID, clientID),
```

runtime/http_client.go
```diff
@@ -75,17 +75,18 @@ func NewHTTPClientContext(
...
 			zap.String("clientMethod", methodName),
```
ChuntaoLu:
```suggestion
			zap.String(logFieldClientMethod, methodName),
```

runtime/http_client.go
ChuntaoLu: This suggestion is to align with tchannel client log fields, https://github.com/uber/zanzibar/blob/master/runtime/tchannel_client.go#L115-L118

runtime/client_http_request_test.go
```diff
@@ -276,6 +280,7 @@ func TestMakingClientCallWithRespHeaders(t *testing.T) {
 		"statusCode":                       float64(200),
 		"clientMethod":                     "Normal",
 		"clientID":                         "bar",
+		"targetEndpoint":                   "Bar::normal",
```
ChuntaoLu:
```suggestion
		"clientThriftMethod":               "Bar::normal",
```

test/endpoints/bar/bar_metrics_test.go
```diff
@@ -229,6 +231,7 @@ func TestCallMetrics(t *testing.T) {
 		"statusCode":                     float64(200),
 		"clientID":                       "bar",
 		"clientMethod":                   "Normal",
+		"targetEndpoint":                 "Bar::normal",
```
ChuntaoLu:
```suggestion
		"clientThriftMethod":             "Bar::normal",
```

PR #595, opened 2019-05-17
Circuit breaker metrics grouping by name

abhishekparwal:
It is needed for easy grouping and aliasing on m3, as it doesn't have a groupByNode function.


runtime/plugins/m3_aggregator.go
```diff
@@ -62,25 +62,26 @@ func InitializeM3Collector(scope tally.Scope) *M3CollectorClient {
...
-		failuresPrefix:          name + ".circuitbreaker" + ".failures",
-		rejectsPrefix:           name + ".circuitbreaker" + ".rejects",
-		shortCircuitsPrefix:     name + ".circuitbreaker" + ".shortCircuits",
-		timeoutsPrefix:          name + ".circuitbreaker" + ".timeouts",
-		fallbackSuccessesPrefix: name + ".circuitbreaker" + ".fallbackSuccesses",
-		fallbackFailuresPrefix:  name + ".circuitbreaker" + ".fallbackFailures",
-		totalDurationPrefix:     name + ".circuitbreaker" + ".totalDuration",
-		runDurationPrefix:       name + ".circuitbreaker" + ".runDuration",
+		scope:                   m.scope.Tagged(tagMap),
+		attemptsPrefix:          "circuitbreaker" + ".attempts",
```
ChuntaoLu: seems like string concatenation is no necessary, should just be "circuitbreaker.attempts". Same below.

PR #597, opened 2019-06-07
Add ability to configure the log level of the gateway logger

mattgode:
The gateway logger is used to log mos requests and responses. In order to provide clients the ability to control the default log level for these logs, we add a new configuration value, `logger.level` and default to the previous `zap.InfoLevel` if no level is provided

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -544,7 +544,17 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
 func (gateway *Gateway) setupLogger(config *StaticConfig) error {
 	var output zapcore.WriteSyncer
 	logEncoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
+
 	logLevel := zap.InfoLevel
+	if config.ContainsKey("logger.level") {
+		levelString := config.MustGetString("logger.level")
+		var ok bool
+		logLevel, ok = levelMap[levelString]
+		if !ok {
+			return errors.Errorf("unknown log level for gateway logger: %s", logLevel)
+		}
+	}
+
 	tempLogger := zap.New(
 		zapcore.NewCore(
 			logEncoder,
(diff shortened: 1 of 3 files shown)


runtime/gateway.go
```diff
@@ -544,7 +544,17 @@ func (gateway *Gateway) setupMetrics(config *StaticConfig) (err error) {
...
 	var output zapcore.WriteSyncer
 	logEncoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
+
 	logLevel := zap.InfoLevel
+	if config.ContainsKey("logger.level") {
+		levelString := config.MustGetString("logger.level")
+		var ok bool
+		logLevel, ok = levelMap[levelString]
+		if !ok {
+			return errors.Errorf("unknown log level for gateway logger: %s", logLevel)
```
ChuntaoLu: `errors.Errorf("unknown log level config for gateway logger: %s", levelString)` would be more informative.

test/bootstrap_test.go
```diff
@@ -65,3 +65,14 @@ func TestBootstrapError(t *testing.T) {
...
 }
+
+func TestBootstrapWithBadLogLevel(t *testing.T) {
+	gateway, err := testGateway.CreateGateway(t, map[string]interface{}{
+		"logger.level": "invalid",
+	}, &testGateway.Options{
+		TestBinary:  util.DefaultMainFile("example-gateway"),
+		ConfigFiles: util.DefaultConfigFiles("example-gateway"),
+	})
+	assert.Error(t, err, "got bootstrap err")
```
ChuntaoLu: Shouldn't we assert that the error we get is unknown log level?

mattgode: Unfortunately the error message gets munged by the test gateway, so it's hard to assert the error message.

mattgode: Added some unit tests as well in another file.

runtime/gateway.go
mattgode: Yeah that was a mistake. Thanks for catching.

PR #600, opened 2019-06-21
new IncrementalBuild to support building only modules provided as arg

rpatali:
this is more or less restoring of the original implementation
@jacobgreenleaf removed in PR #563. We didn't have use case for this
code to exists but our build times are getting longer and we need to way
to reduce the build time.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -1022,18 +1022,50 @@ func (system *ModuleSystem) getSpec(instance *ModuleInstance) (interface{}, bool
 	return spec, true, nil
 }
 
-func (system *ModuleSystem) tryResolveIncrementalBuild(
+// getAllAncestors adds classInstance to classModules if any of the classModules is in the recursive dependency tree of the instance.
+func getAllAncestors(classInstance *ModuleInstance, classModules map[string]*ModuleInstance) {
+	for _, instance := range classModules {
+		classInstanceTransitives, ok := classInstance.RecursiveDependencies[instance.ClassName]
+		if !ok {
+			continue
+		}
+
+		for _, classInstanceDependency := range classInstanceTransitives {
+			if classInstanceDependency.InstanceName == instance.InstanceName && classInstanceDependency.ClassName == instance.ClassName {
+				// toBeBuiltInstance is in the recursive dependency tree of classInstance
+
+				classModules[classInstance.InstanceName] = classInstance
+				fmt.Printf(
+					"Need to generate %q %q %q because it transitively depends on %q %q %q\n",
+					classInstance.InstanceName,
+					classInstance.ClassName,
+					classInstance.ClassType,
+					instance.InstanceName,
+					instance.ClassName,
+					instance.ClassType,
+				)
+
+				break
+			}
+		}
+	}
+}
+
+// collectTransitiveDependencies walks up the tree of dependencies starting at the leaves provided by instances and
+// returns all visited nodes.
+func (system *ModuleSystem) collectTransitiveDependencies(
 	instances []ModuleDependency,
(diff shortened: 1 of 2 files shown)


codegen/module.go
```diff
@@ -1022,6 +1022,134 @@ func (system *ModuleSystem) getSpec(instance *ModuleInstance) (interface{}, bool
...
+					classInstance.ClassName,
+					classInstance.ClassType,
+				)
+				continue
+			}
+
+			toBeBuiltModules[classInstance.ClassName][classInstance.InstanceName] = classInstance
+		}
+
+		// Collect things of the same class that depend on us
```
jacobgreenleaf: This comment is inaccurate. You are collecting *all* things that depend (directly or indirectly) on something in `toBeBuiltModules`.

codegen/module.go
```diff
@@ -1022,6 +1022,134 @@ func (system *ModuleSystem) getSpec(instance *ModuleInstance) (interface{}, bool
...
+		for _, classInstance := range classInstances {
+			// classInstance needs to be built if any of the dependencies that are to be built is in the recursive
+			// dependency tree of classInstance.
+
+			for _, className := range system.classOrder {
```
jacobgreenleaf: You can simplify this by moving this loop into a method like `getAllAncestors(*ModuleInstance, map[string][]*ModuleInstance)`, maybe even defining it as a method on the `map[string][]*ModuleInstance`.

codegen/module.go
```diff
@@ -1022,6 +1022,134 @@ func (system *ModuleSystem) getSpec(instance *ModuleInstance) (interface{}, bool
 	return spec, true, nil
 }
 
+func (system *ModuleSystem) tryResolveIncrementalBuild(
+	instances []ModuleDependency,
+	resolvedModules map[string][]*ModuleInstance,
+) (map[string][]*ModuleInstance, error) {
+
+	toBeBuiltModules := make(map[string]map[string]*ModuleInstance, 0)
```
jacobgreenleaf: nit: rename `toBeBuiltModules` to `toBeBuiltModulesUnique` to note why it's a `map[string]map[string]*ModuleInstance`.

codegen/module.go
```diff
@@ -1022,6 +1022,134 @@ func (system *ModuleSystem) getSpec(instance *ModuleInstance) (interface{}, bool
...
+					"Skipping generation of %q %q class of type %q "+
+						"as not needed for incremental build\n",
+					classInstance.InstanceName,
+					classInstance.ClassName,
+					classInstance.ClassType,
+				)
+				continue
+			}
+
+			toBeBuiltModules[classInstance.ClassName][classInstance.InstanceName] = classInstance
```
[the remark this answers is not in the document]
rpatali: I can definitely move this statement in the `if` block of the inner `for` loop but i think i still want to break that loop when a match is found. `found` is just an easier way for me to print that message although that can be derived too, i prefer this.

codegen/module.go
```diff
@@ -1022,6 +1022,141 @@ func (system *ModuleSystem) getSpec(instance *ModuleInstance) (interface{}, bool
...
+	commitChange bool,
+) (map[string][]*ModuleInstance, error) {
+	if len(instances) == 0 {
+		fmt.Println("Skipping build since no module dependency is provided")
+		return make(map[string][]*ModuleInstance), nil
+	}
+
+	toBeBuiltModules, err := system.collectTransitiveDependencies(instances, resolvedModules)
+	if err != nil {
+		// if incrementalBuild fails, perform a full build.
```
rpatali: infof falling back

PR #605, opened 2019-07-02
Swap out julienschmidt/httprouter

ChuntaoLu:
This PR swaps out [julienschmidt/httprouter](https://github.com/julienschmidt/httprouter) with an inhouse router implementation that has similar interface to httprouter. The rationale is to overcome the limitations of https://github.com/julienschmidt/httprouter/issues/6 and https://github.com/julienschmidt/httprouter/issues/175.

### DIFF
--- a/runtime/router/router.go
+++ b/runtime/router/router.go
@@ -1,0 +1,152 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package router
+
+import (
+	"context"
+	"net/http"
+)
+
+// Router dispatches http requests to a registered http.Handler.
+// It implements a similar interface to the one in github.com/julienschmidt/httprouter,
+// the main differences are:
+// 1. this router does not treat "/a/:b" and "/a/b/c" as conflicts (https://github.com/julienschmidt/httprouter/issues/175)
+// 2. this router does not treat "/a/:b" and "/a/:c" as different routes (https://github.com/julienschmidt/httprouter/issues/6)
+// 3. this router does tno treat "/a" and "/a/" as different routes
+type Router struct {
+	tries map[string]*Trie
+
+	// If enabled, the router checks if another method is allowed for the
+	// current route, if the current request can not be routed.
+	// If this is the case, the request is answered with 'Method Not Allowed'
+	// and HTTP status code 405.
+	// If no other Method is allowed, the request is delegated to the NotFound
+	// handler.
+	HandleMethodNotAllowed bool
+
+	// Configurable http.Handler which is called when a request
+	// cannot be routed and HandleMethodNotAllowed is true.
+	// If it is not set, http.Error with http.StatusMethodNotAllowed is used.
+	// The "Allow" header with allowed request methods is set before the handler
+	// is called.
+	MethodNotAllowed http.Handler
+
+	// Configurable http.Handler which is called when no matching route is
+	// found. If it is not set, http.NotFound is used.
+	NotFound http.Handler
+
+	// Function to handle panics recovered from http handlers.
+	// It should be used to generate a error page and return the http error code
+	// 500 (Internal Server Error).
+	// The handler can be used to keep your server from crashing because of
+	// unrecovered panics.
+	PanicHandler func(http.ResponseWriter, *http.Request, interface{})
+
+	// TODO: (clu) maybe support OPTIONS
+}
+
+type paramsKey string
+
+// urlParamsKey is the request context key under which URL params are stored.
+const urlParamsKey = paramsKey("urlParamsKey")
+
+// ParamsFromContext pulls the URL parameters from a request context,
+// or returns nil if none are present.
+func ParamsFromContext(ctx context.Context) []Param {
+	p, _ := ctx.Value(urlParamsKey).([]Param)
+	return p
+}
+
+// Handle registers a http.Handler for given method and path.
+func (r *Router) Handle(method, path string, handler http.Handler) error {
+	if r.tries == nil {
+		r.tries = make(map[string]*Trie)
+	}
+
+	trie, ok := r.tries[method]
+	if !ok {
+		trie = NewTrie()
+		r.tries[method] = trie
+	}
+	return trie.Set(path, handler)
+}
+
+// ServeHTTP dispatches the request to a register handler to handle.
+func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if r.PanicHandler != nil {
+		defer func(w http.ResponseWriter, req *http.Request) {
+			if recovered := recover(); recovered != nil {
+				r.PanicHandler(w, req, recovered)
+			}
+		}(w, req)
+	}
+
+	reqPath := req.URL.Path
+	if trie, ok := r.tries[req.Method]; ok {
+		if handler, params, err := trie.Get(reqPath); err == nil {
+			ctx := context.WithValue(req.Context(), urlParamsKey, params)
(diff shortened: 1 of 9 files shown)


runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+			t.children = append(t.children, newNode)
+		}
+	}
+
+	return nil
+}
+
+func (t *tnode) get(path string) (http.Handler, []Param, error) {
+	m, n := len(t.key), len(path)
```
jacobgreenleaf: `m` -> `myLength`
`n` -> `pathLength`

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+	var params []Param
+
+	// find the longest matched prefix
+	var j, k int
```
jacobgreenleaf: `j` -> `myIndex`
`k` -> `pathIndex`

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+	for j < m && k < n {
+		if t.key[j] == ':' {
+			oj, ok := j+1, k
+			for j < m && t.key[j] != '/' {
+				j++
+			}
+			for k < n && path[k] != '/' {
+				k++
+			}
+			params = append(params, Param{t.key[oj:j], path[ok:k]})
```
jacobgreenleaf: nit: use continue rather than more blocks

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+
+func (t *tnode) get(path string) (http.Handler, []Param, error) {
+	m, n := len(t.key), len(path)
+	var params []Param
+
+	// find the longest matched prefix
+	var j, k int
+	for j < m && k < n {
+		if t.key[j] == ':' {
+			oj, ok := j+1, k
```
jacobgreenleaf: `// wildcard starts - match until next slash`

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+			for j < m && t.key[j] != '/' {
+				j++
+			}
+			for k < n && path[k] != '/' {
+				k++
+			}
+			params = append(params, Param{t.key[oj:j], path[ok:k]})
+		} else if path[k] == ':' { // necessary for conflict check used in set call
```
jacobgreenleaf: Maybe is unexpected that this would also be used in regular request path, so that this is a successful test:

```
func TestTrieMatch(t *testing.T) {
	tree := NewTrie()
	tests := []ts{
		{op: set, path: "/a", value: "foo"},
		{op: get, path: "/:a", expectedValue: "foo"},
```

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+			k++
+		} else {
+			break
+		}
+	}
+
+	if j < m {
+		// path matches up to node key's second to last character,
+		// the last char of node key is "*" and path is no shorter than longest matched prefix
+		if t.key[j:] == "*" && k < n {
```
jacobgreenleaf: I think this line implies that `*` must match at least one character since you didn't write `k <= n`. You might note that somewhere because conventionally the Kleene star means "zero or more", not "one or more".

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// longest matched prefix matches up to node key length and path length
+	if k == n {
+		if t.value != nil {
+			return t.value, params, nil
+		}
+		return nil, nil, errNotFound
+	}
+
+	// TODO: recursion to iteration for speed
```
jacobgreenleaf: Not sure recursion will be intrinsically faster. Golang has no recursion limit :)

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+		} else if t.key[j] == path[k] {
+			j++
+			k++
+		} else {
+			break
+		}
+	}
+
+	// conflicts caused by ":" is only possible when j == m
```
jacobgreenleaf: This comment could be clearer:

```
// Exit condition: We fully matched path against this node using a wildcard
```

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+	if j == m {
+		for _, c := range t.children {
+			if c.key[0] == path[k] || c.key[0] == ':' || path[k] == ':' {
```
jacobgreenleaf: Shouldn't this logic be covered by `c.get` already?

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+				if _, _, err := c.get(path[k:]); err == nil {
+					return errExist
```
jacobgreenleaf: You will probably want to print out the match that conflicted for diagnostic purposes.

runtime/router/trie.go
ChuntaoLu: yeah, I will pass in a flag to indicate whether to treat `:` as a pattern or not in query path.

runtime/router/trie.go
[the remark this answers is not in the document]
ChuntaoLu: https://github.com/uber/zanzibar/blob/133a174f56401d29507c168e51f02feacce18d85/runtime/router/trie.go#L85-L91 guards against registering `/a*`

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+	for i < l && t.key[i] == path[i] {
+		i++
+	}
+
+	// find index j, k in key and path to which they match,
+	// j and k is only useful to check if there is conflict,
+	// they are not where splits happen, splits happen at index i.
+	var j, k int
+	for j < m && k < n {
+		if t.key[j] == ':' || path[k] == ':' {
```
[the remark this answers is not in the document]
ChuntaoLu: Yeah, I was looking at this piece and think about the case, I will add a test for it.

runtime/router/trie.go
```diff
@@ -0,0 +1,293 @@
...
+	if m > n {
+		l = n
+	} else {
+		l = m
+	}
+	for i < l && t.key[i] == path[i] {
+		i++
+	}
+
+	// find index j, k in key and path to which they match,
```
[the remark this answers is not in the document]
ChuntaoLu: Done

runtime/router/trie.go
ChuntaoLu: It is, mostly thinking of minimizing the call stack. Removed this check with the updated code.

runtime/router/trie.go
ChuntaoLu: Done

runtime/router/trie.go
ChuntaoLu: Done

runtime/router/trie.go
ChuntaoLu: Introduced two new flags to keep track of whether the current colon is a pattern, now this case returns not found.

runtime/router/trie.go
ChuntaoLu: Updated `Set` comment to reflect that.

runtime/router/trie.go
ChuntaoLu: Recursion generally has larger call stack compared to iteration. That said, whether the difference is noticeable in our application is another story.

runtime/router/router.go
```diff
@@ -0,0 +1,153 @@
...
+	"net/http"
+	"sort"
+	"strings"
+)
+
+// Router dispatches http requests to a registered http.Handler.
+// It implements a similar interface to the one in github.com/julienschmidt/httprouter,
+// the main differences are:
+// 1. this router does not treat "/a/:b" and "/a/b/c" as conflicts (https://github.com/julienschmidt/httprouter/issues/175)
+// 2. this router does not treat "/a/:b" and "/a/:c" as different routes (https://github.com/julienschmidt/httprouter/issues/6)
```
jacobgreenleaf: nit: this router does not allow `/a/:b` and `/a/:c` to be registered at all.

runtime/router/router.go
```diff
@@ -0,0 +1,153 @@
...
+type Router struct {
+	tries map[string]*Trie
+
+	// If enabled, the router checks if another method is allowed for the
+	// current route, if the current request can not be routed.
+	// If this is the case, the request is answered with 'Method Not Allowed'
+	// and HTTP status code 405.
+	// If no other Method is allowed, the request is delegated to the NotFound
+	// handler.
+	HandleMethodNotAllowed bool
```
jacobgreenleaf: Do we need to support this? When this flag is false, then we return 404. When this flag is true, we return 405. What's wrong with always returning 405 and getting rid of this flag?

runtime/router/trie.go
```diff
@@ -0,0 +1,310 @@
...
+	if keyLength > pathLength {
+		shorterLength = pathLength
+	} else {
+		shorterLength = keyLength
+	}
+	for i < shorterLength && t.key[i] == path[i] {
+		i++
+	}
+
+	// Find the first character that differs between "path" and this node's key, if it exists.
```
jacobgreenleaf: This code is quite confusing. I am guessing you are trying to have not just O(n) on the length of `key` but that you only process each character of `key` and `path` once and allocate no additional memory other than the indices. 

I think it may be much more readable if you split the string into path components and processed the equality of path components in sequence. Splitting the string into path components will roughly double the total memory used to register routes, but it's probably a better trade-off because routes are only registered once within the application.

runtime/router/router.go
ChuntaoLu: The difference is that when both this flag and `MethodNotAllowed` is present, the request is handled by the `MethodNotAllowed` handler. We can totally get rid of this and always return 405, I don't feel strongly about it. How do you think?

runtime/router/trie.go
ChuntaoLu: The trie does not necessarily split node on path segments, so splitting the path into segments and comparing equality doesn't work.

Register `/ab/c`
trie: `[/ab/c]`

Register `/a`
trie: 
```
[/a]
  |
[b/c]
```

runtime/router/router.go
ChuntaoLu: Done.

ChuntaoLu:
Benchmark between zanzibar router and http router:
https://github.com/ChuntaoLu/gorouter-bench

PR #607, opened 2019-07-08
Add ability to rewrite http request body

argouber:
This is useful for middlewares that want to rewrite the body of the request inline, modifying portions of payload.
This change takes care of modifying content-length header accordingly.


runtime/server_http_request.go
```diff
@@ -651,6 +651,20 @@ func (req *ServerHTTPRequest) UnmarshalBody(
 	return true
 }
 
+// WriteBytesBody replace the request body
+// we only change things cached in this layer, not in the raw httpRequest
+// Assumption is that encoding does not change
+func (req *ServerHTTPRequest) WriteBytesBody(
```
ChuntaoLu: `ReplaceBodyWith` is probably a more descriptive name,  also the line breaking is unnecessary.

runtime/server_http_request.go
```diff
@@ -651,6 +651,20 @@ func (req *ServerHTTPRequest) UnmarshalBody(
 	return true
 }
 
+// WriteBytesBody replace the request body
```
ChuntaoLu: "ReplaceBodyWith replaces the raw request body with given body and updates the request  content-length header accordingly. 
This method is only supposed to be used in middlewares where request body needs to be modified. And the encoding of the body should stay the same."

runtime/server_http_request.go
```diff
@@ -651,6 +651,20 @@ func (req *ServerHTTPRequest) UnmarshalBody(
...
+// Assumption is that encoding does not change
+func (req *ServerHTTPRequest) WriteBytesBody(
+	body []byte) bool {
+	// Replace the cached body bytes and fix dependent header
+	req.rawBody = body
+	if _, ok := req.Header.Get("Content-Length"); ok {
+		req.Header.Set("Content-Length", strconv.Itoa(len(body)))
+	}
+
+	return true
```
ChuntaoLu: Seems like it always return `true`? Why do we need return a bool?

runtime/server_http_request.go
argouber: Looks god. Will change.

runtime/server_http_request.go
argouber: Bool was because other similar functions were doing so.  And my 1st implementation had an error return with false.  But looking at this, it makes no sense to return anything.  Will change.

runtime/server_http_request.go
argouber: agree

runtime/server_http_request.go
```diff
@@ -651,6 +651,19 @@ func (req *ServerHTTPRequest) UnmarshalBody(
...
+// This method is only supposed to be used in middlewares where request body needs to be modified.
+// The encoding of the body should stay the same.
+func (req *ServerHTTPRequest) ReplaceBody(body []byte) {
+	// Replace the cached body bytes and fix dependent header
+	req.rawBody = body
+	if _, ok := req.Header.Get("Content-Length"); ok {
+		req.Header.Set("Content-Length", strconv.Itoa(len(body)))
+	}
+
+	return
```
ChuntaoLu: nit: the `return` here is redundant

PR #611, opened 2019-07-18
Add more context in codegen panic message

argouber:
The panic messages due to codegen failures did not indicate which thrift structure was
the one with the issue.  This prevented easy debugging.  With this, we print out the
offending thrift structure's name so it can be tracked down.


argouber:
will fix the test failures

PR #612, opened 2019-07-18
Run postGenHook per build instead of per instance

ChuntaoLu:
This PR corrects the postGenHook invocation to be per build instead of per module instance.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -1153,6 +1153,12 @@ func (system *ModuleSystem) IncrementalBuild(
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


codegen/module.go
```diff
@@ -1153,6 +1153,12 @@ func (system *ModuleSystem) IncrementalBuild(
 		}
 	}
 
+	for i, hook := range system.postGenHook {
+		if err := hook(resolvedModules); err != nil {
```
rpatali: I think you want to `toBeBuildModules` so that postGenHook is run only on those for incrementalBuild.

PR #616, opened 2019-08-03
Allow query params to be used as a set argument to method

argouber:
With this change, we allow the query params to be marshaled into a set rather than a list as an argument to an endpoint method


argouber:
abandoning and cleaning up

PR #621, opened 2019-08-12
Remove timeouts for test server

jacobgreenleaf:
If we have hard coded timeouts for test server clients, then tests
using these clients are flaky. Remove the timeout to fall back to
default behavior which is no timeout.


jacobgreenleaf:
#615

PR #622, opened 2019-08-13
Fixing nil pointer dereference for optional fields

tejaswiagarwal:
Optional fields should check before dereferencing the header value


codegen/method.go
```diff
@@ -802,7 +802,8 @@ func (ms *MethodSpec) setClientRequestHeaderFields(
 					// Note header values are always string
 					headerNameValuePair = "headers[%q]= string(r%s)"
 				} else {
-					headerNameValuePair = "headers[%q]= string(*r%s)"
+					headerNameValuePair = "if r != nil {" +
```
rpatali: Not sure if this will prevent the panic. I think you need `*r%s != nil`

PR #623, opened 2019-08-14
Change success logging to be debug & fix unit tests

smb158:


runtime/client_http_request_test.go
```diff
@@ -249,50 +246,6 @@ func TestMakingClientCallWithRespHeaders(t *testing.T) {
...
-		"ts",
-		"hostname",
-		"pid",
-	}
-	for _, dynamicValue := range dynamicHeaders {
-		assert.Contains(t, logMsg, dynamicValue)
-		delete(logMsg, dynamicValue)
-	}
-
-	expectedValues := map[string]interface{}{
```
jacobgreenleaf: Why are you removing these tests?

PR #625, opened 2019-08-21
Argo/query params for all methods

argouber:
We want to allow query params for all HTTP methods


codegen/method.go
```diff
@@ -1153,7 +1162,7 @@ func (ms *MethodSpec) setWriteQueryParamStatements(
 			if field.Required {
 				statements.appendf("if r%s == nil {", longFieldName)
 				// TODO: generate correct number of nils...
-				statements.append("\treturn nil, nil, errors.New(")
+				statements.appendf("\treturn %snil, errors.New(", nilResultStr)
```
rpatali: I see why you are doing this but I think an if...else would be more readable here.

argouber: Since this is not in perf path, it is reasonable to do so and... I should remove the old TODO.

PR #631, opened 2019-08-30
Dynamic routing support based on rule matching

abhishekparwal:
Dynamic routing on matching channels for the same client based on the rule engine match whose input is below format.

Though rule engine is more general-purpose in nature, but manifestation for client needs this:

```
{
  ...
  "clients.{client-id}.alternates": {
    "routingConfigs": [
      ...
      {
        "headerName": "header-name-1",
        "headerValue": "header-value-1\\*",
        "serviceName": "service-1"
      }
      ...
    ],
    "servicesDetail": {
      ... 
      "service-1": {
        "ip": "a.b.c.d",
        "port": 1234
      }
      ...
    }
  }
  ...
}
```

This can be used to routing to call the alternate channels dynamically based on the dynamic set of header and its values.  Sample Request with “header-name-1” header passed to the server with value “header-value-145” would be routed through “service-1” based on the above config.

Headers would be dynamically whitelisted in the workflow to be passed to the client. ex. header-name-1


codegen/templates/tchannel_client.tmpl
```diff
@@ -166,9 +213,9 @@ func configureCicruitBreaker(deps *module.Dependencies, timeoutVal int) bool {
 		hystrix.ConfigureCommand("{{$clientID}}", hystrix.CommandConfig{
 			MaxConcurrentRequests:  maxConcurrentRequests,
 			ErrorPercentThreshold:  errorPercentThreshold,
-			SleepWindow:            sleepWindowInMilliseconds,
+			SleepWindow:			sleepWindowInMilliseconds,
```
abhishekparwal: it is properly aligned when viewing in ide

codegen/templates/tchannel_client.tmpl
```diff
@@ -70,22 +74,37 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
+			"port": 5437
+		  },
+		  "mpx-prism": {
+			"ip": "127.0.0.1",
+			"port": 12958
+		  }
+		}
+	  }
+	}*/
+	var re ruleengine.RuleEngine
```
jacobgreenleaf: `ruleengine` is a bit generic and not descriptive of how it's used. You might rename it as `peerchooser` or `peerselector`?

codegen/templates/tchannel_client.tmpl
```diff
@@ -70,22 +74,37 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
+	{
+	  "clients.rider-presentation.alternates": {
+		"routingConfigs": [
+		  {
+			"headerName": "x-api-environment",
+			"headerValue": "*",
+			"serviceName": "presentation-sandbox"
+		  },
+		  {
+			"headerName": "RTAPI-Container",
```
jacobgreenleaf: Please rename these.

examples/example-gateway/config/production.yaml
```diff
@@ -67,6 +67,17 @@ tchannel.processName: example-gateway
 tchannel.serviceName: example-gateway
 tchannel.clients.requestUUIDHeaderKey: x-request-uuid
 useDatacenter: false
-
+clients.baz.alternates: {"routingConfigs":[{"headerName":"x-api-environment","headerValue":"\\*",
```
jacobgreenleaf: Can you include YAML annotations on the struct as well so this can be defined as YAML?

runtime/ruleengine/rule_engine.go
```diff
@@ -0,0 +1,96 @@
...
+
+func convertRawRules(rawRule RawRule) rule {
+	r := rule{value: rawRule.Value}
+	for _, pat := range rawRule.Patterns {
+		r.patterns = append(r.patterns, regexp.MustCompile(pat))
+	}
+	return r
+}
+
+func (r *ruleEngine) GetValue(patternValues []string) (interface{}, bool) {
```
jacobgreenleaf: `// GetValue checks whether a rule matches all pattern values`

You might consider calling this `Match` instead?

runtime/ruleengine/rule_engine.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, rule := range r.rules {
+		matched := true
+		if len(rule.patterns) != len(patternValues) {
+			matched = false
+			continue
+		}
+
+		for i, pat := range rule.patterns {
+			if !r.match(pat, patternValues[i]) {
```
jacobgreenleaf: nit: This should just be `pat.MatchString(patternValues[i])`, `r.match` adds no value.

runtime/ruleengine/rule_engine_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRuleEngine(t *testing.T) {
+
+	rule1 := RawRule{
```
jacobgreenleaf: Please rename these

runtime/static_config.go
```diff
@@ -229,34 +230,44 @@ func (conf *StaticConfig) MustGetStruct(key string, ptr interface{}) {
...
 		panic(errors.Errorf("Cannot GetStruct (%s) into nil ptr", key))
 	}
-
 	if v, contains := conf.seedConfig[key]; contains {
 		rptr.Elem().Set(reflect.ValueOf(v))
 		return
 	}
 
 	if v, contains := conf.configValues[key]; contains {
-		if value, ok := v.(map[string]interface{}); ok {
```
jacobgreenleaf: Can we keep this as-is? We shouldn't be supporting JSON configuration anymore.

runtime/ruleengine/rule_engine.go
jacobgreenleaf: It might save you a bit of code to change this to be:

```
GetValue(values []string, defaultMatch interface{}) interface{}
```

This would remove any nil/false check at the caller.

runtime/ruleengine/rule_engine.go
```diff
@@ -0,0 +1,96 @@
...
+
+type rule struct {
+	patterns []*regexp.Regexp
+	value    interface{}
+}
+
+// RuleEngine provides a way to get a value if rule matched
+type RuleEngine interface {
+	// GetValue returns a value, true if rule matched, else returns a nil, false if no rule matched
+	GetValue(patternValues []string) (interface{}, bool)
```
jacobgreenleaf: Given that you only ever call this with a 2-tuple of header key, header value, maybe it's worth making this less generic and just using `GetValue(header, value string)`?

runtime/tchannel_client.go
```diff
@@ -290,3 +263,24 @@ func (c *TChannelClient) call(
 
 	return call.success, call.resHeaders, err
 }
+
+func (c *TChannelClient) getDynamicChannelWithFallback(reqHeaders map[string]string, sc *tchannel.SubChannel) *tchannel.SubChannel {
+	if c.ruleEngine != nil {
```
jacobgreenleaf: You should unroll these:

```
func (c *TChannelClient) getDynamicChannelWithFallback(...) (ch *tchannel.SubChannel)

ch = sc

if c.ruleEngine == nil {
    return
}

...
```

codegen/templates/tchannel_client.tmpl
```diff
@@ -116,14 +136,16 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
 		&zanzibar.TChannelClientOption{
-			ServiceName:          serviceName,
-			ClientID:             "{{$clientID}}",
-			MethodNames:          methodNames,
-			Timeout:              timeout,
-			TimeoutPerAttempt:    timeoutPerAttempt,
-			RoutingKey:           &routingKey,
-			AltSubchannelName:    scAltName,
+			ServiceName:		  serviceName,
+			ClientID:			 "{{$clientID}}",
```
smb158: nit: these used to be lined up and are now all over the place

scripts/cover.sh
```diff
@@ -128,7 +128,7 @@ cat ./coverage/istanbul.json | jq '[
 ] | from_entries' > ./coverage/istanbul-runtime.json
 
 echo "Checking code coverage for runtime folder"
-./node_modules/.bin/istanbul check-coverage --statements 100 \
+./node_modules/.bin/istanbul check-coverage --statements 99.92 \
```
smb158: do we want to lower this from 100?

abhishekparwal: there were cases in code that were not covered and these flows I didn't touch. But since I touched the tchannel_client.go file the build started failing. Also, those cases are erring cases in the workflow they are part of while ini'ting, which is not super important. Hence lowering the threshold on discussion with @jacobg.

codegen/templates/tchannel_client.tmpl
abhishekparwal: thanks, it was a tab verus space causing alignment issues. fixed.

PR #635, opened 2019-09-12
Set shardkey in tchannel client

smb158:


runtime/context.go
```diff
@@ -41,6 +41,7 @@ const (
 	endpointKey           = contextFieldKey("endpoint")
 	requestUUIDKey        = contextFieldKey("requestUUID")
 	routingDelegateKey    = contextFieldKey("rd")
+	shardKey              = contextFieldKey("sk")
```
jacobgreenleaf: Needs helper like `WithShardKey`

smb158: 👍

PR #636, opened 2019-09-13
Add functionality to mock client interface with custom interface names

tejaswiagarwal:
This adds ability to mock custom interface names. The config files provide a map of importPath to interface name which is then used to mock. If the importPath does not contain a key, then the default value "Client" is used to mock.


examples/example-gateway/build.yaml
```diff
@@ -4,6 +4,9 @@ copyrightHeader: ./copyright_header.txt
 endpointConfig: ./endpoints
 genCodePackage: github.com/uber/zanzibar/examples/example-gateway/build/gen-code
 genMock: true
+clientInterfacesMock:
+  github.com/uber/zanzibar/examples/example-gateway/build/clients/contacts:
```
abhishekparwal: keeping it local to the client-config is better. And then you don't have to give an import path twice. this configuration should be captured in client-config.yaml.

codegen/post_gen_hooks.go
```diff
@@ -141,8 +141,16 @@ func ClientMockGenHook(h *PackageHelper, t *Template) (PostGenHook, error) {
...
-					mock, err := bin.GenMock(importPath, "clientmock", clientInterface)
+					clientInterfacesMap := h.GetInterfacesMap()
+					var mock []byte
+					var err error
+					// generate mock client, this starts a sub process.
+					// if an interfaces name is provided use that, else use "Client"
+					if val, ok := clientInterfacesMap[importPath]; ok {
+						mock, err = bin.GenMock(importPath, "clientmock", val)
+					} else {
+						mock, err = bin.GenMock(importPath, "clientmock", clientInterface)
```
abhishekparwal: AugmentMockWithFixture also takes a clientInterface though unused. either remove it from there or pass it.

codegen/package.go
```diff
@@ -63,6 +63,8 @@ type PackageHelper struct {
 	moduleSearchPaths map[string][]string
 	// defaultDependencies is a dictionary of glob patterns for default dependencies
 	defaultDependencies map[string][]string
+	// clientinterfacesList is a list of interfaces to mock for a given client
+	clientInterfacesMock map[string][]string
```
abhishekparwal: nit : comment -> its a map and not a list.

examples/example-gateway/build.yaml
abhishekparwal: add a client with a different interface name to test if mockable and if a fallback is not being used.

codegen/post_gen_hooks.go
```diff
@@ -141,8 +141,16 @@ func ClientMockGenHook(h *PackageHelper, t *Template) (PostGenHook, error) {
...
 
-					// generate mock client, this starts a sub process
-					mock, err := bin.GenMock(importPath, "clientmock", clientInterface)
+					clientInterfacesMap := h.GetInterfacesMap()
+					var mock []byte
+					var err error
+					// generate mock client, this starts a sub process.
+					// if an interfaces name is provided use that, else use "Client"
+					if val, ok := clientInterfacesMap[importPath]; ok {
+						mock, err = bin.GenMock(importPath, "clientmock", val)
```
abhishekparwal: nit: don't need to repeat the entire statement,

rather do like: 
```
if val, ok := clientInterfacesMap[importPath]; ok {
        clientInterface = val
} 

mock, err = bin.GenMock(importPath, "clientmock", clientInterface)

```

codegen/package.go
tejaswiagarwal: Done

examples/example-gateway/build.yaml
```diff
@@ -4,6 +4,9 @@ copyrightHeader: ./copyright_header.txt
 endpointConfig: ./endpoints
 genCodePackage: github.com/uber/zanzibar/examples/example-gateway/build/gen-code
 genMock: true
+clientInterfacesMock:
```
abhishekparwal: looks like forgot to remove this.

codegen/post_gen_hooks.go
```diff
@@ -140,9 +144,18 @@ func ClientMockGenHook(h *PackageHelper, t *Template) (PostGenHook, error) {
 					genDir := filepath.Join(buildDir, instance.Directory, "mock-client")
 
 					importPath := importPathMap[key]
+					customInterface := customInterfaceMap[key]
+
+					var mock []byte
```
abhishekparwal: nit: inline declaration was better

codegen/post_gen_hooks.go
```diff
@@ -91,16 +93,18 @@ func ClientMockGenHook(h *PackageHelper, t *Template) (PostGenHook, error) {
...
 			}
 
 			importPathMap[key] = importPath
+			customInterfaceMap[key] = customInterface
 
 			// gather all modules that need to generate fixture types
 			f := client.Config.Fixture
 			if f != nil && f.Scenarios != nil {
-				pathSymbolMap[importPath] = clientInterface
+				pathSymbolMap[importPath] = defaulClientInterface
```
abhishekparwal: here also we need to pass customClientInterface if present

examples/example-gateway/clients/contacts/client-config.yaml
```diff
@@ -9,5 +9,6 @@ config:
       - success
   idlFile: clients/contacts/contacts.thrift
   idlFileSha: '{{placeholder}}'
+  customInterface: Client
```
abhishekparwal: Change the Interface name to CustomClient to test if this code here works or nor. otherwise, we won't be able to distinguish between default and custom case.

codegen/post_gen_hooks.go
```diff
@@ -140,9 +144,18 @@ func ClientMockGenHook(h *PackageHelper, t *Template) (PostGenHook, error) {
 					genDir := filepath.Join(buildDir, instance.Directory, "mock-client")
 
 					importPath := importPathMap[key]
+					customInterface := customInterfaceMap[key]
+
+					var mock []byte
+					var err error
+					var clientInterface = "Client"
```
abhishekparwal: defaultInterface is a already a const in the file use that. 

moreover, override it above and use it in pathSymbolMap and create a clientInterfaceMap rather than customInterfaceMap and provide the overriden client value.

examples/example-gateway/build.yaml
tejaswiagarwal: Done!

examples/example-gateway/clients/contacts/client-config.yaml
tejaswiagarwal: I tested this locally and it works. As in, CustomClient for /contacts/ will not work since contacts is not a custom client, if it was, it would be generating the mock correctly.

codegen/post_gen_hooks.go
tejaswiagarwal: Done

examples/example-gateway/clients/contacts/client-config.yaml
abhishekparwal: testing locally is never enough. It needs continuous testing through tests to know if there is any regression in future.

PR #640, opened 2019-09-18
Allow query params on all Rest methods

argouber:
We previously limited query params to only GET requests. There is a clear need for other body-less methods like DELETE to have query params.  Using this opportunity to enable query params for all methods.

- Among other things, a few bugs fixed along the way (delete should not use body).
- Unit test added.
- Explicitly tagged query params in unit-tests

The previous implementation (PR #625) had issues which we backed out.  This is the redone fix.


codegen/method.go
```diff
@@ -278,30 +282,6 @@ func NewMethod(
 	return method, nil
 }
 
-func (ms *MethodSpec) scanForNonParams(funcSpec *compile.FunctionSpec) bool {
```
argouber: This method is unused.  Removing dead code.

examples/example-gateway/idl/clients/bar/bar.thrift
```diff
@@ -179,6 +193,18 @@ service Bar {
         zanzibar.http.status = "200"
     )
 
+    BarResponse argWithQueryParams(
+        1: required string name	(zanzibar.http.ref = "query.name")
+        2: optional string userUUID	(zanzibar.http.ref = "query.userUUID")
+        3: optional list<string> foo	(zanzibar.http.ref = "query.foo")
+        4: required list<i8> bar	(zanzibar.http.ref = "query.bar")
+        // TODO(argouber): Actually allow sets - preferably with go.tag of slice
```
argouber: Removed the optional field that was not being set in the test.  We need a good way to specify set in the query parameter.  Will do this in a separate PR

test/clients/bar/bar_test.go
```diff
@@ -1049,13 +1049,6 @@ func TestDeleteFoo(t *testing.T) {
 	bgateway.HTTPBackends()["bar"].HandleFunc(
 		"DELETE", "/bar/foo",
 		func(w http.ResponseWriter, r *http.Request) {
```
argouber: Delete should not be using the body.

codegen/method.go
```diff
@@ -1127,16 +1108,44 @@ func getQueryEncodeExpression(typeSpec compile.TypeSpec, valueName string) strin
...
 	var hasQueryFields bool
-	var stack = []string{}
+	var stack []string
+	isVoidReturn := funcSpec.ResultSpec.ReturnType == nil
 
 	visitor := func(
 		goPrefix string, thriftPrefix string, field *compile.FieldSpec,
 	) bool {
+		// Skip if there are no query params in the field or its components
+		// However, if by definition there is no request body, untagged fields are mapped to query-params
```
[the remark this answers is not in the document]
argouber: The comment could be clearer ...
By definition => GET/DELETE do not have body.  The reset do ... 
And the code makes the assumption that if a body is to be expected, the unannotated fields are in the body.
If not, they are to be found in the query params.

PR #641, opened 2019-09-18
add pascal lowercasing method

rpatali:
used elsewhere as lib func.

### DIFF
--- a/codegen/casing_test.go
+++ b/codegen/casing_test.go
@@ -32,6 +32,9 @@ var fixtures = map[string]string{
 	"aBCID": "aBCID",
 	"xUUID": "xUUID",
 	"Uuid":  "uUID",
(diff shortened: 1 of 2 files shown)


codegen/casing_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/uber/zanzibar/codegen"
+)
+
+var fixtures = map[string]string{
+	"aBCID": "aBCID",
+	"xUUID": "xUUID",
+	"Uuid":  "uUID",
```
argouber: Technically your first two cases are identical I think ... and you may want to add a multiword or mixed case stuff

rpatali: makes sense. Adding a few more examples.

PR #650, opened 2019-09-26
Replace Timer metrics with Histogram as default

argouber:
Histograms give a better aggregation capability across DCs.  Switching to Histograms from Timers for the latency metrics.

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -303,7 +303,9 @@ type Logger interface {
 // ContextMetrics emit metrics with tags extracted from context.
 type ContextMetrics interface {
 	IncCounter(ctx context.Context, name string, value int64)
-	RecordTimer(ctx context.Context, name string, d time.Duration)
+	// @deprecated -- use Histogram instead
+	// RecordTimer(ctx context.Context, name string, d time.Duration)
+	RecordHistogramDuration(ctx context.Context, name string, d time.Duration)
 }
 
 type contextMetrics struct {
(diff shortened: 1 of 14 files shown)


runtime/context.go
```diff
@@ -319,12 +321,16 @@ func NewContextMetrics(scope tally.Scope) ContextMetrics {
...
+	c.scope.Tagged(GetScopeTagsFromCtx(ctx)).Counter(name).Inc(value)
 }
 
+// @deprecated
 // RecordTimer records the duration with current tags from context
-func (c *contextMetrics) RecordTimer(ctx context.Context, name string, d time.Duration) {
-	tags := GetScopeTagsFromCtx(ctx)
-	c.scope.Tagged(tags).Timer(name).Record(d)
+//func (c *contextMetrics) RecordTimer(ctx context.Context, name string, d time.Duration) {
+//	c.scope.Tagged(GetScopeTagsFromCtx(ctx)).Timer(name).Record(d)
```
rpatali: Not crazy about all the commented code. Can we just deleted all of these.

argouber: Done.

runtime/context.go
```diff
@@ -319,12 +319,10 @@ func NewContextMetrics(scope tally.Scope) ContextMetrics {
...
-	tags := GetScopeTagsFromCtx(ctx)
-	c.scope.Tagged(tags).Counter(name).Inc(value)
+	c.scope.Tagged(GetScopeTagsFromCtx(ctx)).Counter(name).Inc(value)
 }
 
-// RecordTimer records the duration with current tags from context
-func (c *contextMetrics) RecordTimer(ctx context.Context, name string, d time.Duration) {
-	tags := GetScopeTagsFromCtx(ctx)
-	c.scope.Tagged(tags).Timer(name).Record(d)
+// RecordHistogram records the duration with current tags from context
```
rpatali: method name doesn't match.

PR #652, opened 2019-09-30
support binary pass through response from clients

rpatali:
there are situations where an client response may not be able to conform
to a fixed schema as the payload may not be generated by a system that
is controlled by us. when talking to such a system, an artifical schema
may be created but its hard to keep it up to date causing other issues
hence its best to just allow binary pass through.

E.g., downstream is a 3rd party that returns a JSON blob and provides an
SDK to consume the JSON blob. In this case, the schema is negotiated
between the downstream 3rd party and the SDK provided by the 3rd party.

### DIFF
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -277,14 +277,12 @@ func (c *{{$clientName}}) {{$methodName}}(
 	switch res.StatusCode {
 		case {{.OKStatusCode.Code}}:
 		{{- if eq .ResponseType "[]byte"}}
-
 		responseBody, err := res.ReadAll()
 		if err != nil {
 			return defaultRes, respHeaders, err
 		}
 		return responseBody, respHeaders, nil
 		{{ else }}
-
 		var responseBody {{unref .ResponseType}}
 		err = res.ReadAndUnmarshalBody(&responseBody)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


codegen/templates/http_client.tmpl
```diff
@@ -276,17 +276,27 @@ func (c *{{$clientName}}) {{$methodName}}(
 	{{else if eq (len .Exceptions) 0}}
 	switch res.StatusCode {
 		case {{.OKStatusCode.Code}}:
-			var responseBody {{unref .ResponseType}}
-			err = res.ReadAndUnmarshalBody(&responseBody)
-			if err != nil {
-				return defaultRes, respHeaders, err
-			}
+		{{- if eq .ResponseType "[]byte"}}
+
```
argouber: nit: nix the extra blank line :)

codegen/templates/http_client.tmpl
```diff
@@ -324,17 +334,27 @@ func (c *{{$clientName}}) {{$methodName}}(
...
+		responseBody, err := res.ReadAll()
+		if err != nil {
+			return defaultRes, respHeaders, err
+		}
+		return responseBody, respHeaders, nil
+		{{ else }}
```
argouber: and this :)

PR #665, opened 2019-10-29
Add application/json too all requests by default

tejaswiagarwal:
This adds Accept:application/json headers to all requests if Accept header is not already set. We unmarshal on application/json so it's essential to pass this to the downstream.

### DIFF
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -78,6 +78,9 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		timeoutVal,
 	)
 	defaultHeaders := make(map[string]string)
+	if deps.Default.Config.ContainsKey("http.defaultHeaders") {
+		deps.Default.Config.MustGetStruct("http.defaultHeaders", &defaultHeaders)
+	}
 	if deps.Default.Config.ContainsKey("clients.{{$clientID}}.defaultHeaders") {
 		deps.Default.Config.MustGetStruct("clients.{{$clientID}}.defaultHeaders", &defaultHeaders)
 	}
--- a/examples/example-gateway/config/test.yaml
+++ b/examples/example-gateway/config/test.yaml
@@ -67,6 +67,9 @@ tchannel.port: 0
 tchannel.processName: example-gateway
 tchannel.serviceName: example-gateway
 tchannel.clients.requestUUIDHeaderKey: x-request-uuid
+http.defaultHeaders:
+  Accept: application/json
+  Content-Type: application/json
 useDatacenter: false
 clients.baz.alternates:
   routingConfigs:
--- a/examples/example-gateway/config/test.yaml
+++ b/examples/example-gateway/config/test.yaml
@@ -45,6 +45,9 @@ clients.withexceptions.maxConcurrentRequests: 1000
 clients.withexceptions.errorPercentThreshold: 20
 env: test
 envVarsToTagInRootScope: []
+http.defaultHeaders:
+  Accept: application/json
+  Content-Type: application/json
 http.port: 0
 http.clients.requestUUIDHeaderKey: x-request-uuid
 logger.fileName: /tmp/example-gateway.log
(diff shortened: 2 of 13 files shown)


runtime/client_http_request.go
```diff
@@ -160,14 +160,22 @@ func (req *ClientHTTPRequest) WriteJSON(
 		httpReq.Header.Set(headerKey, headerValue)
 	}
 
+	acceptTypePresent := false
 	for k := range headers {
```
rpatali: header is a map!

runtime/client_http_request.go
```diff
@@ -160,14 +160,22 @@ func (req *ClientHTTPRequest) WriteJSON(
...
 		httpReq.Header.Set(k, headers[k])
 	}
 
 	if body != nil {
 		httpReq.Header.Set("Content-Type", "application/json")
 	}
 
+	/* Only unmarshal JSON today so set this as default if none present */
+	if !acceptTypePresent {
+		httpReq.Header.Set("Accept", "application/json")
```
rpatali: declare all of these strings as const.

runtime/client_http_request.go
```diff
@@ -160,14 +160,22 @@ func (req *ClientHTTPRequest) WriteJSON(
 		httpReq.Header.Set(headerKey, headerValue)
 	}
 
+	acceptTypePresent := false
 	for k := range headers {
+		if k == "Accept" {
```
rpatali: Also, the header map stores these headers using textproto.MIMEHeader(val), you don't want to assume casing of the "Accept" constant instead compare:
```
const _acceptHeader = "accept"
...
if k == textproto.MIMEHeader(_acceptHeader)
```

tejaswiagarwal: Great feedback, will get this in.

runtime/client_http_request.go
```diff
@@ -165,7 +172,12 @@ func (req *ClientHTTPRequest) WriteJSON(
 	}
 
 	if body != nil {
-		httpReq.Header.Set("Content-Type", "application/json")
+		httpReq.Header.Set(http.CanonicalHeaderKey(contentTypeHeader), contentTypeHeaderValue)
```
rpatali: you may be overwriting the ContentType if its already present. I think you want to check first and set it only if it isn't set already (sort of same logic as accept header below).

rpatali: I also just noticed `req.defaultHeaders` above.  Why aren't we using default headers to set these default values? If the caller set the header, the loop on line 170 will overwrite the default header which is what you want anyway.

tejaswiagarwal: That's modifying existing logic - outside the scope of this task.  

Also, Content-Type is to tell the downstream how to unmarshal the body - which in this case will always be application/json as long as there is a body.

tejaswiagarwal: The default headers have to be specified using the config files, which individual endpoints have to set. In this case, we want to apply this to all endpoints since this is a platform level logic (always be application/json since unmarshaler is always application/json)

rpatali: What if the incoming request had `vnd+application/json` or something like for content-type? does the downstream not care to know that?

The current implementation requires you to have config per client, but can very easily be extended to use a default global config for all clients as well as config for the said client so instead of one config, using two config properties, is that really that much of scope creep?

examples/example-gateway/config/test.yaml
```diff
@@ -67,6 +67,9 @@ tchannel.port: 0
 tchannel.processName: example-gateway
 tchannel.serviceName: example-gateway
 tchannel.clients.requestUUIDHeaderKey: x-request-uuid
+http.defaultHeaders:
+  Accept: application/json
+  Content-Type: application/json
```
argouber: content in file is alphabetical. can you move this with the other http params, so it is easy for someone to search/edit as needed

codegen/templates/http_client.tmpl
```diff
@@ -78,6 +78,9 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		timeoutVal,
 	)
 	defaultHeaders := make(map[string]string)
+	if deps.Default.Config.ContainsKey("http.defaultHeaders") {
```
rpatali: prefix key with `clients.`?

PR #668, opened 2019-11-07
Avoid nil map deref in workflow/endpoint templates

argouber:
The endpoint-template was iterating through the respHeaders without checking if it
was non-nil.  With this change, we actually eliminate the nil map creation and return
empty map of response headers.

It is better not to generate nils rather than have the caller test for nils

### DIFF
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -257,7 +257,7 @@ type {{$clientName}} struct {
 		{{end -}}
 
 		var success bool
-		var respHeaders map[string]string
+		respHeaders := map[string]string{}
 		var err error
 		if (c.circuitBreakerDisabled) {
 			success, respHeaders, err = c.client.Call(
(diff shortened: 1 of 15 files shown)


codegen/templates/tchannel_client.tmpl
```diff
@@ -257,7 +257,7 @@ type {{$clientName}} struct {
 		{{end -}}
 
 		var success bool
-		var respHeaders map[string]string
+		respHeaders := map[string]string{}
```
rpatali: `make(map[string]string)`
https://golang.org/doc/effective_go.html#allocation_make

runtime/tchannel_client.go
```diff
@@ -222,7 +222,8 @@ func (c *TChannelClient) call(
 	defer cancel()
 
 	err = c.ch.RunWithRetry(ctx, func(ctx netContext.Context, rs *tchannel.RequestState) (cerr error) {
-		call.resHeaders, call.success = nil, false
+		call.resHeaders = map[string]string{}
```
rpatali: same as above comment.

PR #673, opened 2019-11-14
Check before unmarshaling body

tejaswiagarwal:
We were not checking that the request body is empty before unmarshaling. Adding a check to prevent request failures.


rpatali:
I don't think this is the right place for the check. We should probably check in your endpoint logic before trying to unmarshal. Imagine a case where body is expected to have a required field but is not present. In such a case, an exception  is expected but will be bypassed.

PR #676, opened 2019-12-12
Serverless zanzibar

tejaswiagarwal:
Adds a new feature in Zanzibar to have a serverless mode. Endpoints can be created without a downstream, the workflow is short-circuited together to return a response to the caller. A new type convertor flow was added to convert Request to Response type, which is used by the serverless middleware.


codegen/gateway.go
```diff
@@ -373,6 +381,8 @@ func NewEndpointSpec(
 			)
 		}
 		workflowImportPath = iworkflowImportPath.(string)
+	} else if workflowType == "self-serving" {
```
argouber: just a refactoring suggestion -- once we are in this realm of more than 2 string name comparisons, it may make sense to have them as "const" value

codegen/method.go
```diff
@@ -1000,6 +1003,55 @@ func (ms *MethodSpec) setTypeConverters(
...
+) error {
+
+	dummyConverter := NewTypeConverter(h, nil)
+
+	respType := funcSpec.ResultSpec.ReturnType
+
+	dummyConverter.append(
+		"func convert",
+		PascalCase(ms.Name),
+		"DummyResponse(in ", ms.RequestType, ") ", ms.ResponseType, "{")
```
argouber: one space before the last {

codegen/method.go
```diff
@@ -1000,6 +1003,55 @@ func (ms *MethodSpec) setTypeConverters(
...
+		*compile.I16Spec,
+		*compile.I32Spec,
+		*compile.EnumSpec,
+		*compile.I64Spec,
+		*compile.DoubleSpec,
+		*compile.StringSpec:
+
+		dummyConverter.append("out", " := in\t\n")
+	default:
+		// default as struct
```
argouber: can it not be list/set/map?
Do you want a set of cases to either eliminate or covert those?

codegen/method.go
```diff
@@ -1000,6 +1003,55 @@ func (ms *MethodSpec) setTypeConverters(
...
+		*compile.BoolSpec,
+		*compile.I8Spec,
+		*compile.I16Spec,
+		*compile.I32Spec,
+		*compile.EnumSpec,
+		*compile.I64Spec,
+		*compile.DoubleSpec,
+		*compile.StringSpec:
+
+		dummyConverter.append("out", " := in\t\n")
```
argouber: In the conversion code I have seen in this and in type-converter, the pattern is usually to append a line without the terminating \n
and then finally do
strings.Join(lines, "\n") or equivalent

codegen/method.go
```diff
@@ -1000,6 +1003,55 @@ func (ms *MethodSpec) setTypeConverters(
...
+		respFields := respType.(*compile.StructSpec).Fields
+		dummyConverter.append("out", " := ", "&", ms.ShortResponseType, "{}\t\n")
+		err := dummyConverter.GenStructConverter(structType, respFields, reqrespTransforms)
+		if err != nil {
+			return err
+		}
+
+	}
+
+	dummyConverter.append("\nreturn out \t}")
```
argouber: swapped \n and \t ?

codegen/module_system.go
```diff
@@ -44,6 +44,7 @@ type EndpointMeta struct {
 	ClientMethodName       string
 	WorkflowPkg            string
 	ReqHeaders             map[string]*TypedHeader
+	DummyEndpoint          bool
```
argouber: general nit: IsDummyEndpoint -- if it is a bool.  Saying DummyEndpoint gives the impression that it is the endpoint info

codegen/service.go
```diff
@@ -221,9 +221,10 @@ func (ms *ModuleSpec) SetDownstream(
 		clientMethod = e.ClientMethod
 
 		// TODO: move generated middlewares out of zanzibar
-		headersPropagate = e.HeadersPropagate
-		reqTransforms    = e.ReqTransforms
-		respTransforms   = e.RespTransforms
+		headersPropagate  = e.HeadersPropagate
+		reqTransforms     = e.ReqTransforms
+		respTransforms    = e.RespTransforms
+		reqrespTransforms = e.DummyReqTransforms
```
argouber: rename the variable reqrespTransforms is confusing in the context of the previous two lines.

codegen/service.go
```diff
@@ -247,6 +248,17 @@ func (ms *ModuleSpec) SetDownstream(
 			"Service %q does not have method %q\n", serviceName, methodName,
 		)
 	}
+
+	if e.DummyEndpoint {
+		funcSpec := method.CompiledThriftSpec
+		err := method.setDummyTypeConverters(funcSpec, reqTransforms, headersPropagate, respTransforms, reqrespTransforms, h)
+		if err != nil {
+			fmt.Println("Coming to error")
```
argouber: you are going to clean this up. right?

runtime/server_http_request.go
```diff
@@ -130,6 +130,11 @@ func (req *ServerHTTPRequest) Context() context.Context {
 	return req.httpRequest.Context()
 }
 
+// StartTime returns the request's start time.
```
argouber: Where is this used?  I am not seeing it in the diff

codegen/method.go
tejaswiagarwal: Actually, if you look at the generated code, it is adding the space before the "{".

codegen/method.go
tejaswiagarwal: Actually, this is same as what we were doing above for setTypeConvertor method for the regular code-path (real workflow).

codegen/service.go
tejaswiagarwal: Yes

codegen/method.go
tejaswiagarwal: This is similar to the one above, using the same specs as conversion from client response to endpoint response.

runtime/server_http_request.go
tejaswiagarwal: This is from a previous diff, somehow got added here. I'll remove it.

codegen/method.go
tejaswiagarwal: List,sets and maps would default to structs and get converted using the type converter, i.e. it will be a deep copy. I will try and add endpoint test cases for these.

codegen/service.go
tejaswiagarwal: Updated it.

codegen/service.go
```diff
@@ -247,6 +248,16 @@ func (ms *ModuleSpec) SetDownstream(
...
 		)
 	}
+
+	if e.IsDummyEndpoint {
+		funcSpec := method.CompiledThriftSpec
+		err := method.setDummyTypeConverters(funcSpec, reqTransforms, headersPropagate, respTransforms, dummyReqTransforms, h)
+		if err != nil {
+			return err
+		}
+		return nil
```
argouber: Some message on the error? or just dispense with the error generation

codegen/method.go
```diff
@@ -1000,6 +1003,55 @@ func (ms *MethodSpec) setTypeConverters(
...
+		*compile.BoolSpec,
+		*compile.I8Spec,
+		*compile.I16Spec,
+		*compile.I32Spec,
+		*compile.EnumSpec,
+		*compile.I64Spec,
+		*compile.DoubleSpec,
+		*compile.StringSpec:
+
+		dummyConverter.append("out", " := in\t\n")
```
argouber: isn't this just "out := in\n"

codegen/module_system.go
```diff
@@ -1116,15 +1117,15 @@ func (g *EndpointGenerator) generateEndpointFile(
 		PackageName: instance.PackageInfo.GeneratedPackagePath + "/workflow",
 		AliasName:   "workflow",
 	})
-	if e.WorkflowImportPath != "" {
+	if e.WorkflowType == "custom" {
```
argouber: replace the "custom" in this file too :)

rpatali: +1

codegen/service.go
rpatali: having an error message will help with debugging a lot.

examples/example-gateway/endpoints/serverless/serverless-empty.yaml
```diff
@@ -0,0 +1,15 @@
+clientId:
```
abhishekparwal: add a tchannel endpointType example to see this would work.

abhishekparwal:
This is a clientless feature and not a serverless. server has a standard meaning and lets not change it. server is in this case zanzibar.

codegen/templates/serverless-workflow.tmpl
```diff
@@ -0,0 +1,154 @@
+{{/* template to render gateway workflow interface code */ -}}
```
abhishekparwal: lets call this workflow as clientless as discussed offline too. serverless is incorrect.

codegen/templates/endpoint.tmpl
```diff
@@ -10,13 +10,15 @@ package {{$instance.PackageInfo.PackageName}}
...
 {{- $serviceMethod := printf "%s%s" (title .Method.ThriftService) (title .Method.Name) }}
+{{- $dummyEndpoint := .IsDummyEndpoint}}
 {{- $handlerName := printf "%sHandler" $serviceMethod }}
 {{- $clientMethodName := title .ClientMethodName }}
 {{- $endpointId := .Spec.EndpointID }}
 {{- $handleId := .Spec.HandleID }}
 {{- $middlewares := .Spec.Middlewares }}
 {{- $workflowPkg := .WorkflowPkg }}
 {{- $workflowInterface := printf "%sWorkflow" $serviceMethod }}
+{{- $dummyWorkflowInterface := printf "%sDummyWorkflow" $serviceMethod }}
```
abhishekparwal: Not DummyWorkflow and same Workflow. However this would conflict and thus these needs a different package from server Workflow.  caller should know which impl to call based on the type.

codegen/templates/serverless-workflow.tmpl
```diff
@@ -0,0 +1,154 @@
...
+	{{end}}
+	{{- end}}
+
+	"go.uber.org/zap"
+	module "{{$instance.PackageInfo.ModulePackagePath}}"
+)
+
+{{with .Method -}}
+// {{$workflowInterface}} defines the interface for {{$serviceMethod}} workflow
+type {{$workflowInterface}} interface {
```
abhishekparwal: This interface should not be redeclared, it should be common b/w various workflows. program to interface and not to implementation.

codegen/templates/serverless-workflow.tmpl
tejaswiagarwal: Done

examples/example-gateway/endpoints/serverless/serverless-empty.yaml
tejaswiagarwal: This will be done in a follow up diff.

examples/example-gateway/endpoints/serverless/serverless-empty.yaml
```diff
@@ -0,0 +1,15 @@
...
+clientMethod:
+endpointId: serverless
```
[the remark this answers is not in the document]
tejaswiagarwal: Done

PR #691, opened 2020-03-16
Making :var pattern non-conflicting if added after original path

sandeepboys:


ChuntaoLu:
This PR changes the semantics of `:` to depend on the state of the trie, which will introduce ambiguity and create surprises.  

After following settings
```
/a -> foo
/: -> bar
```
What should I expect for `/a`? Since setting routes is an ongoing process, the author who sets `/:` may not know the existence of `/a` (and all that matches `/:`) and expect to cover all cases.


And what about 
```
/: -> bar
/a -> foo
```
Would ordering matter? (It shouldn't)

sandeepboys:
> This PR changes the semantics of `:` to depend on the state of the trie, which will introduce ambiguity and create surprises.
> 
> After following settings
> 
> ```
> /a -> foo
> /: -> bar
> ```
> 
> What should I expect for `/a`? Since setting routes is an ongoing process, the author who sets `/:` may not know the existence of `/a` (and all that matches `/:`) and expect to cover all cases.
> 
> And what about
> 
> ```
> /: -> bar
> /a -> foo
> ```
> 
> Would ordering matter? (It shouldn't)

Regarding your question about having both /a and /: in the routes, it is not an anti-pattern and here are some links that supports the case:
https://guides.rubyonrails.org/routing.html
https://www.playframework.com/documentation/2.8.x/JavaRouting

ChuntaoLu:
> > This PR changes the semantics of `:` to depend on the state of the trie, which will introduce ambiguity and create surprises.
> > After following settings
> > ```
> > /a -> foo
> > /: -> bar
> > ```
> > 
> > 
> > What should I expect for `/a`? Since setting routes is an ongoing process, the author who sets `/:` may not know the existence of `/a` (and all that matches `/:`) and expect to cover all cases.
> > And what about
> > ```
> > /: -> bar
> > /a -> foo
> > ```
> > 
> > 
> > Would ordering matter? (It shouldn't)
> 
> Regarding your question about having both /a and /: in the routes, it is not an anti-pattern and here are some links that supports the case:
> https://guides.rubyonrails.org/routing.html
> https://www.playframework.com/documentation/2.8.x/JavaRouting

Can you pin point the section that covers the topic?

sandeepboys:
> https://guides.rubyonrails.org/routing.html

https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions
If you look at section in above link, both "/photos/new" and "/photos/:id" paths are supported

rpatali:
Lu's concern on the ordering is important and that should be addressed. I would expect the ordering to not matter.

sandeepboys:
AFAIK mg-proxy takes ordering of paths into consideration while finding the route. Replicated the same behavior here. Correct me if I am wrong.

sandeepboys:
> Lu's concern on the ordering is important and that should be addressed. I would expect the ordering to not matter.

With new changes it is now not dependent on order

runtime/router/trie_test.go
```diff
@@ -187,19 +187,29 @@ func TestTriePathsWithPatten(t *testing.T) {
...
 
 	trie = NewTrie()
 	tests = []ts{
-		// test "/:" collides with "/b"
+		// test "/:" does not collides with "/b"
 		{op: set, path: "/b", value: "foo"},
-		{op: set, path: "/:a", errMsg: errExist.Error()},
+		{op: set, path: "/:a", value: "bar"},
 		{op: get, path: "/b/", expectedValue: "foo"},
+		{op: get, path: "/a/", expectedValue: "bar", expectedParams: []Param{{"a", "a"}}},
```
rpatali: why are params expected here?

rpatali:
LGTM but waiting for some more stakeholders to take a look at the diff.

runtime/router/trie_test.go
```diff
@@ -135,10 +135,10 @@ func TestTriePathsWithPatten(t *testing.T) {
 		{op: set, path: "/a/:b", value: "baz", errMsg: errExist.Error()},
 		// test "/*" collides with "/a"
 		{op: set, path: "/*", value: "baz", errMsg: errExist.Error()},
-		// test "/:" collides with "/a"
-		{op: set, path: "/:x", value: "baz", errMsg: errExist.Error()},
-		// test "/:/b" collides with "/a/*"
-		{op: set, path: "/:x/b", value: "baz", errMsg: errExist.Error()},
+		// test "/:" should not collide with "/a"
+		{op: set, path: "/:x", value: "bar"},
+		// test "/:/b" should not collide with "/a/*"
```
ChuntaoLu: This shouldn't be the case, `"/a/b"` is a match for both `"/:/b"` and `"/a/*"`.

runtime/router/trie_test.go
```diff
@@ -159,7 +159,7 @@ func TestTriePathsWithPatten(t *testing.T) {
 	trie = NewTrie()
 	tests = []ts{
 		{op: set, path: "/:a", value: "foo"},
-		{op: get, path: "/:a", expectedValue: "foo", expectedParams: []Param{{"a", ":a"}}},
+		{op: get, path: "/:a", expectedValue: "foo", expectedParams: []Param{{"a", "a"}}},
```
ChuntaoLu: The param is instead `:a` rather than `a` because the query path is `/:a` (note this is not a pattern, rather a literal path).

sandeepboys:
Created a new pull request for the agreed proposal: https://github.com/uber/zanzibar/pull/692

PR #692, opened 2020-03-20
Special router logic for whitelisted paths

sandeepboys:
With these changes for the configured whitelisted paths, the router logic will vary. Router treats "/a" and "/:b" as different paths is they are among the whitelisted paths.


runtime/router/router.go
```diff
@@ -64,6 +64,10 @@ type Router struct {
 	// unrecovered panics.
 	PanicHandler func(http.ResponseWriter, *http.Request, interface{})
 
+	// Used for special behavior using which different handlers can configured
+	// for paths such as /a and /:b in router.
+	WhitelistedPaths []string
```
abhishekparwal: update the comment at the top of router comment.

runtime/router/router.go
```diff
@@ -153,3 +158,14 @@ func (r *Router) allowed(path, reqMethod string) string {
 
 	return strings.Join(allow, ", ")
 }
+
+func (r *Router) isWhitelistedPath(path string) bool {
+	if len(r.WhitelistedPaths) > 0 {
+		for _, whitelistedPath := range r.WhitelistedPaths {
+			if strings.HasPrefix(path, whitelistedPath) {
+				return true
```
abhishekparwal: check for equality rather than prefix

runtime/router/router_test.go
```diff
@@ -47,19 +47,30 @@ func TestParamsFromContext(t *testing.T) {
...
+	})
+	err := r.Handle("GET", "/:var", handlerFunc)
 	assert.NoError(t, err, "unexpected error")
 
 	req, _ := http.NewRequest("GET", "/foo", nil)
 	r.ServeHTTP(nil, req)
 	assert.True(t, handled)
+
+	r = &Router{}
+	r.WhitelistedPaths = []string{"/bar"}
```
abhishekparwal: 1. a test case to check without whitelist it doesn't allow the path these paths i.e. `/a/b` and `/a/:var`
2. a test case to check with whitelist it allows above.

runtime/router/trie_with_speical_cases.go
```diff
@@ -0,0 +1,222 @@
...
+
+package router
+
+import "net/http"
+
+// set sets the handler for given path, creates new child node if necessary
+// lastKeyCharSlash tracks whether the previous key char is a '/', used to decide it is a pattern or not
+// when the current key char is ':'. lastPathCharSlash tracks whether the previous path char is a '/',
+// used to decide it is a pattern or not when the current path char is ':'.
+func (t *tnode) setForWhitelistedPath(path string, value http.Handler, lastKeyCharSlash, lastPathCharSlash bool) error {
```
abhishekparwal: rather than duplicating the entire code for whitelist path, can we just add it in trie.go. It will make it a single place and easier to maintain.

abhishekparwal:
> _No description provided._

Can we add a description of what we are trying to achieve for easier reference later.

runtime/router/router.go
```diff
@@ -153,3 +158,14 @@ func (r *Router) allowed(path, reqMethod string) string {
 
 	return strings.Join(allow, ", ")
 }
+
+func (r *Router) isWhitelistedPath(path string) bool {
+	if len(r.WhitelistedPaths) > 0 {
```
rpatali: *nit*: I don't think this `if` check is necessary, the `for` loop adequately handles this.

runtime/router/trie_with_speical_cases.go
sandeepboys: I have purposefully forked the logic for whitelisted paths because this will enable us to support adding more special cases for whitelisted paths in future. That way the original remains untoched.

runtime/router/router_test.go
sandeepboys: All those variety of test cases already exist in trie_test.go and trie_with_special_cases_test.go files

runtime/router/router.go
sandeepboys: We have to check prefix here because for _get_ method the /:var in the path can get changed but the prefix will remain same. To determine whether this is whitelisted path only prefix will tell that.

runtime/router/router.go
sandeepboys: Sure

runtime/router/router.go
sandeepboys: will update it.

runtime/router/router.go
rpatali: Wouldn't a whitelisted path `/a/b/:var` match with an existing path `/a/b/` because the prefixes are the same then?

runtime/router/trie_with_speical_cases.go
rpatali: I agree with Abhishek here, this is going to be a pain to keep the 2 versions in sync.  Imagine having to fix a bug in the get/set methods, it is possible to miss the fixes in one or the other place. Also, reviewing is a pain, I have to check which of the logic is new here.

sandeepboys: Right now itself the logic is hard to understand. The code will become more messy if we combine this logic. I will leave it here and let you guys take a call on this.

runtime/router/router.go
sandeepboys: In that case, we need do pattern matching for whitelisted paths. May be we need to create another trie for this?

runtime/router/router_test.go
sandeepboys: Done

runtime/router/router.go
```diff
@@ -30,9 +30,10 @@ import (
 // Router dispatches http requests to a registered http.Handler.
 // It implements a similar interface to the one in github.com/julienschmidt/httprouter,
 // the main differences are:
-// 1. this router does not treat "/a/:b" and "/a/b/c" as conflicts (https://github.com/julienschmidt/httprouter/issues/175)
+// 1. this router does not treat "/a/:b" and "/a/b/c" as conflicts (https://github.com/julienschmidt/httprouter/issues/175) except for whitelisted paths
```
rpatali: comment update makes it read as though if they are not whitelisted, they are conflicting which isn't true.

runtime/router/trie.go
```diff
@@ -95,7 +96,12 @@ func (t *Trie) Set(path string, value http.Handler) error {
 		return errors.New("path can not contain more than one *")
 	}
 
-	err := t.root.set(path, value, false, false)
+	colonAsPattern := true
+	if isWhitelisted {
+		colonAsPattern = false
+	}
```
rpatali: `colonAsPattern := !isWhitelisted` but you might as well just use:
`err := t.root.set(path, value, false, false, !isWhitelisted, isWhitelisted)`

runtime/router/trie.go
```diff
@@ -114,14 +121,18 @@ func (t *Trie) Get(path string) (http.Handler, []Param, error) {
 	}
 	// ignore trailing slash
 	path = strings.TrimSuffix(path, "/")
-	return t.root.get(path, false, false, false)
+	colonAsPattern := false
+	if isWhitelisted {
+		colonAsPattern = true
+	}
+	return t.root.get(path, false, false, colonAsPattern, isWhitelisted)
```
rpatali: same as above

runtime/router/trie.go
```diff
@@ -140,8 +151,7 @@ func (t *tnode) set(path string, value http.Handler, lastKeyCharSlash, lastPathC
 	// is immediately after slash, e.g. "/:foo", "/x/:y". "/a:b" is not a colon wildcard segment.
 	var keyMatchIdx, pathMatchIdx int
 	for keyMatchIdx < keyLength && pathMatchIdx < pathLength {
-		if (t.key[keyMatchIdx] == ':' && lastKeyCharSlash) ||
-			(path[pathMatchIdx] == ':' && lastPathCharSlash) {
+		if t.isWildCardPattern(path, keyMatchIdx, pathMatchIdx, lastKeyCharSlash, lastPathCharSlash, false, isWhitelisted) {
```
rpatali: is the colonAsPattern hardcoded to false intentionally?

rpatali:
LGTM except for the few nits.

runtime/router/router.go
```diff
@@ -153,3 +159,20 @@ func (r *Router) allowed(path, reqMethod string) string {
 
 	return strings.Join(allow, ", ")
 }
+
+func (r *Router) isWhitelistedPath(path string) bool {
```
abhishekparwal: lets use fixed path withtout path param in whitelist. santize those urls on init. 

This way it will be much simpler.

runtime/router/router.go
```diff
@@ -153,3 +159,20 @@ func (r *Router) allowed(path, reqMethod string) string {
...
+	for _, whitelistedPath := range r.WhitelistedPaths {
+		whitelistedPathTokens := strings.Split(whitelistedPath, "/")
+		pathTokens := strings.Split(path, "/")
+		if len(whitelistedPathTokens) != len(pathTokens) {
+			continue
+		}
+		for i, token := range whitelistedPathTokens {
+			if pathTokens[i] != token && token[0] != ':' {
```
abhishekparwal: no use of continue here.. irrespective you are continuing.

sandeepboys: what?

runtime/router/router.go
sandeepboys: It will not work for get because the variable path param gets changed for each request

runtime/router/trie.go
```diff
@@ -235,21 +245,21 @@ func (t *tnode) set(path string, value http.Handler, lastKeyCharSlash, lastPathC
...
-func (t *tnode) get(path string, lastKeyCharSlash, lastPathCharSlash, colonAsPattern bool) (http.Handler, []Param, error) {
+func (t *tnode) get(path string, lastKeyCharSlash, lastPathCharSlash, colonAsPattern, isWhitelistedPath bool) (http.Handler, []Param, error) {
 	keyLength, pathLength := len(t.key), len(path)
 	var params []Param
 
 	// find the longest matched prefix
 	var keyIdx, pathIdx int
 	for keyIdx < keyLength && pathIdx < pathLength {
-		if t.key[keyIdx] == ':' && lastKeyCharSlash {
+		if t.isWildCardPattern(path, keyIdx, pathIdx, lastKeyCharSlash, lastPathCharSlash, colonAsPattern, isWhitelistedPath) {
```
abhishekparwal: t.key[keyIdx] == ':' && lastKeyCharSlash && colonAsPattern is sufficient here

runtime/router/router.go
sandeepboys: Saving isWhitelisted at the leaf node does not work for the paths such as "/a/b/c" and "/a/:var/c". We discussed this in our zoom call and agreed to not store this flag in the trie. I do not see any issue with using this method as it looks cleaner and easier to understand IMHO.

runtime/router/trie.go
sandeepboys: It is not sufficient and it will not work for set (set invokes this get func) for the below sequence for non-whitelisted cases
/a/b
/a/:var
Here we expect the second path to fail but with your logic it will succeed.

runtime/router/trie.go
```diff
@@ -301,10 +306,33 @@ func (t *tnode) get(path string, lastKeyCharSlash, lastPathCharSlash, colonAsPat
...
+		// Prepending if child is not a pattern of :var
+		t.children = append([]*tnode{child}, t.children...)
+	} else {
+		// Appending if the child is of pattern :var
+		t.children = append(t.children, child)
+	}
+}
+
+func (t *tnode) isWildCardPattern(path string, keyIdx, pathIdx int, lastKeyCharSlash, lastPathCharSlash, colonAsPattern, isWhitelistedPath bool) bool {
+	if t.key[keyIdx] == ':' && lastKeyCharSlash && path[pathIdx] == ':' && lastPathCharSlash {
```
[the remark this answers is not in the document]
sandeepboys: ok

runtime/router/trie.go
```diff
@@ -301,10 +306,33 @@ func (t *tnode) get(path string, lastKeyCharSlash, lastPathCharSlash, colonAsPat
...
+		return true
+	}
+
```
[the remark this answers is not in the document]
sandeepboys: ok

runtime/router/trie.go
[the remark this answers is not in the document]
sandeepboys: Nope, we need to change the logic other wise it will not work for below sequence of paths for whitelisted paths
/a/b
/a/:var
Without the logic that I added the second path handler will replace the value of the /a/b node.

runtime/router/router.go
abhishekparwal: this is not addressed yet.

runtime/router/trie_with_speical_cases.go
abhishekparwal: resolving as duplication of the entire whitelist code is removed.

runtime/router/trie.go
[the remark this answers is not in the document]
abhishekparwal: sure, but can we make a specific function for getter and setter so we know which path is needed and which is redundant. makes it easier to maintain.

runtime/router/trie.go
abhishekparwal: sure, but can we make a specific function for getter and setter so we know which path is needed and which is redundant. makes it easier to maintain.

runtime/router/router.go
abhishekparwal: If we store at leaf node I am not convinced now why it would not work. 
1. Without whitelist, it would have conflicted.
2. As the concrete path is whitelisted we should allow its storing.
3. For any concrete path not whitelisted we should detect the conflict as in previous case.

abhishekparwal: Also, the option of storing and encoding all the metadata including whitelist info is generally a better option. Coupling it with getter to again consult the metatdata is not considered clean. However I will leave it you to decide what you think is better.

runtime/router/trie.go
```diff
@@ -114,14 +121,18 @@ func (t *Trie) Get(path string) (http.Handler, []Param, error) {
...
+	}
+	return t.root.get(path, false, false, colonAsPattern, isWhitelisted)
 }
 
 // set sets the handler for given path, creates new child node if necessary
 // lastKeyCharSlash tracks whether the previous key char is a '/', used to decide it is a pattern or not
 // when the current key char is ':'. lastPathCharSlash tracks whether the previous path char is a '/',
 // used to decide it is a pattern or not when the current path char is ':'.
-func (t *tnode) set(path string, value http.Handler, lastKeyCharSlash, lastPathCharSlash bool) error {
+func (t *tnode) set(path string, value http.Handler, lastKeyCharSlash, lastPathCharSlash, colonAsPattern, isWhitelisted bool) error {
```
abhishekparwal: Can you check if this handles URLs like /a/:b/c and /a/:x/d. For whitelist, this router should either give conflict as previous expectation OR if it allows, it should overwrite the other. Also, retrieval of them should be possible using concrete non whitelisted URL. Please add test case for this.

runtime/router/trie.go
sandeepboys: The logic is same for both and it is controlled by colonAsPattern flag. This flag is not something that I introduced.

runtime/router/trie.go
sandeepboys: There is already similar kind of test case exist. But I can add explicit one.

runtime/router/trie.go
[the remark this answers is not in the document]
sandeepboys: The logic is same for both and it is controlled by colonAsPattern flag. This flag is not something that I introduced.

runtime/router/router.go
sandeepboys: Saving it in whitelist does not work for all the use cases and it cause other issues. I will go with approach as this looks cleaner and easier to maintain IMO.

runtime/router/trie.go
sandeepboys: Having separate field for colonAsPattern would look it more easier to understand. I will go with your first approach.

runtime/router/trie.go
sandeepboys: Yes

sandeepboys: Fixed it

runtime/router/router.go
sandeepboys: Fixed

runtime/router/trie.go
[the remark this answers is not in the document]
sandeepboys: done

runtime/router/trie.go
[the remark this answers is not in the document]
sandeepboys: Created separate functions for getter and setter

runtime/router/trie.go
sandeepboys: Created separate functions for getter and setter

PR #695, opened 2020-03-27
Export metrics

tejaswiagarwal:
Exporting metrics to capture failure rates in HTTP clients. 

In the generated client code in EG - we want to add metrics to see failures. There is no way of doing that today, since we don't have the Metrics in scope there.

### DIFF
--- a/runtime/constants.go
+++ b/runtime/constants.go
@@ -46,6 +46,9 @@ const (
 	clientSystemErrors = "client.system-errors"
 	clientLatency      = "client.latency"
 	clientLatencyHist  = "client.latency-hist"
+
+	// clientHTTPUnmarshalError is the metric for tracking errors due to unmarshalling json responses
+	clientHTTPUnmarshalError = "client.http-unmarshal-error"
 )
 
 var knownMetrics = []string{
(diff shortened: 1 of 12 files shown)


rpatali:
Lets also update the HTTP and TChannel client in this diff that are going to use the exported metric property.

rpatali:
can you add a similar metric for TChannel? let's keep both in parity.

runtime/constants.go
```diff
@@ -46,6 +46,11 @@ const (
 	clientSystemErrors = "client.system-errors"
 	clientLatency      = "client.latency"
 	clientLatencyHist  = "client.latency-hist"
+
+	// clientHTTPUnmarshalError is the metric for tracking errors due to unmarshalling json responses
+	clientHTTPUnmarshalError = "client.http-unmarshal-error"
+	// clientTchannelReadError is the metric for tracking errors in reading tchannel response
+	clientTchannelReadError = "client.tchannel-read-error"
```
rpatali: technically `read` may be more appropriate but when comparing metrics, it helps if they are consistent. so I would call this `tchannel-unmarshal-error` along with the var name too.

rpatali:
LGTM though update the metric for tchannel.

tejaswiagarwal:
Adding metrics for failures of tchannel and HTTP response reads.

PR #698, opened 2020-04-08
Introduce JSONWrapper interface

styeung:
- this allows users to specify a json implementation to use in app.go
- the json implementation can be a different lib (like jsoniter) or even custom marshaling or unmarshaling behavior
- defaults to using encoding/json


rpatali:
> * this allows users to specify a json implementation to use in app.go
given this information, was [PR#697](https://github.com/uber/zanzibar/pull/697) even necessary?

tejaswiagarwal:
> > * this allows users to specify a json implementation to use in app.go
> >   given this information, was [PR#697](https://github.com/uber/zanzibar/pull/697) even necessary?

PR#697: Removes easyjson and moves to encoding/json - Cleans up generated easyjson code and has test changes for encoding/json.
 
PR#698: Provide a way for users to add more serialization/deserialization techniques apart from standard encoding/json.

rpatali:
@tejaswiagarwal I get that. What I'm trying to say is if this jsonwrapper allows for the user to use whatever library to encode, leaving the easyJSON implementation in here won't have hurt since the users of Zanzibar may choose to not depend on easyJSON at all in their implementations.

In short, internal projects could remove the dependency on easyJSON while Zanzibar could in theory support easyJSON behind the new interface.

runtime/server_http_response_test.go
```diff
@@ -558,3 +519,14 @@ func TestPendingResponseBody(t *testing.T) {
...
 	)
 }
+
+func createDefaultDependencies(bgateway *benchGateway.BenchGateway) *zanzibar.DefaultDependencies {
+	deps := &zanzibar.DefaultDependencies{
+		Scope:         bgateway.ActualGateway.RootScope,
+		Logger:        bgateway.ActualGateway.Logger,
+		ContextLogger: bgateway.ActualGateway.ContextLogger,
+		Tracer:        bgateway.ActualGateway.Tracer,
+		JSONWrapper:   jsonwrapper.NewDefaultJSONWrapper(),
```
rpatali, line 529: How would you initialize an easyJSON encoder? Can you add an example?

examples/example-gateway/app.go
```diff
@@ -34,6 +36,7 @@ import (
 var AppOptions = &zanzibar.Options{
 	GetContextScopeExtractors: getContextScopeTagExtractors,
 	GetContextFieldExtractors: getContextLogFieldExtractors,
+	JSONWrapper:               jsonwrapper.NewDefaultJSONWrapper(),
```
rpatali, line 39: `jsonwrapper.New()` reads better

runtime/client_http_request.go
```diff
@@ -79,6 +81,7 @@ func NewClientHTTPRequest(
 		ContextLogger:        NewContextLogger(client.loggers[clientMethod]),
 		defaultHeaders:       client.DefaultHeaders,
 		ctx:                  ctx,
+		jsonWrapper:          client.JSONWrapper,
```
rpatali, line 84: something like this reads better: `jsonMashaller: json.Marshaller` so you'll need to update the package name, field name and interface name

runtime/jsonwrapper/default_jsonwrapper.go
```diff
@@ -0,0 +1,44 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package jsonwrapper
```
rpatali, line 21: you probably want the package to be called just `json` and the interface `Marshaller` so the default struct would be `DefaultJSONMarshaller` and the new method as just `New()`. Overall it would read much better and make a more intuitive sense of what this interface does.

runtime/jsonwrapper/jsonwrapper.go
```diff
@@ -0,0 +1,27 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package jsonwrapper
+
+// JSONWrapper provides an interface for json marshaling and unmarshaling operations
+type JSONWrapper interface {
```
rpatali, line 24: this doesn't have to be a separate file, you can just put it together with the default implementation. It would make more sense in case this was in a different package than the default impl.

PR #699, opened 2020-04-08
Revert "Remove easyjson  (#697)"

tejaswiagarwal:
This reverts commit f4a7b3344f1660bcafc25bab74ac0810ab50342c.


rpatali:
Reverting since PR#698 supports providing a custom implementation for (de)serializing and a user may/may not use  EasyJSON without having to delete this. Thought, as a platform, if we have moved completely away from EasyJSON, we can revisit this.

PR #702, opened 2020-04-15
support not follow redirect for HTTP client

hothero:
Enable HTTP clients to have a config to choose whether it wants to follow HTTP redirect or not. In other words, if the HTTP client does not follow the redirect, the caller will receive 30* Status with Location header


rpatali:
Please add a test case with redirect set to false for a given client using example/example-gateway/config/*.yaml

runtime/client_http_response_test.go
```diff
@@ -230,6 +242,79 @@ func TestUnknownStatusCode(t *testing.T) {
...
+		bgateway.ActualGateway.ContextMetrics,
+		jsonwrapper.NewDefaultJSONWrapper(),
+		"bar",
+		map[string]string{
+			"echo": "bar::echo",
+		},
+		baseURL,
+		map[string]string{},
+		time.Second,
+		false,
```
hothero, line 295: @rpatali does this test case address your ask? in this test case, it creates the gateway via exampleGateway and use example gateway's config files. And this test case is to verify the client call doesn't follow the redirect, so the caller will receive 303 status code.

rpatali: this is fine but additional also set the config value for `clients.*.followRedirects = false` for one of the clients in the <env>.json to test the behavior since that is what you are going to need.

runtime/client_http_response_test.go
```diff
@@ -230,6 +242,79 @@ func TestUnknownStatusCode(t *testing.T) {
...
+
+	redirectURI := baseURL + "/bar-path"
+	fakeEcho := func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Location", redirectURI)
+		w.WriteHeader(303)
+		_, err := w.Write([]byte(`false`))
+		assert.NoError(t, err)
+	}
+
+	fakeNormal := func(w http.ResponseWriter, r *http.Request) {
```
rpatali, line 275: this doesn't seem to be used.

runtime/client_http_response_test.go
```diff
@@ -230,6 +242,79 @@ func TestUnknownStatusCode(t *testing.T) {
...
+		w.WriteHeader(999)
+		_, err := w.Write([]byte(`false`))
+		assert.NoError(t, err)
+	}
+
+	bgateway.HTTPBackends()["bar"].HandleFunc("POST", "/bar/echo", fakeEcho)
+	bgateway.HTTPBackends()["bar"].HandleFunc("GET", "/bar-path", fakeNormal)
```
rpatali, line 282: this doesn't seem to be used, remove it.

hothero: make sense, since we don't follow the redirect, we don't need it anymore.

PR #703, opened 2020-04-18
parallelize endpoint gen, post gen hook, optimize loop, repeated reading, scanning - part2

abhishekparwal:
parallelize endpoint gen, post gen hook, optimize loops. removing unnecessary reading of files repeatedly, optimize scanning by using memory over cpu.

### DIFF
--- a/codegen/module_system.go
+++ b/codegen/module_system.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 
 	validator2 "gopkg.in/validator.v2"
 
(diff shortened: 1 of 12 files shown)


codegen/module_system.go
```diff
@@ -1355,84 +1415,126 @@ func (generator *GatewayServiceGenerator) ComputeSpec(
...
-		generator.packageHelper,
-	)
-	if err != nil {
-		return nil, errors.Wrapf(
-			err,
-			"Error generating service service.go for %s",
-			instance.InstanceName,
+	var fileMap sync.Map
+	var wg sync.WaitGroup
+	wgCount := 5
```
smb158, line 1420: Why is this hard coded to 5?

abhishekparwal: added a comment to specify its a goroutine count

PR #704, opened 2020-04-20
run check generate on master

rpatali:
New changesets are failing because somewhere before this changeset didn't perform check-generate.


rpatali:
I don't think check-generate on master fixed anything. I think the issue is somewhere else, going to abandon this PR.

PR #705, opened 2020-04-23
Deletes can have a body in rare circumstances, instead we will requir…

smb158:
…e uses to annotate to assign fields as from query param

### DIFF
--- a/codegen/method.go
+++ b/codegen/method.go
@@ -254,7 +254,7 @@ func NewMethod(
 	method.setValidStatusCodes()
 
 	if method.RequestType != "" {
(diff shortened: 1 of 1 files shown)


codegen/method.go
```diff
@@ -254,7 +254,7 @@ func NewMethod(
 	method.setValidStatusCodes()
 
 	if method.RequestType != "" {
-		hasNoBody := method.HTTPMethod == "GET" || method.HTTPMethod == "DELETE"
+		hasNoBody := method.HTTPMethod == "GET"
```
rpatali, line 257: add a test by creating an endpoint in example-gateway that is a DELETE and has a request body.

smb158: Good point, I added a test endpoint to exercise this path

PR #707, opened 2020-04-26
Selective module generation

abhishekparwal:
This feature allows you to select module to do generation for. So, if a certain module has changed, then it only generates code for it and anyone who is dependent on that module all the way up the dependency tree. This feature will help in reducing binary size

testing plan: code-gen for a new selective-gateway module is a verification test in addition to unit test.


codegen/runner/runner.go
```diff
@@ -154,9 +157,10 @@ func main() {
 		if config.ContainsKey("parallelizeFactor") {
 			parallelizeFactor = int(config.MustGetInt("parallelizeFactor"))
 		}
-		moduleSystem, err = codegen.NewDefaultModuleSystemWithMockHook(packageHelper, true, true, true, "test.yaml", parallelizeFactor)
+		moduleSystem, err = codegen.NewDefaultModuleSystemWithMockHook(packageHelper, true, true, true, "test.yaml",
+			parallelizeFactor, true)
```
smb158, line 161: Theres some weird spacing here.

codegen/module.go
```diff
@@ -1395,6 +1275,80 @@ func (system *ModuleSystem) IncrementalBuild(
...
+		for _, instance := range instances {
+			if instance.SelectiveBuilding {
+				selectiveModuleInstances[instanceFQN(instance)] = instance
+			}
+		}
+	}
+	return selectiveModuleInstances
+}
+
+// recursiveSelectiveDependencies gets  a recursive dependencies of direct dependencies including itself
```
smb158, line 1301: Extra space between gets and a. Also this comment is pretty unclear.

PR #708, opened 2020-04-28
parallelization module

abhishekparwal:
Parallelize functions in golang

This library works in few modes: a) for doing short io bound work default, it spawns 2x goroutines of CPU. b) If its unsuitable parallel count can be specified. c) for running long-running io bound work provides a function to spawns unbounded goroutine.

It returns early with the first error it sees. note, first error it sees from parallel goroutine and not first error as in parallel many goroutines may simultaneously return an error.

### DIFF
--- a/parallelize/parallelize.go
+++ b/parallelize/parallelize.go
@@ -1,0 +1,208 @@
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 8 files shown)


parallelize/parallelize.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+	}()
+}
+
+// NewFixedBoundedRunner creates fixed bounded goroutines by factor of available CPU based on default setting.
+// Default setting is to have parallelization as factor of 4x if iobound work which is supposed to be short.
+// If its a long work use NewBoundedRunner instead with optimal config.
+// If cpu bound work parallel go routine will be bound by cpu.
+func NewFixedBoundedRunner(workSize int, ioBound bool) *Runner {
+	parallelCount := runtime.NumCPU()
```
jcorbin, line 70: NOTE: this should be using `runtime.GOMAXPROCS(-1)` so that it's aware of container CPU quota as set by our https://github.com/uber-go/automaxprocs/ library

abhishekparwal: so you are saying we should do at top of the parallelize.go underscore imports (_ "go.uber.org/automaxprocs"), and then fetch Num of CPU using runtime.GOMAXPROCS(-1) ?

abhishekparwal: makes sense, to not impose the automaxprocs on this lib users.

PR #709, opened 2020-04-29
support for system wide whitelisting of default headers

rpatali:
we have a concept of default middleware, this is a same concept where we
don't need a whitelist of headers per endpoint but rather a system-wide
whitelist that will be always forwarded when available. These should be
used for more of control plane kind of headers rather than application
logic.

### DIFF
--- a/codegen/runner/runner.go
+++ b/codegen/runner/runner.go
@@ -114,6 +114,9 @@ func main() {
 	defaultDependencies := make(map[string][]string, 0)
 	config.MustGetStruct("defaultDependencies", &defaultDependencies)
 
+	defaultHeaders := make([]string, 0)
+	config.MustGetStruct("defaultHeaders", &defaultHeaders)
+
 	options := &codegen.PackageHelperOptions{
 		RelThriftRootDir:              config.MustGetString("thriftRootDir"),
 		RelTargetGenDir:               config.MustGetString("targetGenDir"),
@@ -127,6 +130,7 @@ func main() {
 		TraceKey:                      config.MustGetString("traceKey"),
 		ModuleSearchPaths:             searchPaths,
 		DefaultDependencies:           defaultDependencies,
+		DefaultHeaders:                defaultHeaders,
 	}
 
 	packageHelper, err := codegen.NewPackageHelper(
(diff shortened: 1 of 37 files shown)


codegen/runner/runner.go
```diff
@@ -114,6 +114,9 @@ func main() {
 	defaultDependencies := make(map[string][]string, 0)
 	config.MustGetStruct("defaultDependencies", &defaultDependencies)
 
+	defaultHeaders := make([]string, 0)
+	config.MustGetStruct("defaultHeaders", &defaultHeaders)
```
argouber, line 118: Meaning this is mandatory?  Unspecified = empty semantics hold?
Also, just asking -- since this is the place - any validation needed? i.e. disallowed values, values with spaces etc... just like any header ...

codegen/templates/workflow.tmpl
```diff
@@ -133,10 +134,24 @@ func (w {{$workflowStruct}}) Handle(
 	{{end}}
 
 	clientHeaders := map[string]string{}
+	{{if (ne (len $defaultHeaders) 0) }}
+	var ok bool
+	var h string
+	{{range $i, $k := $defaultHeaders}}
+	k := textproto.CanonicalMIMEHeaderKey("{{$k}}")
```
argouber, line 141: You may not need to do this because `Get()` should handle it internally

codegen/templates/workflow.tmpl
```diff
@@ -133,10 +134,24 @@ func (w {{$workflowStruct}}) Handle(
...
+	h, ok = reqHeaders.Get(k)
+	if ok {
+		clientHeaders[k] = h
+	}
```
argouber, line 145: nit:  a more go-idiomatic way would have been to do something like
```
if h, ok := reqHeaders.Get(k); ok {
  ...
}
```
That would limit the variable scope

rpatali: either way, it's not a big deal but what we have prevents creating several new variables and is intentional.

codegen/templates/workflow.tmpl
rpatali: yeah, we need it either way, unfortunately.

codegen/runner/runner.go
rpatali: good point, it shouldn't be must get, we should first check if the key exists and if it does, only then read it since this is an optional config.

rpatali:
> Rather than code-gen, can we not do this in the runtime code -- to look at and add this.

we currently only look at <env>.json which are for configurations that are more dynamic in nature and usually change based on the env. For more static kind of config (for a given gateway), its canonical to be configured via build.yaml

PR #713, opened 2020-05-20
glide-up and travis-tools to go1.14

argouber:
- Updated dependencies in glide
   - needed thriftrw-go and tchannel to be the latest
  
- Moved to go v1.14 for CI
  - And some extra dependencies needed to be added
  - codegen needed to be rerun and fixes applied.


.travis.yml
```diff
@@ -1,12 +1,10 @@
 language: go
 sudo: true
 go:
-  - 1.11.x
-  - 1.12.x
```
smb158, line 5: I think this is fine, but we should notify the presentation folks that we made this change

glide.yaml
```diff
@@ -20,7 +20,7 @@ import:
 - package: github.com/kardianos/osext
   version: master
 - package: go.uber.org/thriftrw
-  version: ^v1.16.0
+  version: ^v1.23.0
```
smb158, line 23: Lets make sure we bump thriftrw in EG and EGO as well. Want to keep them all in sync.

argouber: ack

.travis.yml
argouber: they're on 1.13 and will go to 1.14 soon (there is some version lag for their repo). So, this should be okay.  If needed, and they don't move fast enough, we can add 1.13.

PR #718, opened 2020-06-03
logging for conflicting url paths

abhishekparwal:


runtime/router/router.go
```diff
@@ -95,7 +96,22 @@ func (r *Router) Handle(method, path string, handler http.Handler) error {
...
+
+// urlFailure captures errors for conflicting paths
+type urlFailure struct {
+	method string
+	url    string
+}
+
+// Error returns the error string
+func (e *urlFailure) Error() string {
+	return fmt.Sprintf("panic: path: %q method: %q conflicts with an existing path", e.url, e.method)
```
rpatali, line 114: don't think `panic` should be part of the message itself.

abhishekparwal: acked. we can improve message in future diff.

scripts/cover.sh
```diff
@@ -128,7 +128,7 @@ cat ./coverage/istanbul.json | jq '[
 ] | from_entries' > ./coverage/istanbul-runtime.json
 
 echo "Checking code coverage for runtime folder"
-./node_modules/.bin/istanbul check-coverage --statements 99.92 \
+./node_modules/.bin/istanbul check-coverage --statements 99.85 \
```
rpatali, line 131: I'm sure there are tests for duplicate URL case, I would have not expected the coverage to drop. If this is not the case, we should add coverage instead of changing this.

abhishekparwal: it is there.. https://github.com/uber/zanzibar/blob/master/runtime/router_test.go#L120-L129

abhishekparwal: and see the previous run of this, where test fails and you can see it didn't decrease for this package.

PR #720, opened 2020-06-17
skipping module only when thrift is not present

abhishekparwal:
thrift being present or failing if not present is to be determined by previous steps than incremental. In one of the implementations utilizing Zanzibar code gen, it is doing while generating minified thrift specs and if not valid or not present for a module, fails at that layer. thus, at this layer,  if a spec is not present, it is skippable as determined by the previous layer.

### DIFF
--- a/codegen/module.go
+++ b/codegen/module.go
@@ -1085,7 +1085,9 @@ func (system *ModuleSystem) populateSpec(instance *ModuleInstance) error {
 
 	spec, err := specProvider.ComputeSpec(instance)
 	if err != nil {
-		return fmt.Errorf("error when running computespec: %s", err.Error())
+		// Don't create new errors, rather wrap the error as we need the original error at caller
+		fmt.Println("error when running computespec", err.Error())
+		return err
 	}
 	instance.genSpec = spec
 	// HACK: to get get of bad modules, which should not be there at first place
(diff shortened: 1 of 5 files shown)


codegen/module.go
```diff
@@ -1190,13 +1190,16 @@ func (system *ModuleSystem) IncrementalBuild(
...
-		ch := make(chan *ModuleInstance, len(resolvedModules[className]))
+		ch := make(chan *populateSpecRes, len(resolvedModules[className]))
 		for _, instance := range resolvedModules[className] {
 			go func(instance *ModuleInstance) {
 				defer wg.Done()
 				if err := system.populateSpec(instance); err != nil {
-					// HACK: to get get of bad modules, which should not be even be loaded in dag at first place
-					ch <- instance
+					if _, ok := err.(*ErrorSkipCodeGen); ok {
+						// HACK: to get get of bad modules, which should not be even be loaded in dag at first place
```
rpatali, line 1199: probably should update this comment to explain clearly as what is considered a bad module.

abhishekparwal: error type name which is explanatory with a godoc comment on that, now I feel is sufficient. avoiding comment explanation here again to abstract those reasons to modules that may do it in different conditions and thus suited there.

PR #723, opened 2020-07-02
Allow histogram buckets to be customized by gateways

argouber:
Currently the histogram duration buckets for tally stats is hard-coded in the common code.
We have gateways that need customized set of buckets as per their use-case.
Therefore, it is better to have the option available in the gateway config.
This change enables that.

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -73,6 +73,7 @@ const (
 // Options configures the gateway
 type Options struct {
 	MetricsBackend            tally.CachedStatsReporter
+	MetricsDefaultBuckets     tally.Buckets // To support non-default histogram metric bucketing specific to a gateway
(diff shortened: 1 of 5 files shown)


runtime/gateway.go
```diff
@@ -163,6 +166,11 @@ func CreateGateway(
 	if opts.MetricsBackend != nil {
 		metricsBackend = opts.MetricsBackend
 	}
+	if opts.MetricsDefaultBuckets != nil {
+		metricsDefaultBuckets = opts.MetricsDefaultBuckets
+	} else {
+		metricsDefaultBuckets = tally.DefaultBuckets
```
argouber, line 172: Here we are letting the default behavior be rather than the exponential 11 bucket scheme we had set things to previously.  That scheme should be in the specific gateway configuration.

Issue #725, opened 2020-07-07
nil pointer dereference


Xiot:
I'm trying to generate my clients (grpc) however I keep getting a `nil pointer dereference`

```
Need to generate "apps/rider/helix/rider-presentation" "service" "gateway" because it transitively depends on "apps/rider/helix/X" "endpoint" "tchannel"
Generating         grpc       client fulfillment-gateway-triggers   in build/clients/fulfillment-gateway-triggers         1/3
Generating     tchannel     endpoint apps/rider/helix/X       in build/apps/rider/helix/endpoints/X           2/3
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xc0 pc=0x12f093d]

goroutine 1 [running]:
github.com/uber/zanzibar/codegen.sortByClientName.Less(...)
        external/com_github_uber_zanzibar/codegen/module_system.go:1566
sort.insertionSort(0x1af8560, 0xc0073b6aa0, 0x0, 0x9)
        GOROOT/src/sort/sort.go:27 +0xc4
sort.quickSort(0x1af8560, 0xc0073b6aa0, 0x0, 0x9, 0x8)
        GOROOT/src/sort/sort.go:209 +0x201
sort.Sort(0x1af8560, 0xc0073b6aa0)
        GOROOT/src/sort/sort.go:218 +0x79
github.com/uber/zanzibar/codegen.readClientDependencySpecs(0xc0006f0800, 0x0, 0xc003e1d1b8, 0x10d3851)
        external/com_github_uber_zanzibar/codegen/module_system.go:1483 +0x173
github.com/uber/zanzibar/codegen.(*EndpointGenerator).ComputeSpec(0xc000a03a90, 0xc0006f0800, 0x0, 0x210fb58, 0x1040200, 0xc000000180)
        external/com_github_uber_zanzibar/codegen/module_system.go:949 +0x43
```

PR #726, opened 2020-07-08
Fix grpcClientGenerator

ChuntaoLu:
Currently when an endpoint depends on multiple grpc clients, endpoint code generation fails due to missing grpc client specs. This PR fixes the problem by returning the client spec at the end of grpc client generation.

Fixes https://github.com/uber/zanzibar/issues/725.


ChuntaoLu:
> LGTM (sans the build failures which I'm sure you'll fix, appears to be caused by formatting changes).

I am not sure what is causing the format problems, we vendor `goimports` and I believe `gofmt` is bundled with go so there shouldn't be a versioning problem. And I can't reproduce the problem locally, also the travis job git diff output shows the formatting on that particular job is actually wrong (in reverse alphabetical order). 

I am going to merge the pr for now, and investigate further if the problem persists.

PR #728, opened 2020-07-11
Support different gen code packages for different idl types

ChuntaoLu:
Currently generated code for different idl types (thrift and proto) must reside in the same generated code package under `build` dir, as indicated by `genCodePackage` filed value in application config file (build.yaml). This PR adds support so that different idl types can generate code to different gen code packages. 

The rationale is to add flexibilty so that zanzibar takes less assumption about pre-processing (i.e., generate code from idl files).


codegen/package.go
```diff
@@ -94,8 +95,9 @@ type PackageHelperOptions struct {
 	// relative path to the default middleware config dir, defaults to ""
 	RelDefaultMiddlewareConfigDir string
 
-	// package path to the generated code, defaults to PackageRoot + "/" + RelTargetGenDir + "/gen-code"
-	GenCodePackage string
+	// map of idl type to package path of the generated code
+	// map should be keyed with idl file extension, e.g., .thrift, .proto
+	GenCodePackage map[string]string
```
rpatali, line 100: This is fine but probably we when reading from config, we lowercase the ext and store in the map

codegen/package.go
```diff
@@ -256,36 +263,42 @@ func (p PackageHelper) DefaultMiddlewareSpecs() map[string]*MiddlewareSpec {
 	return p.defaultMiddlewareSpecs
 }
 
-// GenCodePackage returns the file path to the idl generated code folder
-func (p PackageHelper) GenCodePackage() string {
+// GenCodePackage returns map of idl type to the file path of the idl generated code folder
+func (p PackageHelper) GenCodePackage() map[string]string {
```
rpatali, line 267: probably also provide a lookup method that takes the string input, does lower case it and then look up the map. Can't safely assume the ext is always .proto and not .Proto or .PROTO

codegen/package.go
ChuntaoLu: They have to be lowercase when provided by user, otherwise won't be valid file extension. The lowercase is also mandated by the the schema defined in this PR https://github.com/uber/zanzibar/pull/728/files/fa0b790cda65e487f2d88a6c1d34e64ca91b56a8#diff-d750b667b8c67069eee0381c084dc1caR23

codegen/package.go
ChuntaoLu: I don't think `.Proto` and `.PROTO` are valid protobuf file extension. And as mentioned above the key is prescribed in the schema.

PR #733, opened 2020-07-28
Add support for fetching multiple header values

vermagav:


runtime/server_header.go
```diff
@@ -33,7 +33,14 @@ import (
...
 // Header defines methods on ServerHeaders
 type Header interface {
+	// Get mirrors the implementation of http.header and returns a single header value.
+	// When a key contains multiple values, -only- the first one is returned.
+	// Ref: https://golang.org/pkg/net/http/#Header.Get
 	Get(key string) (string, bool)
+	// Values mirrors the implementation of http.header and returns a slice of header values.
+	// When a key contains multiple values, the entire collection is returned.
+	// Ref: https://golang.org/pkg/net/http/#Header.Values
+	//Values(key string) ([]string, bool)
```
rpatali, line 43: Forgot to uncomment?

vermagav: Yep, that was from a stale commit. Updates pending.

runtime/server_header.go
```diff
@@ -151,6 +169,19 @@ func (th ServerTChannelHeader) Get(key string) (string, bool) {
...
 
+// Values retrieves the entire collection of values stored on a given header.
+// Bool return value is used to distinguish between the presence of a
+// header with golang's zerovalue slice and the absence of the header.
+func (th ServerTChannelHeader) Values(key string) ([]string, bool) {
+	if value, ok := th.Get(key); ok {
+		// In the case of TChannel, ServerTChannelHeader does not support
+		//  multiple, disparate values so we defer to Get and package it
+		// in a slice to meet the interface's requirement.
+		return []string{value}, ok
```
tejaswiagarwal, line 180: Can we do this instead? 

`return th.Get(key)`

ChuntaoLu:
The ci failure is due to format inconsistency on travis ci, safe to ignore.

PR #734, opened 2020-07-29
Set explicit import alias for zanzibar runtime package in templates

ChuntaoLu:
When generated code is formatted by `goimports`, import `github.com/uber/zanzibar/runtime` is removed if the library is not in `GOPATH`, because `goimports` can't resolve the alias `zanzibar` to `github/uber/zanzibar/runtime`. This PR fixes such problem by setting explicit import alias for `github/uber/zanzibar/runtime` in templates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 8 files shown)


rpatali:
files inside build dir didn't change?

ChuntaoLu:
> files inside build dir didn't change?

Yes, no changes are expected. When zanzibar is in `GOPATH`, everything is fine. Only when zanzibar is absent in `GOPATH`, `goimports` will incorrectly remove `github.com/uber/zanzibar/runtime` if `zanzibar` alias is not explicitly set, with wrong assumption that `github.com/uber/zanzibar/runtime` is not used.

PR #735, opened 2020-07-29
Add alternative routing support for HTTP clients

smb158:


codegen/templates/http_client.tmpl
```diff
@@ -206,7 +229,24 @@ func (c *{{$clientName}}) {{$methodName}}(
 
 	{{if $sidecarRouter -}}
 	headers[c.callerHeader] = c.callerName
-	headers[c.calleeHeader] = c.calleeName
+
+	// Set the service name if dynamic routing header is present
```
smb158, line 233: I realized I should invert this... iterate through alternative routes and see if header is present in header map


There will be less routes than headers generally (most often 0!)

codegen/templates/http_client.tmpl
```diff
@@ -124,6 +133,20 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	}
 }
 
+{{if $sidecarRouter -}}
+func initializeAlternativeRoutingMap(altServiceDetail config.AlternateServiceDetail) map[string]map[string]string {
```
argouber, line 137: :bikeshed: use "Alternate" in keeping with the other naming

codegen/templates/http_client.tmpl
argouber: yes. good idea. 👍

codegen/templates/http_client.tmpl
```diff
@@ -206,7 +229,24 @@ func (c *{{$clientName}}) {{$methodName}}(
...
-	headers[c.calleeHeader] = c.calleeName
+
+	// Set the service name if dynamic routing header is present
+	for headerKey, headerVal := range headers {
+		if routeMap, ok := c.alternateRoutingMap[headerKey]; ok {
+			for routeRegex, altServiceName := range routeMap {
+				//if headerVal matches routeRegex regex, set the alternative service name
+				matchFound, _ := regexp.MatchString(routeRegex, headerVal)
+				if matchFound {
+					headers[c.callerHeader] = altServiceName
```
argouber, line 240: should be callee

codegen/templates/http_client.tmpl
```diff
@@ -206,7 +229,24 @@ func (c *{{$clientName}}) {{$methodName}}(
...
+				}
+			}
+		}
+	}
+
+	// If serviceName was not set in the dynamic routing section above, set as the default
+	if _, ok := headers[c.callerHeader]; !ok {
+		headers[c.callerHeader] = c.callerName
```
argouber, line 248: callee

codegen/templates/http_client.tmpl
```diff
@@ -206,7 +229,24 @@ func (c *{{$clientName}}) {{$methodName}}(
...
 	{{if $sidecarRouter -}}
 	headers[c.callerHeader] = c.callerName
-	headers[c.calleeHeader] = c.calleeName
+
+	// Set the service name if dynamic routing header is present
+	for headerKey, headerVal := range headers {
+		if routeMap, ok := c.alternateRoutingMap[headerKey]; ok {
+			for routeRegex, altServiceName := range routeMap {
+				//if headerVal matches routeRegex regex, set the alternative service name
+				matchFound, _ := regexp.MatchString(routeRegex, headerVal)
```
argouber, line 238: nit **only**:  
if matchFound, _ .....; matchFound {

codegen/templates/http_client.tmpl
```diff
@@ -105,6 +113,7 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		calleeHeader: calleeHeader,
 		callerName: callerName,
 		calleeName: calleeName,
+		alternateRoutingMap: initializeAlternativeRoutingMap(alternateServiceDetail),
```
argouber, line 116: Keep the naming same -- Alternate vs Alternative

codegen/templates/http_client.tmpl
```diff
@@ -206,7 +229,24 @@ func (c *{{$clientName}}) {{$methodName}}(
...
+
+	// Set the service name if dynamic routing header is present
+	for headerKey, headerVal := range headers {
+		if routeMap, ok := c.alternateRoutingMap[headerKey]; ok {
+			for routeRegex, altServiceName := range routeMap {
+				//if headerVal matches routeRegex regex, set the alternative service name
+				matchFound, _ := regexp.MatchString(routeRegex, headerVal)
+				if matchFound {
+					headers[c.calleeHeader] = altServiceName
+				}
```
argouber, line 241: you want to break at first match.

codegen/templates/http_client.tmpl
```diff
@@ -124,6 +133,20 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	}
 }
 
+{{if $sidecarRouter -}}
+func initializeAltRoutingMap(altServiceDetail config.AlternateServiceDetail) map[string]map[string]string {
+	routingMap := make(map[string]map[string]string)
+	for _, alt := range altServiceDetail.RoutingConfigs {
+		if headerValueToServiceMap, ok := routingMap[alt.HeaderName]; ok {
+			headerValueToServiceMap[alt.HeaderValue] = alt.ServiceName
```
argouber, line 141: we're trying to handle duplicates here with a merge?

examples/example-gateway/config/production.json
```diff
@@ -27,6 +27,21 @@
 	"clients.corge-http.serviceName": "CorgeHttp",
 	"clients.corge-http.maxConcurrentRequests": 1000,
 	"clients.corge-http.errorPercentThreshold": 20,
+	"clients.corge-http.alternates": {
+	  "routingConfigs": [
+		{
```
argouber, line 32: should we also test with multiple, given that code is trying to do multiple....

codegen/templates/http_client.tmpl
```diff
@@ -206,7 +229,24 @@ func (c *{{$clientName}}) {{$methodName}}(
 
 	{{if $sidecarRouter -}}
 	headers[c.callerHeader] = c.callerName
-	headers[c.calleeHeader] = c.calleeName
+
+	// Set the service name if dynamic routing header is present
+	for headerKey, headerVal := range headers {
+		if routeMap, ok := c.alternateRoutingMap[headerKey]; ok {
```
rpatali: it might be better if you iterate over alternateRoutingMap and see if any of those keys are present in the incoming headers. By doing it this way, even when no alternative routes are configured, you would have iterated over all the headers and did a lookup in the map (which is the common case) and the alternative route headers are quite small in number (usually 1) whereas there are several req headers (10+)

examples/example-gateway/config/production.json
smb158: yeah ill add this

codegen/templates/http_client.tmpl
smb158: I was trying to handle one header, can have different end routing destinations depending on the value

For example 

		"routingConfigs": [
			{
				"headerName": "Rpc-Header-Dest",
				"headerValue": "some-value1",
				"serviceName": "some-service"
			},
			{
				"headerName": "Rpc-Header-Dest",
				"headerValue": "another-value2",
				"serviceName": "another-service"
			}

You can argue that we don't need to support this, but its what I was thinking.

examples/example-gateway/build/clients/corge-http/corge-http.go
```diff
@@ -85,6 +88,13 @@ func NewClient(deps *module.Dependencies) Client {
 	calleeHeader := deps.Default.Config.MustGetString("sidecarRouter.default.http.calleeHeader")
 	callerName := deps.Default.Config.MustGetString("serviceName")
 	calleeName := deps.Default.Config.MustGetString("clients.corge-http.serviceName")
+
+	var altServiceDetail = config.AlternateServiceDetail{}
```
rpatali, line 92: just do `altServiceDetail := config.AlternateServiceDetail{}`

examples/example-gateway/config/production.json
```diff
@@ -27,6 +27,31 @@
...
+		  "headerValue": "^canary",
+		  "serviceName": "CorgeHttp-canary"
+		},
+		{
+		  "headerName": "x-alt-router",
+		  "headerValue": "^alternate",
+		  "serviceName": "CorgeHttp-alt"
+		}
+	  ],
+	  "servicesDetail": {
```
lzlicd, line 48: Do we even care about this section?

PR #737, opened 2020-08-10
Extend TestGateway request creation with header multi-values

vermagav:
In order to use the changes in https://github.com/uber/zanzibar/pull/733 in services that consume Zanzibar, we extend the `TestGateway` interface to mirror its existing `MakeRequest` method with one that creates a test request with all, disparate values for each header's key. We choose extending the interface instead of refactoring `MakeRequest` to preserve backwards compatibility with all of its many consumers.

### DIFF
--- a/test/lib/bench_gateway/bench_gateway.go
+++ b/test/lib/bench_gateway/bench_gateway.go
@@ -271,12 +271,12 @@ func (gateway *BenchGateway) MakeRequest(
 	fullURL := "http://" + gateway.ActualGateway.RealHTTPAddr + url
 
 	req, err := http.NewRequest(method, fullURL, body)
-	for headerName, headerValue := range headers {
-		req.Header.Set(headerName, headerValue)
-	}
-
 	if err != nil {
 		return nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


test/lib/bench_gateway/bench_gateway.go
```diff
@@ -282,6 +282,34 @@ func (gateway *BenchGateway) MakeRequest(
...
+	keys := headers.Keys()
+	for _, key := range keys {
+		if values, found := headers.Values(key); found {
+			for _, value := range values {
+				req.Header.Add(key, value)
+			}
+		}
+	}
+
+	if err != nil {
```
rpatali, line 306: any reason why this isn't right after where this happens?

rpatali: Just noticed its just how other ones are too but I don't think they need to be (or at least the new ones need to be).

vermagav: Good catch. I moved them to the correct location in both the old and new functions. Leaving it where it was also has the potential for a nil-pointer dereference with the following block accessing a property on `req`.

PR #742, opened 2020-09-15
Query param of type set implemented as slice

argouber:


runtime/server_http_request.go
```diff
@@ -516,148 +516,161 @@ func (req *ServerHTTPRequest) GetQueryValueList(key string) ([]string, bool) {
...
+	set := make(map[int8]struct{}, len(values))
 	for _, value := range values {
-		number, err := strconv.ParseInt(value, 10, 8)
+		number, err := strconv.ParseInt(value, 0, 8)
 		if err != nil {
 			req.LogAndSendQueryError(err, "int8", key, value)
 			return nil, false
 		}
-		ret[int8(number)] = _nullVal
+		set[int8(number)] = _nullVal
```
rpatali, line 543: why create a separate typedef for `struct{}{}`. In go, it's quite standard to use empty struct as value when you want to use map as a set but now one has to look up what is `_nullVal` to understand you actually want a set.

rpatali (approved):
LGTM

runtime/server_http_request.go
argouber: I did it last time because the struct{} was not well known to me -- and a stack-overflow question had that kind of thing.  No longer needed.  I just let the thing be - with minimal changes.

PR #747, opened 2020-10-22
Use context logger

shashipuchak:
- Using ContextLogger will add all the default log fields we set in context to log
- When user looks at an error log, related default context log fields will help in understanding the error better


tejaswiagarwal (approved):
Looks good!

rpatali (approved):

PR #748, opened 2020-10-28
Update use context logger

shashipuchak:


codegen/templates/grpc_client.tmpl
```diff
@@ -75,6 +75,7 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		{{ end -}}
 		opts: zanzibar.NewGRPCClientOpts(
 		deps.Default.Logger,
+		deps.Default.ContextLogger,
```
rpatali, line 78: can we not just replace the deps.Default.Logger with ContextLogger instead of adding both of them.

rpatali:
Note this is a breaking change. make sure to note that in the changes.md

rpatali (approved):
LGTM

PR #749, opened 2020-12-08
Debug code + hack to check for tchannel client side bug

argouber:
An issue we ran into got us to suspect that the tchannel call was successful, but the "success" flag was not being set by the client call in some cases.  Adding a print to catch the event and then treat it as successful.
We will remove this code once the issue is root-caused.

### DIFF
--- a/examples/example-gateway/build/clients/baz/baz.go
+++ b/examples/example-gateway/build/clients/baz/baz.go
@@ -412,6 +412,9 @@ func (c *bazClient) EchoBinary(
 
 	if err == nil && !success {
 		switch {
+		case result.Success != nil:
+			logger.Error("Internal error. Success flag is not set for EchoBinary. Overriding")
+			success = true
 		default:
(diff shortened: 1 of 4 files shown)


smb158:
Can we emit a metric when this happens? Maybe including the client name as well

smb158 (approved):

examples/example-gateway/build/clients/baz/baz.go
```diff
@@ -412,6 +412,9 @@ func (c *bazClient) EchoBinary(
 
 	if err == nil && !success {
 		switch {
+		case result.Success != nil:
+			logger.Error("Internal error. Success flag is not set for EchoBinary. Overriding")
```
rpatali, line 416: WARN perhaps and not really an error.

rpatali (approved):
LGTM, given the temporary nature of the log line anyway.

PR #753, opened 2021-02-08
add a hook to the /health API to reject traffic on SIGTERM,SIGINIT

rpatali:
When the app gets a shutdown request, the HTTP/TChannel/gRPC servers start to shutdown but the app may continue to accept requests leading to so requests being dropped. This PR catches the SIGTERM, SIGINIT signals and marks the app unhealthy preventing further requests to be received while the app is shutting down.

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -127,6 +127,7 @@ type Gateway struct {
 	tracerCloser          io.Closer
 
 	requestUUIDHeaderKey string
+	isUnhealthy          bool
 }
 
 // DefaultDependencies are the common dependencies for all modules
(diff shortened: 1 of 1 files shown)


runtime/gateway.go
```diff
@@ -127,6 +127,7 @@ type Gateway struct {
 	tracerCloser          io.Closer
 
 	requestUUIDHeaderKey string
+	isUnhealthy          bool
```
abhishekparwal, line 130: should be atomic or its read/write protected by a mutex

runtime/gateway.go
```diff
@@ -378,6 +379,14 @@ func (gateway *Gateway) handleHealthRequest(
 	req *ServerHTTPRequest,
 	res *ServerHTTPResponse,
 ) {
+	if gateway.isUnhealthy {
```
abhishekparwal, line 382: consider writing a test, calling shutdown from one goroutine and another one calling health to check the expectation?

rpatali: I thought about it but I can't see a way to make the test deterministic.

runtime/gateway.go
rpatali: good point.

runtime/gateway.go
abhishekparwal: though it may be a pretty involved effort: there can be few ways, 1 way is - to mock HTTPServer (which is a dependency to a gateway) and have it return after you return from a health call with increasing the cancel timeout to a larger value like 1 min. this way you are guaranteed to run it reliably.

rpatali: as long as some aspect of timing is involved, I wouldn't say that is a guarantee.

abhishekparwal: long cancel time is as good as a guarantee. however, if you are looking for a solution that is absolute, then you pass Shutdown context, which you can construct in your test without a timeout.

argouber (approved):
LGTM. Sufficient

PR #754, opened 2021-02-23
skip writing content when status code is 204 (StatusNoContent)

amit-lamba:


argouber (commented):
Could you add a unit-test for the 204.
Also, are there other responses that should not have body?

amit-lamba:
> Could you add a unit-test for the 204.
> Also, are there other responses that should not have body?

Added the unit-test.
Through the logs, only the one which have a 204 were having this error log.
In other cases even if the body is blank should not cause this issue.

argouber (approved):

rpatali:
> Could you add a unit-test for the 204.
> Also, are there other responses that should not have body?

304

runtime/server_http_response.go
```diff
@@ -291,7 +292,9 @@ func (res *ServerHTTPResponse) flush(ctx context.Context) {
 
 	res.flushed = true
 	res.writeHeader(res.pendingStatusCode)
-	res.writeBytes(res.pendingBodyBytes)
+	if res.pendingStatusCode != http.StatusNoContent {
```
rpatali, line 295: we use 304 for use cases so you probably want to accommodate for that status code as well.

runtime/constants.go
```diff
@@ -137,3 +137,8 @@ var knownStatusCodes = map[int]bool{
 	http.StatusNotExtended:                   true, // 510
 	http.StatusNetworkAuthenticationRequired: true, // 511
 }
+
+var noContentStatusCodes = map[int]bool{
```
rpatali, line 141: noBodyStatusCodes or something similar would be better since `noContentStatusCode` implies only 204.

rpatali (approved):

runtime/constants.go
amit-lamba: I think, Body and Content sound similar, and noBodyStatusCode**s** plural may signify that we are including all no content ones, and avoid confusion.

PR #756, opened 2021-03-03
Override env for shadow request

ashish-agg:
Overriding the environment value depending on the request type

### DIFF
--- a/examples/example-gateway/config/production.yaml
+++ b/examples/example-gateway/config/production.yaml
@@ -90,5 +90,4 @@ grpc.clientServiceNameMapping:
   echo: echo
 router.whitelistedPaths:
   - /path/whitelisted
-service.shadow.request.header: x-shadow-request
 service.shadow.env.override.enable: true
--- a/examples/example-gateway/config/production.yaml
+++ b/examples/example-gateway/config/production.yaml
@@ -90,3 +90,5 @@ grpc.clientServiceNameMapping:
   echo: echo
 router.whitelistedPaths:
   - /path/whitelisted
+service.shadow.request.header: x-shadow-request
+service.shadow.env.override.enable: true
--- a/runtime/constants.go
+++ b/runtime/constants.go
@@ -51,6 +51,11 @@ const (
 	clientHTTPUnmarshalError = "client.http-unmarshal-error"
 	// clientTchannelReadError is the metric for tracking errors in reading tchannel response
 	clientTchannelUnmarshalError = "client.tchannel-unmarshal-error"
+
+	// shadow headers and environment
+	shadowHeader 		= "X-Uber-Shadow-Request"
+	shadowEnvironment 	= "shadow"
+	environmentKey 		= "env"
 )
 
 var knownMetrics = []string{
--- a/runtime/constants.go
+++ b/runtime/constants.go
@@ -53,7 +53,7 @@ const (
 	clientTchannelUnmarshalError = "client.tchannel-unmarshal-error"
 
 	// shadow headers and environment
-	shadowHeader      = "X-Uber-Shadow-Request"
+	shadowHeader      = "X-Shadow-Request"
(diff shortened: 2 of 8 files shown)


aloksinghaluber (approved):

runtime/constants.go
```diff
@@ -51,6 +51,11 @@ const (
 	clientHTTPUnmarshalError = "client.http-unmarshal-error"
 	// clientTchannelReadError is the metric for tracking errors in reading tchannel response
 	clientTchannelUnmarshalError = "client.tchannel-unmarshal-error"
+
+	// shadow headers and environment
+	shadowHeader      = "X-Uber-Shadow-Request"
```
argouber, line 56: Open source code cannot have uber specific anything. You should probably have a generic shadow header mechanism and set it to a specific value in the uber internal use-case

runtime/server_http_request_test.go
```diff
@@ -2273,13 +2273,20 @@ func TestIncomingHTTPRequestServerLog(t *testing.T) {
 	)
 	assert.NoError(t, err)
 
-	_, err = gateway.MakeRequest("GET", "/foo?bar=bar", nil, nil)
+	var headers map[string]string
+	if isShadowRequest {
+		headers = map[string]string{
+			"X-Uber-Shadow-Request": "true",
```
argouber, line 2279: same deal as my other comment

argouber (changes requested):
Remove company-specific headers.

runtime/constants.go
ashish-agg: Valid point. I have renamed it to a more generic shadow header "X-Shadow-Request"

runtime/server_http_request_test.go
ashish-agg: done

runtime/constants.go
ashish-agg: done

examples/example-gateway/config/production.yaml
```diff
@@ -90,3 +90,5 @@ grpc.clientServiceNameMapping:
   echo: echo
 router.whitelistedPaths:
   - /path/whitelisted
+service.shadow.request.header: x-shadow-request
```
rpatali, line 93: I think these configs should go in the gateway config rather than env based config, see another similar header config example here: https://github.com/uber/zanzibar/blob/master/examples/example-gateway/build.yaml#L20

ashish-agg: Hi @rpatali , Actually I took the reference of **http.clients.requestUUIDHeaderKey: x-request-uuid**  and "**"sidecarRouter.default.http.callerHeader": "RPC-Caller"**" the way it implemented here [](https://github.com/uber/zanzibar/blob/master/examples/example-gateway/config/production.yaml#L53), but yes your point is valid that this is not environment specific and it should be present while gateway is bootstraping.
I tried to put the shadow header key in the build.yaml of example gateway but I unable to fetch it through static config.

ashish-agg: Request changes have been incorporated.

examples/example-gateway/config/production.yaml
```diff
@@ -90,3 +90,4 @@ grpc.clientServiceNameMapping:
   echo: echo
 router.whitelistedPaths:
   - /path/whitelisted
+service.shadow.env.override.enable: true
```
rpatali, line 93: why is this needed? if you have shadowRequestHeader in build.yaml, you override it otherwise you don't.

ashish-agg: As discussed with Tejas, Zanzibar is used by multiple teams(edge-gateway and presentation), etc. 
This shadow header can pass to downstream services as well but we don't want to enforce env override for them as well. So I enable it through config.

rpatali (approved):
LGTM

argouber (approved):

PR #760, opened 2021-05-19
Upgrade tally version

ashish-agg:

PR #761, opened 2021-05-20
[deps] Upgrade Tally to v3.4.0

mway:
This is a duplicate of #760, but fixes imports/type usage/tests as well.

This PR unblocks the internal monorepo upgrade of Tally, as the upstream thrift definitions were relocated. (Please don't use these, they will be internal in future versions - generate mock reporters and set up expectations instead.)


mway:
Once this is merged, we need to cut a new release so internal pins can be upgraded.

argouber (approved):

PR #768, opened 2021-06-17
Add metrics to hystrix

nkonjeti:
Goal: Time the hystrics.Doc function called in all the client template codegen file. This is to check that hystrix.Doc isn't slow and isn't taking time from the actual request. This problem was stated in this ticket: https://t3.uberinternal.com/browse/EDGE-7509 . A dashboard will be created to graph time taken by hystrix for each client in Grafana.
From baz_metrics_test.go (client: baz, method: call) : 
Screen Shot 2021-06-20 at 10 28 12 PM

- [x] Add metric for timing hystrix.DoC function called "hystrix-timer"
- [x] Add "client" and "methodName" tags to scope
- [x] Test that the "hystrix-timer" metric is included in baz_metrics_test.go


codegen/templates/tchannel_client.tmpl
```diff
@@ -266,7 +267,15 @@ type {{$clientName}} struct {
 		} else {
 			// We want hystrix ckt-breaker to count errors only for system issues
 			var clientErr error
+			scope := c.defaultDeps.Scope.Tagged(map[string]string{
+			"client" : "{{$svc.Name}}",
+			})
+		  start := time.Now()
 			err = hystrix.DoC(ctx, "{{$clientID}}", func(ctx context.Context) error {
+				t := time.Now()
+			  elapsed := t.Sub(start)
```
rpatali, line 276: **nit**: can be refactored to `time.Now().Sub(start)`. Having t as a var doesn't add extra clarity here.

codegen/templates/tchannel_client.tmpl
```diff
@@ -266,7 +267,15 @@ type {{$clientName}} struct {
 		} else {
 			// We want hystrix ckt-breaker to count errors only for system issues
 			var clientErr error
+			scope := c.defaultDeps.Scope.Tagged(map[string]string{
+			"client" : "{{$svc.Name}}",
```
rpatali, line 271: we should add the methodName too

codegen/templates/tchannel_client.tmpl
```diff
@@ -266,7 +267,15 @@ type {{$clientName}} struct {
...
+			})
+		  start := time.Now()
 			err = hystrix.DoC(ctx, "{{$clientID}}", func(ctx context.Context) error {
+				t := time.Now()
+			  elapsed := t.Sub(start)
+			  size := scope.Timer("hystrix-timer")
```
rpatali, line 277: `size` doesn't suit here, probably timer. You may even just do `scope.Timer("hystrix-timer").Record(elapsed)` without requiring a new var.

tejaswiagarwal (approved):
Looks good, thanks for adding the new test cases!

PR #769, opened 2021-06-24
Rewrite main.go to allow taking custom fx options

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Feature | Medium |

Internal JIRA: https://t3.uberinternal.com/browse/EDGE-7773

## Description

- Expose a method `GetOverrideFxOptions` to allow the callers to supply an arbitrary number of `fx` dependencies that can be used in the `main.go` generated using `main.tmpl`
- Default implementation for `GetOverrideFxOptions` is to return an empty `[] fx.Option`

## Motivation & Context
- For the API gateways based off of Zanzibar, there's *no way to wire custom fx dependencies during app initialization* other than adding it in the template because of the way `main.tmpl` is created. Changing the template isn't always feasible.

- We need to be able to rewrite `main.go` to allow a configurable number of custom `fx.options`
- Since gateways are generated code, the `app` needs to be able to override custom `fx options`

## How was this tested?
- `main.go` gets correctly updated with generated code
image

- Tested that the default example gateway runs as expected using the new generation flow.
image

- WIP testing with Uber Edge Gateway

### DIFF
--- a/examples/example-gateway/app.go
+++ b/examples/example-gateway/app.go
@@ -23,6 +23,8 @@ package app
 import (
 	"net/textproto"
 
+	"go.uber.org/fx"
+
 	"github.com/uber/zanzibar/runtime/jsonwrapper"
 
 	"go.uber.org/zap"
@@ -45,6 +47,10 @@ func getContextScopeTagExtractors() []zanzibar.ContextScopeTagsExtractor {
 	}
 
 	return extractors
+}
+
+func GetOverrideFxOptions() []fx.Option {
+	return []fx.Option{}
 }
 
 func getContextLogFieldExtractors() []zanzibar.ContextLogFieldsExtractor {
--- a/examples/example-gateway/app.go
+++ b/examples/example-gateway/app.go
@@ -41,16 +41,17 @@ var AppOptions = &zanzibar.Options{
 	JSONWrapper:               jsonwrapper.NewDefaultJSONWrapper(),
 }
 
+// GetOverrideFxOptions provides a hook to configure a non zero number of fx options
+func GetOverrideFxOptions() []fx.Option {
+	return []fx.Option{}
+}
+
 func getContextScopeTagExtractors() []zanzibar.ContextScopeTagsExtractor {
 	extractors := []zanzibar.ContextScopeTagsExtractor{
 		getRequestTags,
 	}
 
 	return extractors
-}
-
-func GetOverrideFxOptions() []fx.Option {
-	return []fx.Option{}
 }
 
 func getContextLogFieldExtractors() []zanzibar.ContextLogFieldsExtractor {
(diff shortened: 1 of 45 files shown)


glide.yaml
```diff
@@ -1,127 +1,127 @@
 package: github.com/uber/zanzibar
 import:
-- package: golang.org/x/net
```
isopropylcyanide, line 3: No changes here. Not sure why I see a diff against master here.

runtime/gateway_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestCreatGatewayLoggingConfig(t *testing.T) {
+func TestCreateGatewayLoggingConfig(t *testing.T) {
```
isopropylcyanide, line 35: Fixed typo in existing test

runtime/gateway_test.go
```diff
@@ -99,15 +99,18 @@ func TestGetServiceNameFromEnv(t *testing.T) {
...
-	os.Setenv("TEST", "overridden")
-
+	if err := os.Setenv("TEST", "overridden"); err != nil{
+		t.Errorf("no error expected but got %v", err)
+	}
 	g1, err := CreateGateway(cfg, opts)
 	assert.Nil(t, err)
 	assert.Equal(t, g1.ServiceName, "overridden")
-	os.Unsetenv("TEST")
+	if err = os.Unsetenv("TEST"); err != nil{
```
isopropylcyanide, line 108: Go linter reports warnings because `os.Unsetenv` returns an error which was not tested against. Hence adding these checks

examples/example-gateway/app.go
```diff
@@ -39,6 +41,11 @@ var AppOptions = &zanzibar.Options{
 	JSONWrapper:               jsonwrapper.NewDefaultJSONWrapper(),
 }
 
+// GetOverrideFxOptions provides a hook to configure a non zero number of fx options
+func GetOverrideFxOptions() []fx.Option {
+	return []fx.Option{}
+}
+
```
rpatali, line 48: I think adding a new field in zanzibar.Options is a more suitable way for passing such controls instead of exposing a new adhoc method. What do you think?

isopropylcyanide: Good point. I'd have liked that too there are two limiting reasons:
- fx is bound in the app prior to creating the gateway. 
- These options are more like Gateway options while we need a way to bind custom modules that may or may not be necessarily related to the gateway config.

Added some logs to share how FX is bound before gateway is created.
image

Also, `go-autoprofiler` is also a candidate for a future fx module. It'd fit `GetOverrideFxOptions` instead of `zanzibar.Options`

rpatali: that is a fair point.

rpatali (approved):
LGTM

PR #774, opened 2021-07-12
Bind pprof/trace in router

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Feature | Medium |

## Description

- Binds `trace` pprof symbol to `/debug/trace` in the router.

## Motivation & Context
- Required to trace gateway executions at load. Currently, because the router doesn't expose any, pprof trace yields nada.

## How was this tested?
- Built example gateway and tested the endpoint works as expected

image

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -339,25 +339,18 @@ func (gateway *Gateway) Bootstrap() error {
 }
 
 func (gateway *Gateway) registerPredefined() {
-	gateway.HTTPRouter.Handle("GET", "/debug/pprof", http.HandlerFunc(pprof.Index))
-	gateway.HTTPRouter.Handle("GET", "/debug/pprof/cmdline", http.HandlerFunc(pprof.Cmdline))
-	gateway.HTTPRouter.Handle("GET", "/debug/pprof/profile", http.HandlerFunc(pprof.Profile))
-	gateway.HTTPRouter.Handle("GET", "/debug/pprof/symbol", http.HandlerFunc(pprof.Symbol))
-	gateway.HTTPRouter.Handle("POST", "/debug/pprof/symbol", http.HandlerFunc(pprof.Symbol))
-	gateway.HTTPRouter.Handle(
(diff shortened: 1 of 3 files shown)


aloksinghaluber (approved):

runtime/gateway.go
```diff
@@ -339,25 +339,18 @@ func (gateway *Gateway) Bootstrap() error {
...
+	_ = gateway.HTTPRouter.Handle("GET", "/debug/pprof/profile", http.HandlerFunc(pprof.Profile))
+	_ = gateway.HTTPRouter.Handle("POST", "/debug/pprof/trace", http.HandlerFunc(pprof.Trace))
+	_ = gateway.HTTPRouter.Handle("GET", "/debug/pprof/symbol", http.HandlerFunc(pprof.Symbol))
+	_ = gateway.HTTPRouter.Handle("POST", "/debug/pprof/symbol", http.HandlerFunc(pprof.Symbol))
+	_ = gateway.HTTPRouter.Handle("GET", "/debug/pprof/goroutine", pprof.Handler("goroutine"))
+	_ = gateway.HTTPRouter.Handle("GET", "/debug/pprof/heap", pprof.Handler("heap"))
+	_ = gateway.HTTPRouter.Handle("GET", "/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
+	_ = gateway.HTTPRouter.Handle("GET", "/debug/pprof/block", pprof.Handler("block"))
+	_ = gateway.HTTPRouter.Handle("GET", "/debug/loglevel", gateway.atomLevel)
+	_ = gateway.HTTPRouter.Handle("PUT", "/debug/loglevel", gateway.atomLevel)
```
rpatali, line 353: **nit**: if you are not using the response at all, I think it's better to use the previous syntax instead of using the unnamed var.

isopropylcyanide: Previous syntax had go lint errors. Here, the response of `Handle()` is an `err` which wasn't declared. Since we don't care about the error we specify the `_` to silence the linter.

PR #775, opened 2021-07-14
Change circuit breaking naming to client id and method name

nkonjeti:
Circuit breaker naming was changed to prevent different clients with same method names to have overlap in circuit breaker key values. The recent circuit breaker change configured the circuit breaker settings to be stored by the method name. In prod a client had a duplicate client staging file with the same method names, which led to the timeout value of the non staging client to have its circuit breaker have the timeout value of 5 seconds instead of 15 seconds. This change to name the circuit breakers by client id + '-' + method name will allow naming to be unique and prevent the overlap issue. Also testing was added to baz and contacts test files to check if the correct settings were set for all method circuit breakers.

### DIFF
--- a/codegen/templates/grpc_client.tmpl
+++ b/codegen/templates/grpc_client.tmpl
@@ -76,8 +76,9 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		circuitBreakerDisabled = deps.Default.Config.MustGetBoolean("clients.{{$clientID}}.circuitBreakerDisabled")
 	}
 	if !circuitBreakerDisabled {
-		for methodKey := range methodNames {
-			configureCircuitBreaker(deps, timeoutInMS, methodNames[methodKey])
+		for _, methodName := range methodNames {
+			circuitBreakerName := "{{$clientID}}"  + "-" + methodName
+			configureCircuitBreaker(deps, timeoutInMS, circuitBreakerName)
 		}
 	}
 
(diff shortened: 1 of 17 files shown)


codegen/templates/grpc_client.tmpl
```diff
@@ -76,8 +76,9 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		circuitBreakerDisabled = deps.Default.Config.MustGetBoolean("clients.{{$clientID}}.circuitBreakerDisabled")
 	}
 	if !circuitBreakerDisabled {
-		for methodKey := range methodNames {
-			configureCircuitBreaker(deps, timeoutInMS, methodNames[methodKey])
+		for _, methodName := range methodNames {
+			circuitBreakerName := "{{$clientID}}"  + "-" + methodName
```
tejaswiagarwal, line 80: Are we doing something different for grpc_client.tmpl and http_client.tmpl? http Client has methodKey vs here we are using methodName

nkonjeti: Here I'm doing methodName (the value of the map) because the map of methods has the method name as the value and the service method as the key. 
Screen Shot 2021-07-14 at 9 28 08 AM
But in http_client.tmpl I use the key value of the map because the map has the methodname as the value.
Screen Shot 2021-07-14 at 9 27 00 AM
I could change the naming so both files have methodName as the variable in the loop, what do you think?

tejaswiagarwal: Thanks for clarifying! That makes sense. Yeah - I think keeping it consistent across templates would improve readability.

tejaswiagarwal (approved):
Looks good!

PR #777, opened 2021-07-28
Code gen changes for qps levels in circuit breaking (work in progress) 

nkonjeti:
Draft for configuring circuit breaker parameters by qps levels

DONE:
- added qps levels to test endpoint yaml files 
- added qpsLevels variable to client meta config 
- mapped client method name to qps level by going through endpoint yaml files
- added qps levels to client template files (which maps method name to qps level)
- read from test.yaml to get circuit breaker parameters for qps level
- add test to check correct qps levels are populated and circuit breaker parameters are configured


codegen/module_system.go
```diff
@@ -485,7 +488,15 @@ func (g *httpClientGenerator) Generate(
...
-
+	// transfer only the methods that belong to the client with the qps level
+	var clientQPSLevels map[string]int = make(map[string]int)
+	for _, v := range exposedMethods {
+		if qps, ok := qpsLevels[v]; ok {
+			clientQPSLevels[v] = qps
+		}
+	}
+	// adding QPS Levels here will allow access in client go file
+	print(instance.BaseDirectory)
```
tejaswiagarwal, line 499: Remove?

codegen/templates/tchannel_client.tmpl
```diff
@@ -127,6 +128,14 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		{{ end -}}
 	}
 
+	levels := map[string]string{
+				{{range $methodName, $qpsLevel := $QPSLevels -}}
+				"{{$methodName}}": "{{$qpsLevel}}",
+				{{end}}
+	}
+	// had to use levels variable to get over error
+	print(levels)
```
tejaswiagarwal, line 137: Remove from the template?

codegen/templates/http_client.tmpl
```diff
@@ -126,6 +127,15 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		}
 	}
 
+	levels := map[string]string{
```
tejaswiagarwal, line 130: It looks like we are not using this eventually to set the different parameters in the circuit breaker, is that part of a follow-up diff?

codegen/module_system.go
```diff
@@ -59,6 +60,8 @@ type EndpointMeta struct {
 	DefaultHeaders         []string
 }
 
+var qpsLevels map[string]int = make(map[string]int)
```
tejaswiagarwal, line 63: Does this need to be a global?

tejaswiagarwal (commented):
Looks good overall. Can we add an additional check in the code that is QPSLevels is missing it just defaults to a default level and applies default values? This will help with backward compatibility issues if someone upgrades their Zanzibar dependency.

codegen/module_system.go
```diff
@@ -618,6 +638,19 @@ func (g *tchannelClientGenerator) Generate(
 
 	sort.Sort(clientSpec.ModuleSpec.Services)
 
+	var clientQPSLevels map[string]string = make(map[string]string)
+	for _, methodName := range exposedMethods {
```
tejaswiagarwal, line 642: Since we are doing this in multiple places, can we abstract this into a function?

codegen/module_system.go
```diff
@@ -1014,6 +1062,7 @@ func (g *EndpointGenerator) ComputeSpec(
 	wg.Add(len(endpointYamls))
 	ch := make(chan endpointSpecRes, len(endpointYamls))
 	for _, yamlFile := range endpointYamls {
+		UpdateQPSLevels(yamlFile)
```
tejaswiagarwal, line 1065: Can we do this inside the go func(). That way you might not need a mutex below.

codegen/module_system.go
```diff
@@ -1049,6 +1098,45 @@ func (g *EndpointGenerator) ComputeSpec(
...
 }
 
+// UpdateQPSLevels updates map from client-method name to qps level
+func UpdateQPSLevels(yamlFile string) {
+	var config map[string]interface{}
+	file, err := ioutil.ReadFile(yamlFile)
+	if err != nil {
+		fmt.Printf("error reading file to update qps level")
+	}
+	if err == nil {
```
tejaswiagarwal, line 1108: If we reach here, we don't really need to check err == nil.

codegen/module_system.go
```diff
@@ -1049,6 +1098,45 @@ func (g *EndpointGenerator) ComputeSpec(
...
+	qpsLevel, qpsOK := config["qpsLevel"]
+	if methodOK && clientOK && qpsOK && err == nil {
+		// edge case where clientID or clientMethod is nil
+		if clientID != nil && clientMethod != nil {
+			// unique key because of potential clients having same method names (staging)
+			key := clientID.(string) + "-" + clientMethod.(string)
+			currentQPSLevel := qpsLevels[key]
+			// store highest qps level for circuit breaker in qpsLevels map
+			thisQPSLevel := int(qpsLevel.(float64))
+			if thisQPSLevel > currentQPSLevel {
```
tejaswiagarwal, line 1124: This can be something like this:

```
mutex.Lock()
if ()
 {
}
else {
}
mutex.Unlock()
```

That way we are doing lock/unlock only once.

codegen/templates/grpc_client.tmpl
```diff
@@ -78,7 +85,11 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	if !circuitBreakerDisabled {
 		for _, methodName := range methodNames {
 			circuitBreakerName := "{{$clientID}}"  + "-" + methodName
-			configureCircuitBreaker(deps, timeoutInMS, circuitBreakerName)
+			qpsLevel := ""
```
tejaswiagarwal, line 88: Should this be "default"? That we will never send "" to configureCircuitBreaker which might break things

codegen/templates/http_client.tmpl
```diff
@@ -165,27 +175,48 @@ func initializeAltRoutingMap(altServiceDetail config.AlternateServiceDetail) map
...
+	var sleepWindowInMilliseconds int
+	// maxConcurrentRequests sets how many requests can be run at the same time, beyond which requests are rejected
+	var maxConcurrentRequests int
+	// errorPercentThreshold sets the error percentage at or above which the circuit should trip open
+	var errorPercentThreshold int
+	// requestVolumeThreshold sets a minimum number of requests that will trip the circuit in a rolling window of 10s
+	// For example, if the value is 20, then if only 19 requests are received in the rolling window of 10 seconds
+	// the circuit will not trip open even if all 19 failed.
+	var requestVolumeThreshold int
+	key := "circuitbreaking-configurations"
```
tejaswiagarwal, line 195: Can you make this a constant at the top of the template file?

test/endpoints/baz/baz_simpleservice_method_call_test.go
```diff
@@ -100,26 +97,6 @@ func BenchmarkCall(b *testing.B) {
 		}
 	})
 
-	// test circuit breaker settings for each method match prod values
-	settings := hystrix.GetCircuitSettings()
-	names := [27]string{"baz-EchoBinary", "baz-EchoBool", "baz-EchoDouble", "baz-EchoEnum", "baz-EchoI16", "baz-EchoI32", "baz-EchoI64", "baz-EchoI8", "baz-EchoString", "baz-EchoStringList", "baz-EchoStringMap", "baz-EchoStringSet", "baz-EchoStructList", "baz-EchoStructSet", "baz-EchoTypedef", "baz-Call", "baz-Compare", "baz-GetProfile", "baz-HeaderSchema", "baz-Ping", "baz-DeliberateDiffNoop", "baz-TestUUID", "baz-Trans", "baz-TransHeaders", "baz-TransHeadersNoReq", "baz-TransHeadersType", "baz-URLTest"}
```
tejaswiagarwal, line 105: Can we keep this test and only let the "default" methods be here?

codegen/module_system.go
nkonjeti: I tried moving it to inside go func() but it was failing codegen tests.

codegen/module_system.go
```diff
@@ -544,6 +573,57 @@ func (g *httpClientGenerator) Generate(
...
 
+// PopulateQPSLevels loops through endpoint dir and gets qps levels
+func PopulateQPSLevels(EndpointsBaseDir string) map[string]int {
+	qpsLevels := make(map[string]int)
+	endpointFiles := GetListOfAllFilesInEndpointDir(EndpointsBaseDir)
+	for _, endpointFile := range endpointFiles {
+		var config map[string]interface{}
+		bytes, _ := ioutil.ReadFile(endpointFile)
+		err := yaml.Unmarshal(bytes, &config)
+		if err != nil {
```
tejaswiagarwal, line 584: We can unmarshal here by checking the file type:

fileExtension := filepath.Ext(path)

if fileExtension == ".json" {
    // Read JSON
}

codegen/module_system.go
```diff
@@ -544,6 +573,57 @@ func (g *httpClientGenerator) Generate(
...
+			// try json
+			_ = json.Unmarshal(bytes, &config)
```
tejaswiagarwal, line 586: catch error here?

codegen/module_system.go
```diff
@@ -544,6 +573,57 @@ func (g *httpClientGenerator) Generate(
...
+		clientMethod, methodOK := config["clientMethod"]
+		clientID, clientOK := config["clientId"]
+		qpsLevel, qpsOK := config["qpsLevel"]
+		if methodOK && clientOK && qpsOK && err == nil {
+			// edge case where clientID or clientMethod is nil
+			if clientID != nil && clientMethod != nil && clientID != "" && clientMethod != "" {
+				// unique key because of potential clients having same method names (staging)
+				key := clientID.(string) + "-" + clientMethod.(string)
+				// store highest qps level for circuit breaker in qpsLevels map
+				thisQPSLevel := int(qpsLevel.(float64))
```
tejaswiagarwal, line 597: This should be int instead of float?

codegen/module_system.go
```diff
@@ -544,6 +573,57 @@ func (g *httpClientGenerator) Generate(
...
+		if err != nil {
+			// try json
+			_ = json.Unmarshal(bytes, &config)
+		}
+		clientMethod, methodOK := config["clientMethod"]
+		clientID, clientOK := config["clientId"]
+		qpsLevel, qpsOK := config["qpsLevel"]
+		if methodOK && clientOK && qpsOK && err == nil {
+			// edge case where clientID or clientMethod is nil
+			if clientID != nil && clientMethod != nil && clientID != "" && clientMethod != "" {
```
tejaswiagarwal, line 593: Do we want  || or && checks here?

codegen/module_system.go
```diff
@@ -544,6 +573,57 @@ func (g *httpClientGenerator) Generate(
...
+				// unique key because of potential clients having same method names (staging)
+				key := clientID.(string) + "-" + clientMethod.(string)
+				// store highest qps level for circuit breaker in qpsLevels map
+				thisQPSLevel := int(qpsLevel.(float64))
+				if currentQPSLevel, ok := qpsLevels[key]; ok {
+					if thisQPSLevel > currentQPSLevel {
+						qpsLevels[key] = thisQPSLevel
+					}
+				} else {
```
tejaswiagarwal, line 602: The else {} is redundant. We can remove the else statement

codegen/module_system.go
```diff
@@ -544,6 +573,57 @@ func (g *httpClientGenerator) Generate(
...
+	}
+	return qpsLevels
+}
+
+// GetListOfAllFilesInEndpointDir gets all the endpoint config files
+func GetListOfAllFilesInEndpointDir(baseDir string) []string {
+	filesList := []string{}
+	items, _ := ioutil.ReadDir(baseDir)
+	for _, item := range items {
+		subitems, _ := ioutil.ReadDir(baseDir + "/" + item.Name())
```
tejaswiagarwal, line 616: catch err and return appropriate err here?

codegen/module_system.go
nkonjeti: That is if the circuit breaker doesn't already exist in the map so then it initializes the value

codegen/module_system.go
nkonjeti: It parses it as a float so I had to convert it to an int

codegen/module_system.go
nkonjeti: I was thinking && so it only goes in the if block if clientID and clientMethod are valid

codegen/module.go
```diff
@@ -1267,7 +1267,11 @@ func (system *ModuleSystem) IncrementalBuild(
 	for _, moduleList := range toBeBuiltModules {
 		moduleCount += len(moduleList)
 	}
-
+	qpsLevels, err := PopulateQPSLevels(baseDirectory + "/endpoints")
+	if err != nil {
+		fmt.Printf("Error in populating qps levels")
```
tejaswiagarwal, line 1272: I think you can do something like this here: 

`return nil, return nil, errors.Wrapf(err, "error in populating qps levels")

codegen/module_system.go
```diff
@@ -544,6 +567,121 @@ func (g *httpClientGenerator) Generate(
...
+	return config, nil
+}
+
+// GetListOfAllFilesInEndpointDir gets all the endpoint config files
+// takes empty filesList array as parameter to add to during recursion
+func GetListOfAllFilesInEndpointDir(filePath string, filesList []string) ([]string, error) {
+	fileInfo, err := os.Stat(filePath)
+	if err != nil {
+		return nil, errors.Errorf(
+			"Error in doing Stat with file path %q",
```
tejaswiagarwal, line 658: Go error strings should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings Can we make this "error" in all places? Thanks for adding error handling in all places

codegen/module_system.go
```diff
@@ -544,6 +567,121 @@ func (g *httpClientGenerator) Generate(
...
+					qpsLevels[key] = thisQPSLevel
+				}
+			}
+		}
+	}
+	return qpsLevels, nil
+}
+
+// UnMarshalEndpointFile unmarshals endpoint file into config
+func UnMarshalEndpointFile(endpointFile string) (map[string]interface{}, error) {
```
tejaswiagarwal, line 618: Let's call this UnmarshalEndpointFile :)

tejaswiagarwal (approved):
Thanks for addressing all the comments - looks good!

PR #764, opened 2021-05-28
fix race condition in shutdown

amdalal:
Currently http, tchannel servers and grpc client dispatcher are stopped in parallel. It is possible that gprc client dispatcher closes all outbounds while http/tchannel servers are still serving requests leading to failures.

This change waits for http/tchannel servers to shutdown before closing gprc client dispatcher.

**Test:**
_To reproduce:_

- Add artificial slowness in tchannel shutdown
```
go func() {
		defer swg.Done()
		gateway.Logger.Info("sleeping before shutting down tchannel server")
		time.Sleep(1 * time.Minute)
		gateway.Logger.Info("resuming tchannel server shutdown")
		if err := gateway.shutdownTChannelServer(ctx); err != nil {
			ec <- errors.Wrap(err, "error shutting down tchannel server")
		}
	}()
```
 - Issue `kill -15` on the pid
 - invoke `bounce` endpoint while tchannel shutdown in ongoing. 
 ```
 yab -s selective-gateway -m Bounce::bounce -t $GOPATH/src/github.com/uber/zanzibar/examples/selective-gateway/idl/endpoints/bounce/bounce.thrift --caller yab-amdalal --request '{"msg":"bdfb7a20-d7fa-48f4-96f3-d175a49fe2ae"}' -header 'x-uber-source:curl' --timeout=30000 -p 127.0.0.1:5050
```
- Error seen
```
`{"level":"error","ts":1622190556.827372,"msg":"Endpoint failure: handler returned error","zone":"unknown","env":"test","hostname":"amdalal-C02YVFT0LVCG","service":"selective-gateway","pid":83591,"endpointID":"bounce","endpointHandler":"bounce","endpointThriftMethod":"Bounce::bounce","x-uber-source":"curl","requestUUID":"848a7e6d-faa6-497d-831d-a7a229dcf76a","error":"code:failed-precondition message:error waiting for grpc outbound to start for service: echo: could not wait for instance to start running: current state is \"stopped\""}`

```

After the fix, calls to `echo` go through.

shutdown logs
```{"level":"info","ts":1622196809.018711,"msg":"Channel.Close called.","serviceName":"selective-gateway","process":"selective-gateway","chID":1,"hostPort":"127.0.0.1:5050"}
{"level":"info","ts":1622196809.018789,"msg":"Channel closed.","serviceName":"selective-gateway","process":"selective-gateway","chID":1,"hostPort":"127.0.0.1:5050"}
[...]

PR #778, opened 2021-07-29
Dedicated TChannel Client Per Callee Service

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Feature | High |

Internal JIRA: https://t3.uberinternal.com/browse/EDGE-326 (ERD attached in the ticket)

### Changed
- **BREAKING** `gateway.Channel` has been renamed to `gateway.ServerTChannel` to distinguish between client and server TChannels.

- **BREAKING** `gateway.TChannelRouter` has been renamed to `gateway.ServerTChannelRouter`

### Added
- A new boolean flag `dedicated.tchannel.client` is introduced. When set to `true`, each TChannel client creates a dedicated connection instead of registering on the default shared server connection.
Clients need not do anything as the default behaviour is to use a shared connection.([#778](https://github.com/uber/zanzibar/pull/778))

- A new field has been added to the gateway to propagate the `TChannelSubLoggerLevel` for clients with dedicated connections.

---

## Reviewer Notes
- Only 6-7 files need a review. Rest are generated / tests
- Estimated time to review: 20 mins given that you've read the internal ERD :)

## Description

- Changes the way Edge `TChannel` Clients connect with the sidecar. Instead of using a common gateway channel and router, we create dedicated channels for each client. This is done based on the flag `dedicated.tchannel.client: true`
image

- Shutdown hooks now include the shutdown for clients in a separate goroutine.
- Refactor existing gateway channel and router to more readable names
- **Does not** change templates for tchannel endpoints. They still get registered on the common server router. Only clients get new routers.
- Fixes some formatting consistencies and typos.

## Motivation & Context
- Current TChannel clients speak like below
image

- Maintaining a dedicated connection ensures that in the event of a misbehaving downstream, requests to other “healthy” tchannel clients do not time out. Only the backlog requests in the queue for the unhealthy service would be affected if the timeout expires. This leads to a more fault tolerant request layer in Edge. As a Tier 0 service, this offers significant platform stability.

- Since each queue would be separate and multiple router-sidecar goroutines would be reading requests in parallel, we expect good improvements in request latency.

- Slight although insignificant improvements may be seen in CPU (more utilization) and memory (reduced pressure of requests in the queue).

## How was this tested?
Tested this version of Zanzibar in Uber's Edge Gateway and verifying that 
- Multiple tchannel clients in Edge get their own channels and are able to make RPCs
- Statsd metrics are emitted
```
tchannel.client.running+client=bazService,dc=unknown,env=test,service=test-gateway
      Value:v2.MetricValue{MetricType:2, Count:0, Gauge:1, Timer:0},..., 
```
- Shutdown hooks are called

Tested that conditional client generation works and when `dedicated.tchannel.client: false | nil`, the existing behaviour remains true

## Benchmarks

**Verdict:** No considerable change in op/sec. Zero change in memory allocs

### Branch
```

========

Branch
time -p sh -c "go test -run _NONE_ -bench . -benchmem -benchtime 7s -cpu 2 ./test/... | grep -v '^ok ' | grep -v '\[no test files\]' | grep -v '^PASS'"
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  108061	     67386 ns/op	   21593 B/op	     165 allocs/op
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test/endpoints/baz
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
   36069	    232823 ns/op	   85245 B/op	     716 allocs/op
   36259	    229871 ns/op	   80946 B/op	     713 allocs/op
   39594	    213469 ns/op	   74620 B/op	     671 allocs/op
   38223	    213605 ns/op	   74171 B/op	     657 allocs/op
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test/endpoints/contacts
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
   42102	    190982 ns/op	   73774 B/op	     559 allocs/op
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test/endpoints/googlenow
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
   52650	    155615 ns/op	   63501 B/op	     441 allocs/op
real 96.87
user 121.16
sys 39.36
========
````

### Master

```
========
time -p sh -c "go test -run _NONE_ -bench . -benchmem -benchtime 7s -cpu 2 ./test/... | grep -v '^ok ' | grep -v '\[no test files\]' | grep -v '^PASS'"
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  126316	     56844 ns/op	   21978 B/op	     165 allocs/op
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test/endpoints/baz
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
   35053	    228300 ns/op	   88222 B/op	     716 allocs/op
   36102	    226970 ns/op	   83794 B/op	     713 allocs/op
   39868	    213884 ns/op	   76952 B/op	     671 allocs/op
   39051	    214590 ns/op	   76357 B/op	     657 allocs/op
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test/endpoints/contacts
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
   42651	    195224 ns/op	   78067 B/op	     559 allocs/op
goos: darwin
goarch: amd64
pkg: github.com/uber/zanzibar/test/endpoints/googlenow
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
   51500	    151183 ns/op	   60568 B/op	     441 allocs/op
real 98.17
user 129.94
sys 42.91
========
```

### DIFF
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -69,7 +69,7 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	}
 	gateway := deps.Default.Gateway
 	channel := createNewTchannelForClient(deps, serviceName)
-	gateway.ClientTchannels[serviceName] = channel
+	gateway.ClientTChannels[serviceName] = channel
 
 	{{if $sidecarRouter -}}
 	ip := deps.Default.Config.MustGetString("sidecarRouter.{{$sidecarRouter}}.tchannel.ip")
@@ -186,7 +186,7 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 func createNewTchannelForClient(deps *module.Dependencies, serviceName string) *tchannel.Channel {
 	processName := deps.Default.Config.MustGetString("tchannel.processName")
 	gateway := deps.Default.Gateway
-	level := gateway.TchannelSubLoggerLevel
+	level := gateway.TChannelSubLoggerLevel
 
 	channel, err := tchannel.NewChannel(
 		serviceName,
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -67,18 +67,29 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	if deps.Default.Config.ContainsKey("tchannel.clients.requestUUIDHeaderKey") {
 		requestUUIDHeaderKey = deps.Default.Config.MustGetString("tchannel.clients.requestUUIDHeaderKey")
 	}
-	gateway := deps.Default.Gateway
-	channel := createNewTchannelForClient(deps, serviceName)
-	gateway.ClientTChannels[serviceName] = channel
 
 	{{if $sidecarRouter -}}
-	ip := deps.Default.Config.MustGetString("sidecarRouter.{{$sidecarRouter}}.tchannel.ip")
-	port := deps.Default.Config.MustGetInt("sidecarRouter.{{$sidecarRouter}}.tchannel.port")
+		ip := deps.Default.Config.MustGetString("sidecarRouter.{{$sidecarRouter}}.tchannel.ip")
+		port := deps.Default.Config.MustGetInt("sidecarRouter.{{$sidecarRouter}}.tchannel.port")
 	{{else -}}
-	ip := deps.Default.Config.MustGetString("clients.{{$clientID}}.ip")
-	port := deps.Default.Config.MustGetInt("clients.{{$clientID}}.port")
+		ip := deps.Default.Config.MustGetString("clients.{{$clientID}}.ip")
+		port := deps.Default.Config.MustGetInt("clients.{{$clientID}}.port")
 	{{end -}}
-	channel.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+
+	gateway := deps.Default.Gateway
+	var channel *tchannel.Channel
+
+	// If dedicated.tchannel.client : true, each tchannel client will create a
+	// dedicated connection with local Muttley, else it will use a shared connection
+	if deps.Default.Config.ContainsKey("dedicated.tchannel.client") &&
+		deps.Default.Config.MustGetBoolean("dedicated.tchannel.client") {
+		channel = deps.Default.ServerTChannel
+		channel.GetSubChannel(serviceName, tchannel.Isolated).Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+	} else {
+		channel = createNewTchannelForClient(deps, serviceName)
+		channel.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+		gateway.ClientTChannels[serviceName] = channel
+	}
 
 	/*Ex:
 	{
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -69,8 +69,8 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	}
 	gateway := deps.Default.Gateway
 	channel := createNewTchannelForClient(deps, serviceName)
-	gateway.TchannelChannels[serviceName] = channel
-	gateway.TchannelRouters[serviceName] = zanzibar.NewTChannelRouter(channel, gateway)
+	gateway.ClientTchannels[serviceName] = channel
+	gateway.ClientTchannelRouters[serviceName] = zanzibar.NewTChannelRouter(channel, gateway)
 
 	{{if $sidecarRouter -}}
 	ip := deps.Default.Config.MustGetString("sidecarRouter.{{$sidecarRouter}}.tchannel.ip")
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -80,7 +80,7 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	var channel *tchannel.Channel
 
 	// If dedicated.tchannel.client : true, each tchannel client will create a
-	// dedicated connection with local Muttley, else it will use a shared connection
+	// dedicated connection with local sidecar, else it will use a shared connection
 	if deps.Default.Config.ContainsKey("dedicated.tchannel.client") &&
 		deps.Default.Config.MustGetBoolean("dedicated.tchannel.client") {
 		channel = deps.Default.ServerTChannel
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -86,7 +86,7 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		channel = gateway.SetupClientTChannel(deps.Default.Config, serviceName)
 		channel.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
 	} else {
-		channel = deps.Default.ServerTChannel
+		channel = gateway.ServerTChannel
 		channel.GetSubChannel(serviceName, tchannel.Isolated).Peers().Add(ip + ":" + strconv.Itoa(int(port)))
 	}
 
--- a/codegen/templates/tchannel_client.tmpl
+++ b/codegen/templates/tchannel_client.tmpl
@@ -133,6 +133,10 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		{{ end -}}
 		{{ end -}}
 	}
+	for _, method := range methodNames {
+		//todo we were asserting that the registering of methods happen in separate tchannels
+		fmt.Printf("For Client: {{$clientName}} we are registering %v on the explicit client channel\n", method)
+	}
 
 	qpsLevels := map[string]string{
 				{{range $methodName, $qpsLevel := $QPSLevels -}}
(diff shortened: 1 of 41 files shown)

### CODE REVIEW

File: test/endpoints/bar/bar_metrics_test.go [hunk 1]
```diff
@@ -62,7 +62,7 @@ func TestCallMetrics(t *testing.T) {
 		},
 	)
 
-	numMetrics := 12
+	numMetrics := 13
```
isopropylcyanide, line 65: This is because we added a new gauge

runtime/gateway.go
```diff
@@ -89,25 +89,27 @@ type Options struct {
...
+	ContextExtractor       ContextExtractor
+	RootScope              tally.Scope
+	Logger                 *zap.Logger
+	ServiceName            string
+	Config                 *StaticConfig
+	HTTPRouter             HTTPRouter
+	ServerTChannelRouter   *TChannelRouter
+	TChannelSubLoggerLevel zapcore.Level
+	Tracer                 opentracing.Tracer
+	JSONWrapper            jsonwrapper.JSONWrapper
```
isopropylcyanide, line 113: The diff here might make it hard to understand what has changed in the struct. Refer below

```
Channel -> .ServerTChannel
TChannelRouter -> ServerTChannelRouter
+ TChannelSubLoggerLevel
+ ClientTChannels
```

runtime/gateway.go
```diff
@@ -800,6 +803,7 @@ func (gateway *Gateway) setupTChannel(config *StaticConfig) error {
 	if !ok {
 		return errors.Errorf("unknown sub logger level for tchannel server: %s", levelString)
 	}
+	gateway.TChannelSubLoggerLevel = level
```
isopropylcyanide, line 807: Required in case `dedicated.tchannel.client: true` in which case we need this config while creating clients in `init phase`

test/endpoints/bar/bar_metrics_test.go
isopropylcyanide, line 65: This is because we added a new gauge `tchannel.clients.running`

runtime/gateway.go
```diff
@@ -422,7 +425,7 @@ func (gateway *Gateway) Shutdown() {
 	swg.Add(1)
 	go func() {
 		defer swg.Done()
-		if err := gateway.shutdownTChannelServer(ctx); err != nil {
+		if err := gateway.shutdownTChannelServerAndClients(ctx); err != nil {
 			ec <- errors.Wrap(err, "error shutting down tchannel server")
```
amit-lamba, line 429: The error string can also be improved

runtime/gateway.go
amit-lamba: qq: How is logger level related to TChannel?

examples/example-gateway/config/production.yaml
```diff
@@ -122,4 +122,7 @@ circuitbreaking-configurations:
       sleepWindowInMilliseconds: 1000
       errorPercentThreshold: 30
       requestVolumeThreshold: 25
-      maxConcurrentRequests: 30
\ No newline at end of file
+      maxConcurrentRequests: 30
+
+service.shadow.env.override.enable: true
+dedicated.tchannel.client: true
```
amit-lamba, line 128: qq: How have we handled the default behaviour? Does it remain the same? Should the example be for default / old behaviour or new behaviour?

isopropylcyanide: - Default behaviour: Same as before
- `Example-gateway` example has the new behaviour. `Selective-gateway` example has the default behaviour

runtime/gateway.go
isopropylcyanide: - When we create a new TChannel, we need to specify a `tchannel.Logger`. 
- To pass this, we create a new TChannelLogger value. 
- To pass this value, we create a sublogger based on `gateway.logger` which essentially clones the original logger. 
- The level we pass to this logger `debug|info|...` is determined by `gateway. TChannelSubLoggerLevel`
- `gateway.TChannelSubLoggerLevel` is populated by `subLoggerLevel.tchannel`

```
func NewTChannelLogger(logger *zap.Logger) tchannel.Logger {}
func (gateway *Gateway) SubLogger(name string, level zapcore.Level) *zap.Logger {}
```

runtime/gateway.go
isopropylcyanide: Updated

codegen/templates/tchannel_client.tmpl
```diff
@@ -67,16 +67,29 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
+		ip := deps.Default.Config.MustGetString("clients.{{$clientID}}.ip")
+		port := deps.Default.Config.MustGetInt("clients.{{$clientID}}.port")
 	{{end -}}
-	sc.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+
+	gateway := deps.Default.Gateway
+	var channel *tchannel.Channel
+
+	// If dedicated.tchannel.client : true, each tchannel client will create a
+	// dedicated connection with local Muttley, else it will use a shared connection
```
amdalal, line 83: nit: let's not mention "local Muttley" here.

isopropylcyanide: Agreed. Missed it here.

codegen/templates/tchannel_client.tmpl
```diff
@@ -67,16 +67,29 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
+	// If dedicated.tchannel.client : true, each tchannel client will create a
+	// dedicated connection with local sidecar, else it will use a shared connection
+	if deps.Default.Config.ContainsKey("dedicated.tchannel.client") &&
+		deps.Default.Config.MustGetBoolean("dedicated.tchannel.client") {
+		channel = deps.Default.ServerTChannel
+		channel.GetSubChannel(serviceName, tchannel.Isolated).Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+	} else {
+		channel = createNewTchannelForClient(deps, serviceName)
+		channel.Peers().Add(ip + ":" + strconv.Itoa(int(port)))
+		gateway.ClientTChannels[serviceName] = channel
```
amdalal, line 91: Is this ok given one service can have multiple clients?

runtime/gateway.go
```diff
@@ -909,17 +906,29 @@ func GetHostname() string {
...
 	if gateway.Config.ContainsKey("shutdown.pollInterval") {
 		shutdownPollInterval = time.Duration(gateway.Config.MustGetInt("shutdown.pollInterval")) * time.Millisecond
 	}
 	ticker := time.NewTicker(shutdownPollInterval)
 	defer ticker.Stop()
 
+	for serviceName, clientTchannel := range gateway.ClientTChannels {
+		go func(service string, ch *tchannel.Channel) {
+			gateway.Logger.Info(fmt.Sprintf("Closing tchannel client for [%v]", service))
+			ch.Close()
```
amdalal, line 923: This could lead to a situation where client is closed but server is not? Leading to failures in pending outbound calls during shtudown (similar to https://github.com/uber/zanzibar/pull/764).

Can we close all the servers first, wait for their server.close() to complete, and then shutdown clients?

codegen/templates/tchannel_client.tmpl
```diff
@@ -180,7 +194,35 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	}
 }
 
-func initializeDynamicChannel(deps *module.Dependencies, headerPatterns []string, altChannelMap map[string]*tchannel.SubChannel, re ruleengine.RuleEngine) ([]string, ruleengine.RuleEngine) {
+func createNewTchannelForClient(deps *module.Dependencies, serviceName string) *tchannel.Channel {
```
rpatali, line 197: this doesn't depend on the template so let's move this out so you don't have to generated this several times during the code-gen step.

runtime/gateway.go
rpatali: +1

codegen/templates/tchannel_client.tmpl
rpatali: probably someplace here: https://github.com/uber/zanzibar/blob/master/runtime/tchannel_client.go

runtime/gateway.go
isopropylcyanide: Good catch. Updating

codegen/templates/tchannel_client.tmpl
isopropylcyanide: Didn't get it. The clients is the local tchannel client  for each backend

codegen/templates/tchannel_client.tmpl
isopropylcyanide: Will lead to an explosion of parameters in the `tchannel_client.go` as it needs a tracer, logger etc.

I found a better place for it in `gateway.go` which will also clean up the `defaultDependencies struct` a bit and would avoid repeated codegen. Thanks for the input.

codegen/templates/tchannel_client.tmpl
isopropylcyanide: As discussed offline, this is the required behaviour. For e.g presentation services where multiple tchannel clients talk to the same presentation backend.

amit-lamba (approved):

amdalal (approved):

PR #779, opened 2021-08-06
Skip zanzibar logging based on flag 2

amit-lamba:


runtime/context.go
```diff
@@ -245,18 +260,18 @@ func (c *ContextExtractors) ExtractLogFields(ctx context.Context) []zap.Field {
 
 // ContextLogger is a logger that extracts some log fields from the context before passing through to underlying zap logger.
 type ContextLogger interface {
-	Debug(ctx context.Context, msg string, fields ...zap.Field)
-	Error(ctx context.Context, msg string, fields ...zap.Field)
-	Info(ctx context.Context, msg string, fields ...zap.Field)
-	Panic(ctx context.Context, msg string, fields ...zap.Field)
-	Warn(ctx context.Context, msg string, fields ...zap.Field)
+	Debug(ctx context.Context, msg string, fields ...zap.Field) context.Context
```
isopropylcyanide, line 263: Now that we are returning a `context`, Please update the comment on line 261 about the new responsibility for the interface and what implementers should take care of while returning a `context`

runtime/context.go
```diff
@@ -281,54 +296,74 @@ type contextLogger struct {
...
 		c.log.Info(msg, accumulateLogFields(ctx, userFields)...)
+	} else {
+		ctx = accumulateLogMsgAndFieldsInContext(ctx, msg, userFields)
 	}
+	return ctx
 }
 
-func (c *contextLogger) PanicZ(ctx context.Context, msg string, userFields ...zap.Field) {
+func (c *contextLogger) PanicZ(ctx context.Context, msg string, userFields ...zap.Field) context.Context {
 	if !c.skipZanzibarLogs {
```
isopropylcyanide, line 352: Please invert the if condition to make it easier to read `!skip = do not skip = slightly hard to read`

```
if c.skipZanzibarLogs {
	ctx = accumulateLogMsgAndFieldsInContext(ctx, msg, userFields)	
} else {
	c.log.Panic(msg, accumulateLogFields(ctx, userFields)...)	
}
```

runtime/context.go
```diff
@@ -207,6 +209,19 @@ func accumulateLogFields(ctx context.Context, newFields []zap.Field) []zap.Field
 	return append(previousFields, newFields...)
 }
 
+func accumulateLogMsgAndFieldsInContext(ctx context.Context, msg string, newFields []zap.Field) context.Context {
+	ctxLogCounter := 1
+	v := ctx.Value(ctxLogCounterName)
+	if v != nil {
+		ctxLogCounter = v.(int)
+		ctxLogCounter++
```
isopropylcyanide, line 217: Please **confirm** if this is in the scope of a single request and atomic update is not required.

examples/example-gateway/config/test.json
```diff
@@ -45,6 +45,7 @@
 	"clients.withexceptions.timeout": 10000,
 	"clients.withexceptions.maxConcurrentRequests": 1000,
 	"clients.withexceptions.errorPercentThreshold": 20,
+	"contextlogger.skipzanzibarlogs": false,
```
isopropylcyanide, line 48: Nit: Do we need to mention `zanzibarLogs`  again?

Suggestion `contextlogger.zanzibar.skip: true` is more readable and easily extensible.

 For example, I can say the following for future use cases
```
contextlogger.Y.skip : true
contextlogger.X.skip : false
contextlogger.{{""}}.skip : false

``` 

It also makes it easier to search for `contextLogger.(.*).skip` in the regex.

runtime/context.go
amit-lamba: separate requests maintain separate contexts, so this should work fine.

examples/example-gateway/config/test.json
amit-lamba: contextlogger.zanzibar.skip does not make full sense, made it contextlogger.zanzibar.skiplogs

examples/example-gateway/config/test.json
```diff
@@ -45,6 +45,7 @@
 	"clients.withexceptions.timeout": 10000,
 	"clients.withexceptions.maxConcurrentRequests": 1000,
 	"clients.withexceptions.errorPercentThreshold": 20,
+	"contextlogger.zanzibar.skiplogs": false,
```
ashish-agg, line 48: I think you should add this config in test.yaml and production.yaml with default value in zanzibar.

runtime/context.go
```diff
@@ -244,19 +259,20 @@ func (c *ContextExtractors) ExtractLogFields(ctx context.Context) []zap.Field {
...
-	PanicZ(ctx context.Context, msg string, fields ...zap.Field)
-	WarnZ(ctx context.Context, msg string, fields ...zap.Field)
+	Debug(ctx context.Context, msg string, fields ...zap.Field) context.Context
+	Error(ctx context.Context, msg string, fields ...zap.Field) context.Context
+	Info(ctx context.Context, msg string, fields ...zap.Field) context.Context
+	Panic(ctx context.Context, msg string, fields ...zap.Field) context.Context
+	Warn(ctx context.Context, msg string, fields ...zap.Field) context.Context
+
+	// DebugZ skips logs, and adds to context if skipZanzibarLogs is set to true otherwise behaves as normal Debug, similarly for other XxxxZ's
+	DebugZ(ctx context.Context, msg string, fields ...zap.Field) context.Context
```
ashish-agg, line 271: What is the purpose of returning the context ? If we pass the context pointer then returning context might not be required.

isopropylcyanide: `context.Context` interface has fields that point to pointers under the hood only. 
We need not pass a pointer to a context.

amit-lamba: The standard way of using the context is to update the context and return the context. Do we see any reason why we should pass a pointer and not return a context?

runtime/context.go
isopropylcyanide: Bump

examples/example-gateway/config/test.json
amit-lamba: done

runtime/context.go
amit-lamba: done

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -81,12 +81,12 @@ func (h *{{$handlerName}}) Handle(
 	ctx context.Context,
 	reqHeaders map[string]string,
 	wireValue *wire.Value,
-) (isSuccessful bool, response zanzibar.RWTStruct, headers map[string]string, e error) {
+) (ctxRes context.Context, isSuccessful bool, response zanzibar.RWTStruct, headers map[string]string, e error) {
```
isopropylcyanide, line 84: First param should always be named `ctx` for consistency. You can name the local member as `ctxRes`

examples/example-gateway/clients/contacts/fixture/fixture.go
```diff
@@ -15,7 +15,7 @@ var saveContactsFixtures = &mc.SaveContactsScenarios{
 			},
 		},
 
-		Ret0: &gen.SaveContactsResponse{},
+		Ret1: &gen.SaveContactsResponse{},
```
isopropylcyanide, line 18: Curious why should the `ret0` argument be changed. We're not adding any new args. So there shouldn't be any changes here.

runtime/tchannel_interfaces.go
```diff
@@ -43,5 +43,5 @@ type TChannelCaller interface {
 type TChannelHandler interface {
 	// Handle should read the request from the given reqReader, and return the response struct.
 	// The arguments returned are (whether there was an application error, response headers, result struct, unexpected error)
```
isopropylcyanide, line 45: The interface description might need an amendment here.

isopropylcyanide (approved):
Looks good. Might need clarification on few points (not land blocking)

Also, kindly add a detailed PR description, update `changelog.md` for breaking changes and remember to cut a release

examples/example-gateway/clients/contacts/fixture/fixture.go
amit-lamba: This is a fixture to test that the second parameter Ret1 is the correct response, because Ret0 is ctx after the change, and we can not equate that.

runtime/tchannel_interfaces.go
amit-lamba: Amended

codegen/templates/tchannel_endpoint.tmpl
amit-lamba: ctx is function argument, so the same name can not be used in response arguments

PR #787, opened 2021-09-17
Expose API Environment in Zanzibar metrics

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Feature | Medium |

## Description
- Introduce `apienvironment` tag to distinguish metrics arising from multiple `X-Api-Environment`. Default value is `production` and the default header is `X-Api-Environment` (case insensitive)

## Motivation 
- Internal gateways siphon Zanzibar metrics where they are rolled up and aggregated. Because `apienvironment` is a feature ask from users, we cannot simply implement it in internal gateways alone and need to expose this in Zanzibar.

Internal gateways can then override the custom header according to their `apiEnvironmentHeader` property

## Cardinality concerns
- This tag in all likelihood is going to have a finite set of values (< 5) and hence no metric explosion is expected from introduction of this tag.
- No new counters/histograms are being added. Only the scopes are updated.

---

#### Tested that all requests now carry the new tag in the scope
Expression

#### Tested that all responses also carry the new tag
Expression

#### Integration test on example gateway shows that the tag is going to be emitted for `endpoint
Pasted Graphic 2


runtime/server_http_request.go
```diff
@@ -101,6 +101,16 @@ func NewServerHTTPRequest(
...
+	apiEnvironment := apiEnvironmentDefault
+	if endpoint.config != nil &&
+		endpoint.config.ContainsKey("apiEnvironmentHeader") &&
+		r.Header.Get(endpoint.config.MustGetString("apiEnvironmentHeader")) != "" {
+		apiEnvironment = r.Header.Get(endpoint.config.MustGetString("apiEnvironmentHeader"))
+	}
+	scopeTags[scopeTagsAPIEnvironment] = apiEnvironment
+	logFields = append(logFields, zap.String(apienvironmentKey, apiEnvironment))
+
 	// Overriding the environment for shadow requests
```
isopropylcyanide, line 114: The block above is the crux. We "always" emit this tag and default it to prod unless specified in request header. The header value can also be customized at internal gateways.

runtime/server_http_request_test.go
```diff
@@ -41,24 +41,11 @@ import (
 )
 
 func TestInvalidReadAndUnmarshalBody(t *testing.T) {
-	gateway, err := benchGateway.CreateGateway(
```
isopropylcyanide, line 44: Just refactored this test. Nothing changes

amit-lamba:
LGTM

amit-lamba (approved):

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "dependencies": {
     "eclint": "^1.1.5",
+    "istanbul": "^0.4.5",
```
isopropylcyanide, line 4: This was added by running `make cover` on local. We should check this missing dependency in

https://github.com/gotwarlost/istanbul/releases/tag/v0.4.5

PR #792, opened 2021-10-11
Make populating QPS levels in the config optional

bennytran-uber:
Because not all zanzibar applications require QPS levels, we are making this
optional by specifying `qpsLevelsEnabled` in the main app config file. Additionally,
the following changes were made

* Add a new method for fetching or providing a default value for booleans when reading a config file
* Add QPS level checks  to existing code gen tests
* Add a new test with QPS levels disabled
* Refactored some common assertion logic from tests into a single method


runtime/static_config.go
```diff
@@ -115,11 +115,33 @@ func NewStaticConfigOrDie(
...
-func (conf *StaticConfig) MustGetBoolean(key string) bool {
+func (conf *StaticConfig) checkConfDestroyed(key string) {
 	if conf.destroyed {
-		panic(errors.Errorf("Cannot get(%s) because destroyed", key))
+		panic(errors.Errorf("Cannot get (%s) because destroyed", key))
 	}
+}
+
+// GetOrDefaultBoolean returns the boolean value for a key or returns a default value
+func (conf *StaticConfig) GetOrDefaultBoolean(key string, defaultBool bool) bool {
```
rpatali, line 125: why do you need this method? All the exiting use cases rely on `ContainsKey` method.

codegen/test-service/endpoints/health/embedded-client/client-config.yaml
```diff
@@ -3,3 +3,6 @@ config:
     x-service-name: example
 name: embedded
 type: http
+clientId: "embeddedClientTest"
+clientMethod: "TestMethod"
+qpsLevel: 1
```
rpatali, line 8: qpsLevel is for the endpoint config file right?

runtime/static_config.go
bennytran-uber: I initially created it as a convenience method, but that's a good point. I removed this and refactored the logic to rely on `ContainsKey` instead for consistency.

bennytran-uber:
Latest change includes changes to address feedback and rebased.

rpatali (approved):
LGTM

PR #805, opened 2022-01-10
Expose overhead metric

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Observability | Medium |

## Description
- Introduce  two new metrics `endpoint.overhead.latency` and  `endpoint.overhead.latency-hist` metrics for each TChannel/HTTP endpoint. These correspond to the time spent solely doing gateway specific overhead.

## Motivation 
Currently, Zanzibar Gateways expose no metric to figure out the exact overhead being added by

- Gateway middlewares (including default) and request / response path middlewares
- runtime serde overhead
- other baggage.

This is essential and should have been a key indicator in all endpoint dashboards right from the start. Every team should have key visibility into knowing the exact overhead that Zanzibar gateways add so that they have a better understanding of their resource utilization and the cost of the middlewares they add.

Additionally, when we perform runtime optimizations at Zanzibar Gateways, it is invaluable to know whether the Gateway overhead has increased or decreased for high RPS / business critical endpoints.

#### Tested that for both TChannel and HTTP flows, the `delta` metric is being populated. 
More details on the internal diff. 
image

### DIFF
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -368,6 +368,10 @@ func (c *{{$clientName}}) {{$methodName}}(
 		respHeaders[k] = res.Header.Get(k)
 	}
 
+	defer func() {
+		respHeaders[zanzibar.ClientResponseDurationKey] = res.Duration.String()
+	}()
+
 	{{- if .ResHeaders }}
 	// TODO(jakev): verify mandatory response headers
 	{{- end}}
@@ -471,9 +475,6 @@ func (c *{{$clientName}}) {{$methodName}}(
 			}
 	}
 	{{else}}
-	defer func() {
-		respHeaders[zanzibar.ClientResponseDurationKey] = res.Duration.String()
-	}()
 
 	switch res.StatusCode {
 		case {{.OKStatusCode.Code}}:
--- a/codegen/templates/http_client.tmpl
+++ b/codegen/templates/http_client.tmpl
@@ -471,6 +471,10 @@ func (c *{{$clientName}}) {{$methodName}}(
 			}
 	}
 	{{else}}
+	defer func() {
+		respHeaders[zanzibar.ClientResponseDurationKey] = res.Duration.String()
+	}()
+
 	switch res.StatusCode {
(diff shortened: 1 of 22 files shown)


codegen/templates/http_client.tmpl
```diff
@@ -471,6 +471,10 @@ func (c *{{$clientName}}) {{$methodName}}(
 			}
 	}
 	{{else}}
+	defer func() {
+		respHeaders[zanzibar.ClientResponseDurationKey] = res.Duration.String()
```
isopropylcyanide, line 475: Regardless of HTTP status code response, the duration is set.

examples/example-gateway/endpoints/tchannel/baz/baz_call_test.go
```diff
@@ -45,6 +45,14 @@ func TestBazCall(t *testing.T) {
 	success, resHeaders, err := ms.MakeTChannelRequest(
 		ctx, "SimpleService", "Call", reqHeaders, args, &result,
 	)
+	dynamicRespHeaders := []string{
+		"client.response.duration",
```
isopropylcyanide, line 49: Since this is a dynamically generated header key/val, we exclude this from unit tests.

runtime/constants.go
```diff
@@ -23,12 +23,14 @@ package zanzibar
...
-	endpointLatency      = "endpoint.latency"
-	endpointLatencyHist  = "endpoint.latency-hist"
+	endpointRequest             = "endpoint.request"
+	endpointSuccess             = "endpoint.success"
+	endpointStatus              = "endpoint.status"
+	endpointSystemErrors        = "endpoint.system-errors"
+	endpointLatency             = "endpoint.latency"
+	endpointLatencyHist         = "endpoint.latency-hist"
+	endpointOverheadLatency     = "endpoint.overhead.latency"
+	endpointOverheadLatencyHist = "endpoint.overhead.latency-hist"
```
isopropylcyanide, line 33: These two are the new metrics

uday2602 (approved):

uday2602 (approved):

PR #807, opened 2022-01-13
Add Client ID in scope tag against latency metrics

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Observability | Medium |

## Description
- Introduce client type tag on server endpoint response status/latency metrics.

## Note to reviewers
There are 2 commits in total.

- To save time, only review [this commit](https://github.com/uber/zanzibar/pull/807/commits/853b6c2812d3dd4a54ceaef9f71d0ceafd14690a) which is the real change (Only 7 files changed)
- Rest 51 files are all generated and are added as part of the [2nd commit](https://github.com/uber/zanzibar/pull/807/commits/7a111efaa68c058cae2f36c6c07c0b1363f49941)

## Motivation 
After our work on #805 and #806, we'd like to be able to setup observability on metrics by client type. This will help us identify the overhead per client type. TChannel clients for example need different optimizations as compared to HTTP

As you can see, all we have when it comes to protocol information is `http` since all Gateway endpoints are mostly HTTP traffic.

image

## How was this tested
- Unit tests
- Ran example gateway on local and issued requests against endpoints with `clientless, tchannel and http` clients. Could see the `clienttype` tag being marked.
productionserviceexample-gateway

### DIFF
--- a/test/endpoints/bar/bar_metrics_test.go
+++ b/test/endpoints/bar/bar_metrics_test.go
@@ -62,7 +62,7 @@ func TestCallMetrics(t *testing.T) {
 		},
 	)
 
-	numMetrics := 13
+	numMetrics := 16
 	cg := gateway.(*testGateway.ChildProcessGateway)
 	cg.MetricsWaitGroup.Add(numMetrics)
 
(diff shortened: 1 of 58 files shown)


test/endpoints/bar/bar_metrics_test.go
```diff
@@ -62,7 +62,7 @@ func TestCallMetrics(t *testing.T) {
 		},
 	)
 
-	numMetrics := 13
+	numMetrics := 16
```
isopropylcyanide, line 65: New metrics added are

- endpoint.overhead.latency
- endpoint.overhead.latency-hist
- endpoint.overhead.latency.ratio

uday2602 (approved):

PR #813, opened 2022-02-25
Don't leak internal meta header in the server response

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Bug | Medium |

## Description
- Fix the template that prevents leaking internal client specific headers in case request methods have no body.

## Notes to reviewers
Out of 92, only one file needs to be reviewed `codegen/templates/endpoint.tmpl`. Skim through the rest of the files. All are generated by Zanzibar.

## Motivation 
There's a small bug. In Edge Gateway, we are leaking the internal client specific headers such as `client.type` and `client.response.duration` to the server response when the endpoint takes in no struct. It's an edge case that we missed.

The bug is that the logic for resetting these header values when sending the server response is broken for this edge case. The code generation template executes all block from {{ if }} to {{ end }} and this block was basically a no-op

While we absolutely need the client type in scope tags [#807](https://github.com/uber/zanzibar/pull/807) and client response for calculating latency overhead [#805](https://github.com/uber/zanzibar/pull/805), we shouldn't be leaking this.

## Test Plan
I tested this by reproducing this on a sample endpoint and generating code before/after. 
As you can clearly see, we are not leaking these internal header to the server response anymore.

image


codegen/templates/endpoint.tmpl
```diff
@@ -178,20 +178,6 @@ func (h *{{$handlerName}}) HandleRequest(
...
-		}
-		if val, ok := cliRespHeaders.Get(zanzibar.ClientTypeKey); ok {
-			res.ClientType = val
-			cliRespHeaders.Unset(zanzibar.ClientTypeKey)
-		}
-	}
-
 	// log downstream response to endpoint
 	if ce := h.Dependencies.Default.ContextLogger.Check(zapcore.DebugLevel, "stub"); ce != nil {
 		zfields := []zapcore.Field{
```
isopropylcyanide, line 183: Just moving the whole block to after the `"{{end - }}"` block

aloksinghaluber (approved):

PR #815, opened 2022-03-02
Optimized template generated code for maxAttempts(retry count)

YuktaChauhan1006:
With this pull request we are trying to optimize the code that get generated for every tchannel client. 
Since by default the maxAttempts should be 5 for all tchannel clients if not specifically mentioned, so through this change we will be passing 5 if it is not set.


codegen/templates/tchannel_client.tmpl
```diff
@@ -166,51 +166,31 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
-					Timeout:              timeout,
-					TimeoutPerAttempt:    timeoutPerAttempt,
-					RoutingKey:           &routingKey,
-					RuleEngine:           re,
-					HeaderPatterns:       headerPatterns,
-					RequestUUIDHeaderKey: requestUUIDHeaderKey,
-					AltChannelMap:        altChannelMap,
-				},
-			)
+		maxAttempts = 5
```
isopropylcyanide, line 173: Declare a const int with the right name. Magic numbers will not pass lint

amit-lamba: +1

isopropylcyanide (approved):
LGTM

ashish-agg (approved):

PR #811, opened 2022-02-16
Use streaming "no-wire" thriftrw-go serde

isopropylcyanide:
## Summary
| PR Status  | Type  | Impact level |
| :---: | :---: | :---: |
| Ready  | Enhancement | High |

## Description
- Use streaming thrift io for thriftrw serialization and deserialization that was supported in [thriftrw-go/v1.29.0](https://github.com/thriftrw/thriftrw-go/releases/tag/v1.29.0)

image

## Notes to reviewers
- None of the functionality is changed. Only the way serialization / deserialization happens.
- Out of the 76 files changed, **only 10 files** need to be reviewed. Rest are all generated. 
- To make it easier to review just look at the non generated [commit](https://github.com/uber/zanzibar/pull/811/commits/25710e6dc595408c232e6788c0c6edae1236e7be)
- Feel free to look at some of the generated code to see what has changed in [this commit](https://github.com/uber/zanzibar/pull/811/commits/1037a01be9a4137bead29637915d9964cea28dd1)

## Motivation
- Edge Gateway heavily relies on [thriftrw-go](https://github.com/thriftrw/thriftrw-go) library for its serde on thrift structs.
- With  [thriftrw-go/v1.29.0](https://github.com/thriftrw/thriftrw-go/releases/tag/v1.29.0) there is now a way to support serialization and deserialization of thrift structs directly without converting them into an intermediate wire representation. 

- Expected impact

| Component | Impact | Remarks | 
| --- | --- | --- |
| CPU Usage | Decrease | Serde is now faster without the wire representation |
| Heap in Use | Decrease | No buffers for the intermediate representation are required |
| Endpoint Latency | Decrease | As serialization is a key component of TChannel client call |
| Latency Overhead | Decrease | As serialization is a key component of the data plane latency overhead |
| Throughput / Core | Increase | More CPU is available to serve requests |
| CPU / request | Decrease | Lesser CPU for serde is required per request |
| Memory / request | Decrease | Lesser heap for serde is required per request |
---

## How was this tested?
- Unit tests and all Zanzibar integration tests
- Tested using local YAB on a zanzibar endpoint using the new serialization/deserialization format. The endpoint is a TChannel endpoint that talks to a TChannel client. This way we can test both the decoding and encoding flows

Pasted Graphic 6

```
[...]

PR #818, opened 2022-03-24
Add missing import for thriftrw stream package

isopropylcyanide:
## Summary
| :---: | :---: | :---: |
| Ready  | Chore | Low |

## Description
- Add a dedicated import for `"go.uber.org/thriftrw/protocol/stream"``
- Remove unused import `"go.uber.org/thriftrw/wire"`
- Only one line is changed in `tchannel_endpoint.tmpl`. Rest is auto generated.

## Motivation 
- When we enabled streaming interfaces in Zanzibar, #811 we missed adding an import in the template. Usually this isn't a problem because we run `goimports` on the generated code. However, some other repositories aren't smart enough and fail to resolve the `stream.Reader` imports

### DIFF
--- a/codegen/templates/tchannel_endpoint.tmpl
+++ b/codegen/templates/tchannel_endpoint.tmpl
@@ -10,8 +10,9 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"go.uber.org/thriftrw/wire"
+	"go.uber.org/thriftrw/protocol/stream"
 	"go.uber.org/zap"
+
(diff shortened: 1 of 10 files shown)


codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -10,8 +10,9 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"go.uber.org/thriftrw/wire"
```
isopropylcyanide, line 13: `wire` in unused. We replace with `stream` instead.

PR #819, opened 2022-05-25
Tchannel server IP configurability

sean1006:
Add configurability for tchannel server IP when set in the configurat…ion file

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -305,7 +305,9 @@ func (gateway *Gateway) Bootstrap() error {
 
 	// start TChannel server
 	ip := localhost
-	if env != testenv {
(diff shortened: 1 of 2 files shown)


runtime/gateway.go
```diff
@@ -305,7 +305,9 @@ func (gateway *Gateway) Bootstrap() error {
 
 	// start TChannel server
 	ip := localhost
-	if env != testenv {
+	if env != testenv && gateway.Config.ContainsKey("tchannel.server.ip") {
```
bennytran-uber, line 308: Let's remove the `env != testenv` condition here. It makes more sense for the configuration to use the specific ip address that we provide even when it's in a test environment.

PR #823, opened 2022-08-04
Bishnu/ep level retry timeout http clients

bishnuag:
**ERD:**
https://docs.google.com/document/d/1e6Nmhoa5ygsh2lXSzUyJdHqz2pXF0lc5ACr8DzwKBio/edit#

**Summary:**
- Let the user pass TimeoutAndRetry parameters per request to the client (HTTP and channel) 
- If TimeoutAndRetry parameters per request are not passed, client-level TimeoutAndRetry parameters is taken into consideration
- Runtime considers the least timeout value among clients and request level

**fyi**: The branch reads it as the changes are only for HTTP clients but its for both HTTP and TChannel


runtime/client_http_request.go
```diff
@@ -204,11 +212,24 @@ func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
...
 	}
 
-	res, err := req.client.Client.Do(req.httpReq.WithContext(ctx))
+	var res *http.Response
+
+	// MaxAttempts is 0 when (this is just to reduce the overall risk, eventually will be enabled for all eps)
+	// a. ep is tier < 3 and
+	// b. ep is not configured explicitly to use timeout and max retry at ep level
+
+	if req.MaxAttempts == 0 {
```
bishnuag, line 221: Should we make it 1 instead 0?

codegen/templates/http_client.tmpl
```diff
@@ -123,19 +124,32 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
+
+	//get mapping of client method and it's timeout
+	//if mapping is not provided, use client's timeout for all methods
+	clientMethodTimeoutMapping := make(map[string]int64)
+	if deps.Default.Config.ContainsKey("clients.{{$clientID}}.methodTimeoutMapping") {
+		deps.Default.Config.MustGetStruct("clients.{{$clientID}}.methodTimeoutMapping", &clientMethodTimeoutMapping)
+	} else {
+		for _, serviceMethodName := range methodNames {
+			methodName := strings.Split(serviceMethodName, "::")[1]
+			clientMethodTimeoutMapping[methodName] = int64(timeoutVal)
```
anirudhraja, line 136: clientMethodTimeoutMapping lets check if it has method name ?

codegen/templates/workflow.tmpl
```diff
@@ -168,40 +170,49 @@ func (w {{$workflowStruct}}) Handle(
 		}
 	}
 
+	//maxAttempts value 0 will execute all previous code and doesn't changes anything
+	//maxAttempts value 0 will not use requestTimeoutPerAttemptInMs information
+	timeoutAndRetryOps := zanzibar.TimeoutAndRetryOptions{
+		OverallTimeoutInMs: 			time.Duration(3000) * time.Millisecond,
```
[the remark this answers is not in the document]
bishnuag: Made it configurable

runtime/client_http_request.go
```diff
@@ -203,21 +212,76 @@ func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
...
+			}
+		}
+
+		req.ContextLogger.GetLogger().Info("errors while making http outbound request",
+			zap.Error(err),
+			zap.String("clientId", req.ClientID), zap.String("methodName", req.MethodName),
+			zap.String("attempt", string(rune(retryCount))),
+			zap.String("maxAttempts", string(rune(req.timeoutAndRetryOptions.MaxAttempts))))
+
+		//TODO (future releases) - make retry conditional, inspect error/response and then retry
```
anirudhraja, line 268: Cany maybe model as retry strategy and use strategy pattern to have various types of strategies. Will be extensible

scripts/cover.sh
```diff
@@ -128,7 +128,7 @@ cat ./coverage/istanbul.json | jq '[
 ] | from_entries' > ./coverage/istanbul-runtime.json
 
 echo "Checking code coverage for runtime folder"
-./node_modules/.bin/istanbul check-coverage --statements 99.85 \
+./node_modules/.bin/istanbul check-coverage --statements 99.79 \
```
anirudhraja, line 131: Why ? :)

codegen/templates/http_client.tmpl
```diff
@@ -123,19 +124,32 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 	if deps.Default.Config.ContainsKey("clients.{{$clientID}}.circuitBreakerDisabled") {
 		circuitBreakerDisabled = deps.Default.Config.MustGetBoolean("clients.{{$clientID}}.circuitBreakerDisabled")
 	}
+
+	//get mapping of client method and it's timeout
+	//if mapping is not provided, use client's timeout for all methods
```
isopropylcyanide, line 129: Might be too late but `method` **isn't the right nomenclature**. We should be pushing for `procedure` instead to stay in parity with a general router layer whose sole purpose is to dispatch procedures calls to remote outbounds (similar to Muttley)

codegen/templates/http_client.tmpl
isopropylcyanide: Wrap the logic in below idiomatic go pattern to guard against existing timeouts. 
```
if val, ok := clientMethodTimeoutMapping[methodName]; ok {
  // do something with val
}
```

codegen/templates/http_client.tmpl
```diff
@@ -123,19 +124,32 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
+	clientMethodTimeoutMapping := make(map[string]int64)
+	if deps.Default.Config.ContainsKey("clients.{{$clientID}}.methodTimeoutMapping") {
+		deps.Default.Config.MustGetStruct("clients.{{$clientID}}.methodTimeoutMapping", &clientMethodTimeoutMapping)
+	} else {
+		for _, serviceMethodName := range methodNames {
+			methodName := strings.Split(serviceMethodName, "::")[1]
```
isopropylcyanide, line 135: IMO `strings.Split` shouldn't be in the hot path. It performs not so efficiently for a simple use case and this code needs to be executed against every RPC. We should rather prefer to write our own if you are already aware of the format.

Here's where you can see the allocation.
```
a := make([]string, n)
```

image

Additionally: If you'd rather use `strings.Split` ensure you pass `strings.Split(string, separator, n = 1) so that we discard the rest of the string as soon as we find the split.

codegen/templates/http_client.tmpl
```diff
@@ -123,19 +124,32 @@ func {{$exportName}}(deps *module.Dependencies) Client {
...
+	}
+
 	qpsLevels := map[string]string{
 				{{range $methodName, $qpsLevel := $QPSLevels -}}
 				"{{$methodName}}": "{{$qpsLevel}}",
 				{{end}}
 	}
 	if !circuitBreakerDisabled {
-		for methodName := range methodNames {
+		for methodName, methodTimeoutVal := range clientMethodTimeoutMapping{
```
isopropylcyanide, line 146: Nit: Prefer `methodTimeout` for brevity.
```suggestion
		for methodName, methodTimeout := range clientMethodTimeoutMapping{
```

codegen/templates/tchannel_endpoint.tmpl
```diff
@@ -266,7 +268,13 @@ func (h *{{$handlerName}}) redirectToDeputy(
 		},
 	)
 
-	success, respHeaders, err := client.Call(ctx, "{{.ThriftService}}", "{{$methodName}}", reqHeaders, req, res)
+	timeoutAndRetryOps := zanzibar.TimeoutAndRetryOptions{
+		OverallTimeoutInMs: 			(timeoutPerAttempt + zanzibar.DefaultBackOffTimeAcrossRetries) * time.Duration(maxAttempts),
```
isopropylcyanide, line 272: Can you extract this to `tchannel_client.go` and simply invoke from the template. We don't need to generate this much new code for all 600+ tchannel endpoint workflows.

codegen/templates/workflow.tmpl
```diff
@@ -168,40 +173,70 @@ func (w {{$workflowStruct}}) Handle(
...
+	var timeoutPerAttempt time.Duration
+	var backOffTimeAcrossRetries time.Duration
+	var overAllTimeout time.Duration
+
+	//when endpoint level timeout information provided, override it with client level config
+	if w.defaultDeps.Config.ContainsKey("endpoints.{{$handleIdDotEndpointId}}.timeoutPerAttempt") {
+		scaleFactor := w.defaultDeps.Config.MustGetFloat("endpoints.{{$handleIdDotEndpointId}}.scaleFactor")
+		maxRetry = int(w.defaultDeps.Config.MustGetInt("endpoints.{{$handleIdDotEndpointId}}.retryCount"))
+
+		backOffTimeAcrossRetriesConf := w.defaultDeps.Config.MustGetInt("endpoints.{{$handleIdDotEndpointId}}.backOffTimeAcrossRetries")
```
isopropylcyanide, line 188:
```suggestion
		backOffTimeAcrossRetriesCfg := w.defaultDeps.Config.MustGetInt("endpoints.{{$handleIdDotEndpointId}}.backOffTimeAcrossRetries")
```

codegen/templates/workflow.tmpl
```diff
@@ -168,40 +173,70 @@ func (w {{$workflowStruct}}) Handle(
...
+		timeoutPerAttemptConf := w.defaultDeps.Config.MustGetInt("endpoints.{{$handleIdDotEndpointId}}.timeoutPerAttempt")
+
+		//add a scale factor to timeoutPerAttempt
+		timeoutPerAttemptConf = int64(math.Ceil(float64(timeoutPerAttemptConf) * scaleFactor))
+
+		//convert timeoutPerAttemptConf & backOffTimeAcrossRetries to Duration
```
isopropylcyanide, line 194: Nit: Redundant comment

codegen/templates/workflow.tmpl
```diff
@@ -168,40 +173,70 @@ func (w {{$workflowStruct}}) Handle(
...
+		timeoutPerAttempt = time.Duration(timeoutPerAttemptConf) * time.Millisecond
+		backOffTimeAcrossRetries = time.Duration(backOffTimeAcrossRetriesConf) * time.Millisecond
+
+		overAllTimeout = time.Duration((timeoutPerAttemptConf + backOffTimeAcrossRetriesConf) * (int64(maxRetry-1)) + timeoutPerAttemptConf) * time.Millisecond
```
isopropylcyanide, line 198: We don't need to have this logic in `tmpl` This should be sourced from the runtime. tmpl should only be the invoker or contain branching at max. This makes our template bloated and increases code gen size.

examples/example-gateway/clients/multi/client-config.yaml
```diff
@@ -1,7 +1,7 @@
 config:
   exposedMethods:
-    HelloA: ServiceABack::hello
-    HelloB: ServiceBBack::hello
+    HelloA: ServiceABack::helloA
+    HelloB: ServiceBBack::helloB
```
isopropylcyanide, line 4: Doesn't it change the test? Two clients map to the same procedure vs two clients mapping to different procedure

runtime/client_http_request.go
isopropylcyanide: +1

Additionally Run this with `goleak` to make sure there are no leaks

https://github.com/uber-go/goleak

runtime/options.go
```diff
@@ -0,0 +1,31 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zanzibar
+
+import "time"
+
+// TimeoutAndRetryOptions specifies timeout and retries params for outbound calls
```
isopropylcyanide, line 25: Nice practice of adding an options here. However, we could explore a option/decorator pattern similar to what is being done in fx and glue: 

https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/eats/domain/promotion/service/eats-promotion-gateway/component_tests/app/app.go#L60:3

apply(decorator)

Ref:

isopropylcyanide (commented):
Initial review

runtime/client_http_request.go
bishnuag: Added retry check that can be extended with a new strategy in future.

Ran test cases with **goleak** no leakage due to this change

anirudhraja (approved):

codegen/templates/http_client.tmpl
isopropylcyanide: I've no problem with that. My only concern is that we should be calling it as `procedure` instead of `method`

isopropylcyanide: So I have the correct understanding. My only point was that we should start calling it `procedures` instead of methods as a platform. E.g `ep level procedure timeouts` instead of `ep level method timeouts`

But this is just a nit.

examples/example-gateway/clients/multi/client-config.yaml
isopropylcyanide: Got it. Can you create a Github issue for this and reference this PR there? Otherwise we'll lose context.

isopropylcyanide (approved):
LGTM assuming that the retry functionality has been extensively tested across sample/example gateways across both (ep level / client level) timeouts.

examples/example-gateway/clients/multi/client-config.yaml
bishnuag: Done https://github.com/uber/zanzibar/issues/841

codegen/templates/tchannel_client.tmpl
```diff
@@ -143,6 +143,17 @@ func {{$exportName}}(deps *module.Dependencies) Client {
 		{{ end -}}
 	}
 
+	//get mapping of client method and it's timeout
+	//if mapping is not provided, use client's timeout for all the methods
+	clientMethodTimeoutMapping := make(map[string]int64)
```
bennytran-uber, line 148: why did we decide to make the values `int64` here? If we used an `int` instead, we could get rid of the type conversion inside `configureCircuitBreaker`

runtime/client_http_request.go
```diff
@@ -72,16 +75,22 @@ func NewClientHTTPRequest(
...
+		client:                 client,
+		ContextLogger:          client.ContextLogger,
+		defaultHeaders:         client.DefaultHeaders,
+		ctx:                    ctx,
+		jsonWrapper:            client.JSONWrapper,
+		timeoutAndRetryOptions: timeoutAndRetryOptions,
 	}
+	req.ContextLogger.Debug(req.ctx, "ClientHTTPRequest definition",
+		zap.String("clientId", req.ClientID), zap.String("methodName", req.MethodName),
+		zap.String("req.RequestTimeoutPerAttemptInMs", string(rune(req.timeoutAndRetryOptions.RequestTimeoutPerAttemptInMs))),
```
bennytran-uber, line 91: why not use `zap.Int` instead to avoid the type conversion?

runtime/client_http_request.go
```diff
@@ -203,21 +212,80 @@ func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
...
+		res, retryCount, err = req.executeDoWithRetry(ctx) //new code for retry and timeout per ep level
+	}
 
-	res, err := req.client.Client.Do(req.httpReq.WithContext(ctx))
 	span.Finish()
+
 	if err != nil {
-		req.ContextLogger.ErrorZ(req.ctx, "Could not make outbound request", zap.Error(err))
-		return nil, err
+		req.ContextLogger.ErrorZ(req.ctx, fmt.Sprintf("Could not make http outbound %s.%s request",
```
bennytran-uber, line 228: is the formatting in the error log message necessary? Shouldn't we get this information through some of the logger tags from the context logger?

runtime/client_http_request.go
```diff
@@ -203,21 +212,80 @@ func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
...
+		var shouldRetry = false
+		// if attempts are pending, wait for backoff duration before next attempt
+		if i+1 < req.timeoutAndRetryOptions.MaxAttempts {
+			shouldRetry = req.client.CheckRetry(ctx, req.timeoutAndRetryOptions, res, err)
+		}
+
+		req.ContextLogger.GetLogger().Info("errors while making http outbound request",
+			zap.Error(err),
+			zap.String("clientId", req.ClientID), zap.String("methodName", req.MethodName),
+			zap.String("attempt", string(rune(retryCount))),
```
bennytran-uber, line 263: same question here, we can use zap.Int types?

runtime/client_http_request.go
```diff
@@ -203,21 +212,80 @@ func (req *ClientHTTPRequest) Do() (*ClientHTTPResponse, error) {
...
+			return res, retryCount, nil
+		}
+
+		var shouldRetry = false
+		// if attempts are pending, wait for backoff duration before next attempt
+		if i+1 < req.timeoutAndRetryOptions.MaxAttempts {
+			shouldRetry = req.client.CheckRetry(ctx, req.timeoutAndRetryOptions, res, err)
+		}
+
+		req.ContextLogger.GetLogger().Info("errors while making http outbound request",
```
bennytran-uber, line 260: should be an error log?

bennytran-uber: also why did we expose the underlying logger instead of using the context logger directly?

runtime/utils.go
```diff
@@ -0,0 +1,43 @@
...
+
+package zanzibar
+
+import (
+	"math"
+	"time"
+)
+
+//GetTimeoutAndRetryConfig encapsulates timeout and retry configs in TimeoutAndRetryOptions struct
+func GetTimeoutAndRetryConfig(timeoutPerAttemptConf int, backOffTimeAcrossRetriesConf int,
```
bennytran-uber, line 29: nit: `BuildTimeoutAndRetryConfig` might be a better name?

codegen/templates/tchannel_client.tmpl
bishnuag: **int** doesn't for `deps.Default.Config.MustGetStruct("clients.{{$clientID}}.methodTimeoutMapping", &clientMethodTimeoutMapping)` statement, hence had to update it int64

runtime/client_http_request.go
bishnuag: Make sense, Addressed

runtime/client_http_request.go
bishnuag: Ideally, it should have come through the context logger but it wasn't the case. As a quick fix I added it here instead debugging it

runtime/client_http_request.go
bishnuag: Addressed

runtime/client_http_request.go
bishnuag: Have updated it as **Warn** as retries could be pending which says it's not the final error. 
Why not contextLogger? Each attempt may fail with diff reason and may override the message in contextLogger, what we want is individual logs for each retry. Other approach is to fix contextLogger to have same field with multiple values, like msg1, msg2..

runtime/utils.go
bishnuag: :), addressed

runtime/client_http_request.go
bishnuag: RequestTimeoutPerAttemptInMs is in time.Duration format - hence I kept it as string

bennytran-uber (approved):

PR #830, opened 2022-08-16
add additional logs to debug during panic

ThatipallyKoushik:
logged request during panic which may help to debug https://t3.uberinternal.com/browse/LOCATION-4808 .

### DIFF
--- a/runtime/router.go
+++ b/runtime/router.go
@@ -220,6 +220,8 @@ func (router *httpRouter) handlePanic(
 		"A http request handler paniced",
 		zap.Error(err),
 		zap.String("pathname", r.URL.RequestURI()),
+		zap.String("host", r.Host),
+		zap.String("remoteAddr", r.RemoteAddr),
 	)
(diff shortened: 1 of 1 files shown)


runtime/router/router_test.go
```diff
@@ -227,3 +227,13 @@ func TestNotFoundCustom(t *testing.T) {
 	assert.True(t, handled)
 	assert.Equal(t, http.StatusNotFound, res.Result().StatusCode)
 }
+
+func Test_urlFailure_Error(t *testing.T) {
```
bishnuag, line 231: The naming convention is not right. Kindly sync format with other methods

bishnuag (changes requested):

bishnuag:
Squash all commits into one @ThatipallyKoushik

runtime/router/router_test.go
ThatipallyKoushik: done

PR #836, opened 2022-09-23
add latency and availability metrics to gateway

zeeshan-mohd:
Added metrics that needs to be sent for latency ( in case of middleware request and response ) and availability metrics for middlewares requests


zeeshan-mohd:
Closing since ci build are not running 
created another pr

PR #838, opened 2022-10-03
add latency and availability metrics to gateway

zeeshan-mohd:
Added metrics that needs to be sent for latency ( in case of middleware request and response ) and availability metrics for middlewares requests

### DIFF
--- a/runtime/middlewares.go
+++ b/runtime/middlewares.go
@@ -22,8 +22,16 @@ package zanzibar
 
 import (
 	"context"
+	"time"
 
 	jsonschema "github.com/mcuadros/go-jsonschema-generator"
+	"github.com/uber-go/tally"
+)
+
+const (
+	middlewareRequestLatencyTag  = "middleware.request.latency"
+	middlewareResponseLatencyTag = "middleware.response.latency"
+	middlewareRequestStatusTag   = "middleware.request.status"
 )
 
 // MiddlewareStack is a stack of Middleware Handlers that can be invoked as an Handle.
@@ -103,6 +111,7 @@ func (m *MiddlewareStack) Handle(
 	res *ServerHTTPResponse) context.Context {
 
 	shared := NewSharedState(m.middlewares)
+	middlewareRequestStartTime := time.Now()
 
 	for i := 0; i < len(m.middlewares); i++ {
 		ctx, ok := m.middlewares[i].HandleRequest(ctx, req, res, shared)
@@ -110,17 +119,48 @@ func (m *MiddlewareStack) Handle(
 		// then abort the rest of the stack and evaluate the response
 		// handlers for the middlewares seen so far.
 		if ok == false {
+			//record latency for middlewares requests in unsuccesful case as the middleware requests calls are terminated
+			m.recordLatency(middlewareRequestLatencyTag, middlewareRequestStartTime, req.scope)
+
+			middlewareResponseStartTime := time.Now() // start the timer for middleware responses
 			for j := i; j >= 0; j-- {
 				m.middlewares[j].HandleResponse(ctx, res, shared)
+			}
+			//record latency for middlewares responses in unsuccesful case
+			m.recordLatency(middlewareResponseLatencyTag, middlewareResponseStartTime, req.scope)
+
+			//for error metrics only emit when there is gateway error and not request error
+			// the percentage can be calculated via error_count/total_request
+			if res.pendingStatusCode >= 500 {
+				m.emitAvailabilityError(middlewareRequestStatusTag, req.scope)
 			}
 			return ctx
 		}
 	}
+	// record latency for middlewares requests in successful case
+	m.recordLatency(middlewareRequestLatencyTag, middlewareRequestStartTime, req.scope)
 
 	ctx = m.handle(ctx, req, res)
 
+	middlewareResponseStartTime := time.Now()
 	for i := len(m.middlewares) - 1; i >= 0; i-- {
 		m.middlewares[i].HandleResponse(ctx, res, shared)
 	}
+	// record latency for middlewares responses in successful case
+	m.recordLatency(middlewareResponseLatencyTag, middlewareResponseStartTime, req.scope)
 	return ctx
+}
+
+// recordLatency measures the latency as per the tagName and start time given.
+func (m *MiddlewareStack) recordLatency(tagName string, startTime time.Time, scope tally.Scope) {
+	elapsed := time.Now().Sub(startTime)
+	scope.Timer(tagName).Record(elapsed)
+}
+
+// emitAvailability is used to increment the error counter for a particular tagName.
+func (m *MiddlewareStack) emitAvailabilityError(tagName string, scope tally.Scope) {
+	tagged := scope.Tagged(map[string]string{
+		scopeTagStatus: "error",
+	})
+	tagged.Counter(tagName).Inc(1)
 }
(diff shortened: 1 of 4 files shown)


runtime/middlewares.go
```diff
@@ -103,24 +110,59 @@ func (m *MiddlewareStack) Handle(
...
+
+			start = time.Now() // start the timer for middleware responses
 			for j := i; j >= 0; j-- {
 				m.middlewares[j].HandleResponse(ctx, res, shared)
 			}
+			//record latency for middlewares responses in unsuccesful case
+			m.recordLatency(middlewareResponseLatencyTag, start, req)
+
+			//for error metrics only emit when there is gateway error and not request error
+			if res.pendingStatusCode >= 500 {
```
isopropylcyanide, line 132: 4xx is a success?

runtime/middlewares.go
```diff
@@ -22,10 +22,17 @@ package zanzibar
 
 import (
 	"context"
+	"time"
 
 	jsonschema "github.com/mcuadros/go-jsonschema-generator"
 )
 
+const (
+	middlewareRequestLatencyTag  = "middleware.requests.latency"
```
isopropylcyanide, line 31: Should be singular `middleware.request.latency`

runtime/middlewares.go
```diff
@@ -103,24 +110,59 @@ func (m *MiddlewareStack) Handle(
...
 		}
 	}
+	// record latency for middlewares requests in successful case
+	m.recordLatency(middlewareRequestLatencyTag, start, req)
+	//emit success metric for middleware requests
+	m.emitAvailability(middlewareRequestStatusTag, "success", req)
 
 	ctx = m.handle(ctx, req, res)
 
+	start = time.Now()
```
bishnuag, line 147: multiple usage of start var, lets rename to middlewareResponseStartTIme

runtime/middlewares.go
```diff
@@ -103,24 +110,59 @@ func (m *MiddlewareStack) Handle(
...
 }
+
+// recordLatency measures the latency as per the tagName and start time given.
+func (m *MiddlewareStack) recordLatency(tagName string, start time.Time, req *ServerHTTPRequest) {
+	elapsed := time.Now().Sub(start)
+	req.scope.Timer(tagName).Record(elapsed)
+}
+
+// emitAvailability is used to increment the success/error counter for a particular tagName.
+func (m *MiddlewareStack) emitAvailability(tagName string, status string, req *ServerHTTPRequest) {
```
bishnuag, line 163: pass only req.scope instead entire request param

runtime/middlewares.go
bishnuag: I think it's the wrong place to verify statusCode. statusCode would be available only after we make call to downstream client - **ctx = m.handle(ctx, req, res)**

zeeshan-mohd: > 4xx is a success?

Removed the success part as it can be taken from total_requests.Not taking 4xx as error is because the availability defined for gateway should not be dependent on the user's request and rather on the gateway internal errors, so that is why only 5xx.

bishnuag (approved):
LGTM

runtime/middlewares.go
zeeshan-mohd: > I think it's the wrong place to verify statusCode. statusCode would be available only after we make call to downstream client - **ctx = m.handle(ctx, req, res)**

https://github.com/uber/zanzibar/blob/24bea8e47a8464853e5014da44ac30c2b17f091b/runtime/server_http_response.go#L271
We are setting the pendingStatuscode whenever we call the res.WriteJson from the middlewares,so this will always be populated in case of errors.

runtime/middlewares.go
```diff
@@ -103,24 +111,57 @@ func (m *MiddlewareStack) Handle(
...
 	shared := NewSharedState(m.middlewares)
+	middlewareRequestStartTime := time.Now()
 
 	for i := 0; i < len(m.middlewares); i++ {
 		ctx, ok := m.middlewares[i].HandleRequest(ctx, req, res, shared)
 		// If a middleware errors and writes to the response header
 		// then abort the rest of the stack and evaluate the response
 		// handlers for the middlewares seen so far.
 		if ok == false {
+			//record latency for middlewares requests in unsuccesful case as the middleware requests calls are terminated
```
isopropylcyanide, line 122:
```suggestion
			//record latency for middlewares requests in unsuccessful case as the middleware requests calls are terminated
```

isopropylcyanide:
Please add the test plan and a dump of the new metrics from a sample Zanzibar example gateway for both successful / unsuccessful runs.

runtime/middlewares_test.go
```diff
@@ -183,7 +184,7 @@ func TestMiddlewareRequestAbort(t *testing.T) {
 	if !assert.NoError(t, err) {
 		return
 	}
-	assert.Equal(t, resp.StatusCode, http.StatusOK)
+	assert.Equal(t, resp.StatusCode, http.StatusInternalServerError)
```
isopropylcyanide, line 187: Why is this changed?

isopropylcyanide (changes requested):
Gave some comments

runtime/middlewares.go
```diff
@@ -103,24 +111,57 @@ func (m *MiddlewareStack) Handle(
...
+			m.recordLatency(middlewareRequestLatencyTag, middlewareRequestStartTime, req.scope)
+
+			middlewareResponseStartTime := time.Now() // start the timer for middleware responses
 			for j := i; j >= 0; j-- {
 				m.middlewares[j].HandleResponse(ctx, res, shared)
 			}
+			//record latency for middlewares responses in unsuccesful case
+			m.recordLatency(middlewareResponseLatencyTag, middlewareResponseStartTime, req.scope)
```
isopropylcyanide, line 130: Are we tracking req/resp latency for each middleware or all of the edge middlewares?

> If it's the former, I don't think the code does what it is supposed to do

If there are 4 middlewares and handling fails for the 4th one, then we need to run the decrement loop of handling th responses for the 3 that executed. 
In this case, the middleware response latency recorded for the 3rd would be incorrect as it will include the latency of the first 2 middlewares's responses.

```
for j := i; j >= 0; j-- {
				m.middlewares[j].HandleResponse(ctx, res, shared)
			}
m.recordLatency(middlewareResponseLatencyTag, middlewareResponseStartTime, req.scope)
```

isopropylcyanide (commented):
Requested clarification

runtime/middlewares_test.go
zeeshan-mohd: It was changed because if we are reaching here means that the middleware request has aborted with an error and we will return from here without calling the downstream .So sending a 200 ok in this case is not apt.Additionally , it increased the code coverage as well.

runtime/middlewares.go
sachsingh: updated

runtime/middlewares.go
zeeshan-mohd: > Are we tracking req/resp latency for each middleware or all of the edge middlewares?

Yes , it is for all the middlewares that a request goes through , since these metrics are tagged with endpointID we will apply the filtering . So we are not measuring for one middleware but rather all the middleware that a request goes through and filter on the basis of endpointID's.

sachsingh:
> Please add the test plan and a dump of the new metrics from a sample Zanzibar example gateway for both successful / unsuccessful runs.

Added metrics dump https://code.uberinternal.com/P341907

isopropylcyanide (approved):

PR #849, opened 2023-01-05
added blocking reviewers

sachsingh:
Adding blocking reviewers for a module


bishnuag:
resolve conflicts and squash commits into one.

bishnuag (changes requested):
esolve conflicts and squash commits into one.

PR #855, opened 2023-03-06
Propagated context returned from middleware

hbhoyar-uber:
The current code was calling HandleRequest method on middlewares but the returned context was not passed to next middleware and to the actual handler itself.

The way the code was written was leading to creation of new context instead of setting the value of already defined context.

This is leading to issues like some fields (like start timestamp) added by logging middleware were not being present in emitted log as the context returned by HandleRequest method was never propagated further.


anirudhraja:
LGTM

anirudhraja (approved):

PR #857, opened 2023-03-29
Fix grpc client

groshu:
The GRPC Clients return cancelled context to callers which creates a problem if they read it and use it. Ideally, the scope of the child context created inside the client should be local so making that fix in the template. 
TChannel and HTTP clients handle this correctly.

### DIFF
--- a/codegen/templates/grpc_client.tmpl
+++ b/codegen/templates/grpc_client.tmpl
@@ -200,16 +200,16 @@ func (e *{{$clientName}}) {{$methodName}}(
 			opts = append(opts, yarpc.WithHeader(e.opts.RequestUUIDHeaderKey, reqUUID))
 		}
 	}
-	ctx, cancel := context.WithTimeout(ctx, e.opts.Timeout)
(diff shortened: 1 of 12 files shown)


codegen/templates/grpc_client.tmpl
```diff
@@ -200,16 +200,16 @@ func (e *{{$clientName}}) {{$methodName}}(
 			opts = append(opts, yarpc.WithHeader(e.opts.RequestUUIDHeaderKey, reqUUID))
 		}
 	}
-	ctx, cancel := context.WithTimeout(ctx, e.opts.Timeout)
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, e.opts.Timeout)
```
bishnuag, line 203: can we leave a comment here why ctx was not reused instead a new var ctxWithTimeout was declared? Someone else may make some changes and make a mistake here

groshu: sure

groshu: done

bishnuag (approved):

PR #861, opened 2023-04-06
fix log level logging for logZ

sachsingh:
currently log level logging is not followed for Context logger with LogXXZ enabled methods. fixing it


runtime/context_test.go
chirayu, line 1: Write a test that will exercise the new code added to InfoZ, DebugZ, PanicZ and WarnZ functions.

sachsingh: updated test

chirayu:
LGTM

chirayu (approved):

PR #866, opened 2023-04-11
Passed NotFoundHandler in gateway options

hbhoyar-uber:
- Used custom not found handler if config allows
- Read HandleMethodNotAllowed from config


chirayu (approved):
LGTM

PR #867, opened 2023-04-13
Ading Proxy template support in zanzibar

shivam-srivastava28:


codegen/gateway.go
```diff
@@ -401,6 +403,12 @@ func NewEndpointSpec(
 			thriftInfo, yamlFile,
 		)
 	}
+	var config map[string]interface{}
+	if _, ok := endpointConfigObj["config"]; !ok {
+		config = make(map[string]interface{})
+	} else {
+		config = endpointConfigObj["config"].(map[string]interface{})
```
hbhoyar-uber, line 410: can you do 
`config, ok := ....`

codegen/module.go
```diff
@@ -83,6 +83,7 @@ type Options struct {
 	EnableCustomInitialisation bool
 	CommitChange               bool
 	QPSLevelsEnabled           bool
+	ProxyTemplates             *Template
```
groshu, line 86: maybe a better name like custom templates to make it generic ? just a thought no hard opinions

codegen/template.go
```diff
@@ -139,6 +142,24 @@ func NewTemplate(
...
+	defaultTemplate *Template,
+	proxyTemplate *Template,
+	config map[string]interface{},
+) (string, *Template) {
+	// verify if config contains template override
+	if _, ok := config[ProxyTemplate]; !ok {
+		return defaultTemplateName, defaultTemplate
+	}
+	// verify if override template config contains given template
+	templateMap := config[ProxyTemplate].(map[string]interface{})
```
hbhoyar-uber, line 156: can you do 
`templateMap, ok :=  ...`

codegen/module.go
```diff
@@ -1787,6 +1789,8 @@ type ModuleInstance struct {
 	mu                sync.RWMutex
 	// QPSLevels is map of circuit breaker name to qps level for all circuit breakers
 	QPSLevels map[string]int
+	// ProxyTemplates is a collection of proxy templates apart from zanzibar
```
groshu, line 1792: if using custom templates. we can write collection of custom templates that can be passed to zanzibar

hbhoyar-uber (changes requested):
Can you also add this new config to one of the examples to verify the change

groshu (commented):
can we please add unit tests ?

codegen/module.go
shivam-srivastava28: Done

codegen/module.go
shivam-srivastava28: Done

codegen/template.go
shivam-srivastava28: Same as above

shivam-srivastava28:
> 

Added unit test cases

groshu:
should we assert a few things like in the test template default template is not picked but passed custom template is picked ?
is it easy to do that ?

groshu (approved):

hbhoyar-uber (approved):

PR #868, opened 2023-04-14
Moved notFoundHandler init after observability tools init

hbhoyar-uber:
When notFoundHandler was initialised, it used to get logger and scope as nil. Hence, initialising notFoundHandler after all the observability tools are initialised


shivam-srivastava28 (approved):
LGTM

PR #875, opened 2023-05-30
performance optimisation: avoid map regrow calls

chirayu:


runtime/context.go
```diff
@@ -205,9 +206,10 @@ func WithScopeTags(ctx context.Context, newFields map[string]string) context.Con
 
 // GetScopeTagsFromCtx returns the tag info extracted from context.
 func GetScopeTagsFromCtx(ctx context.Context) map[string]string {
-	tags := make(map[string]string)
+	var tags map[string]string
```
hbhoyar-uber, line 209: Same comment as above

runtime/context.go
```diff
@@ -119,9 +119,10 @@ func WithEndpointRequestHeadersField(ctx context.Context, requestHeaders map[str
 
 // GetEndpointRequestHeadersFromCtx returns the endpoint request headers, if it exists on context
 func GetEndpointRequestHeadersFromCtx(ctx context.Context) map[string]string {
-	requestHeaders := make(map[string]string)
+	var requestHeaders map[string]string
```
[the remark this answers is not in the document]
chirayu: good catch..

runtime/context.go
```diff
@@ -119,9 +119,10 @@ func WithEndpointRequestHeadersField(ctx context.Context, requestHeaders map[str
 
 // GetEndpointRequestHeadersFromCtx returns the endpoint request headers, if it exists on context
 func GetEndpointRequestHeadersFromCtx(ctx context.Context) map[string]string {
-	requestHeaders := make(map[string]string)
+	requestHeaders := make(map[string]string, 0)
```
hbhoyar-uber, line 122: nit: 0 is default length hence not required to specify

runtime/context.go
```diff
@@ -205,9 +206,10 @@ func WithScopeTags(ctx context.Context, newFields map[string]string) context.Con
 
 // GetScopeTagsFromCtx returns the tag info extracted from context.
 func GetScopeTagsFromCtx(ctx context.Context) map[string]string {
-	tags := make(map[string]string)
+	tags := make(map[string]string, 0)
```
hbhoyar-uber, line 209: nit: same as above

hbhoyar-uber (approved):

runtime/context.go
chirayu: See https://pkg.go.dev/builtin#make . Go allocates a small size, if it is not specified.

PR #876, opened 2023-05-31
ContextMetrics slowness

chirayu:
ContextMetrics is a slow interface that is unfortunately used by many high throughput clients. This change provides an option to users to bypass it, and replaces it in calls internal to Zanzibar.

### Why is it not removed?
ContextMetrics is part of many Zanzibar API calls and structures. Removing it will break those interface.

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -437,7 +437,7 @@ type Logger interface {
 // Methods that update metrics create a new tally scope on each call. This is unnecessary slow in scenarios that update
 // multiple metrics.
 type ContextMetrics interface {
-	Scope() tally.Scope
+	Scope(ctx context.Context) tally.Scope
 	IncCounter(ctx context.Context, name string, value int64)
 	RecordTimer(ctx context.Context, name string, d time.Duration)
 	RecordHistogramDuration(ctx context.Context, name string, d time.Duration)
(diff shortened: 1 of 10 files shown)


runtime/context.go
```diff
@@ -434,7 +434,10 @@ type Logger interface {
 }
 
 // ContextMetrics emit metrics with tags extracted from context.
+// Methods that update metrics create a new tally scope on each call. This is unnecessary slow in scenarios that update
+// multiple metrics.
 type ContextMetrics interface {
+	Scope() tally.Scope
```
hbhoyar-uber, line 440: maybe we can pass ctx in Scope() which would return tagged scope.

Suggesting this as the name is context metrics

runtime/tchannel_outbound_call.go
```diff
@@ -62,15 +78,15 @@ func (c *tchannelOutboundCall) finish(ctx context.Context, err error) {
 		errCause := tchannel.GetSystemErrorCode(errors.Cause(err))
 		scopeTags := map[string]string{scopeTagError: errCause.MetricsKey()}
 		ctx = WithScopeTags(ctx, scopeTags)
```
hbhoyar-uber, line 80: These tags will not be emitted

hbhoyar-uber:
Suggestion:
Can define `Scope(ctx)` on `ContextMetrics`.
Wherever multiple metrics are emitted, we can get scope object from above API and emit those metrics.
This will also include tags added in context.

runtime/tchannel_outbound_call.go
```diff
@@ -61,16 +77,16 @@ func (c *tchannelOutboundCall) finish(ctx context.Context, err error) {
 	if err != nil {
 		errCause := tchannel.GetSystemErrorCode(errors.Cause(err))
 		scopeTags := map[string]string{scopeTagError: errCause.MetricsKey()}
-		ctx = WithScopeTags(ctx, scopeTags)
-		c.metrics.IncCounter(ctx, clientSystemErrors, 1)
```
hbhoyar-uber, line 65: Maybe we can add the following in  `WithScopeTags` function
```
scope := GetScope(ctx)
scope = scope.Tagged(tags)
ctx = WithScope(ctx, scope)
```

PR #880, opened 2023-06-26
restoring NewClientHTTPRequest and MakeTChannelRequest fn signatures for backward compatibility

chirayu:


bishnuag:
there are test cases that passes timeoutConfig and now need to update them , one such example - https://sourcegraph.uberinternal.com/github.com/uber/zanzibar/-/blob/examples/example-gateway/endpoints/bar/bar_test.go#L58

Other can be seen in here https://github.com/uber/zanzibar/actions/runs/5377416059/jobs/9755855101?pr=880

chirayu:
> there are test cases that passes timeoutConfig and now need to update them ,

fixed

bishnuag:
fix lints ->  lints are failing due to the wrong ordering of imports

bishnuag (approved):
LGTM

PR #881, opened 2023-06-27
Add Error wrapper to enable error tagging

amolmejari2:
Error wrapper will be used to tag errors with errorType and errorLocation which can be used in logging.


codegen/templates/workflow.tmpl
```diff
@@ -227,34 +229,31 @@ func (w {{$workflowStruct}}) Handle(
...
+			err = zErr.Unwrap()
+		}
 		switch errValue := err.(type) {
 			{{range $idx, $cException := $clientExceptions}}
 			case *{{$cException.Type}}:
-				serverErr := convert{{$methodName}}{{title $cException.Name}}(
+				err = convert{{$methodName}}{{title $cException.Name}}(
 					errValue,
 				)
-				{{if eq $responseType ""}}
```
chirayu, line 236: reason behind removing this block?

chirayu:
Pls add/modify a gateway test to check for the error. This is to ensure that the codgen output works as desired

codegen/templates/endpoint.tmpl
```diff
@@ -217,7 +217,10 @@ func (h *{{$handlerName}}) HandleRequest(
 	}
 
 	if err != nil {
-		{{- if eq (len .Exceptions) 0 -}}
+		if zErr, ok := err.(zanzibar.Error); ok {
+			err = zErr.Unwrap()
+		}
+		{{if eq (len .Exceptions) 0 -}}
```
hbhoyar-uber, line 223:
```suggestion
		{{- if eq (len .Exceptions) 0 -}}
```

PR #882, opened 2023-06-29
Log fields cleanup

amolmejari2:
The changes include:
1. Redundant log fields are removed
- timestamp-started, timestamp-finished: removed since these timestamps aren't useful in logs.
-  clientHTTPMethod, url: removed since this is part of client config.
-  endpoint: duplicates endpointId
2. Replaced overridden fields with new fields.
-  statusCode (used for endpoint response status code) -> client_status_code
-  remoteAddr (used for incoming request remoteAddr) -> client_remote_addr


chirayu:
Pls expand on the comment to 1) point out the replacement fields for the redundant ones and 2) the overridden fields.

chirayu (approved):

PR #907, opened 2023-09-01
Expose tracer& tracer closer options in gateway

ChenX1993:
Today zanzibar creates its own Jaeger Tracer in the gateway component. However at Uber, we use a more customized tracer, and we want to make sure services that consume zanzibar use the same tracer.

Expose the tracer& tracer closer options to create gateway so that we are able to inject the customized tracer.

### DIFF
--- a/runtime/gateway.go
+++ b/runtime/gateway.go
@@ -81,6 +81,8 @@ type Options struct {
 	GetContextFieldExtractors func() []ContextLogFieldsExtractor
 	JSONWrapper               jsonwrapper.JSONWrapper
 	NotFoundHandler           func(*Gateway) http.HandlerFunc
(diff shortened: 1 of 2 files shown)


runtime/gateway.go
```diff
@@ -248,8 +250,13 @@ func CreateGateway(
 		return nil, err
 	}
 
-	if err := gateway.setupTracer(config); err != nil {
-		return nil, err
+	if opts.Tracer != nil && opts.TracerCloser != nil {
```
groshu, line 253: can you please make it config driven as done below for not found handler

ChenX1993: Thanks! Updated

groshu (approved):

PR #908, opened 2023-09-15
Ignore case while detecting ReqUUID

sean1006:
Previously, case sensitivity could cause mis-matches in detecting the presence of a ReqUUID which ends up in Zanzibar populating an additional ReqUUID. This addresses the ReqUUID issue but does not make headers in the tcahnnel_server fully case insensitive, as there are further use cases which must be addressed first.

### DIFF
--- a/runtime/tchannel_server.go
+++ b/runtime/tchannel_server.go
@@ -23,6 +23,7 @@ package zanzibar
 import (
 	"context"
 	"fmt"
+	"net/textproto"
 	"strings"
 	"sync"
 
@@ -206,10 +207,7 @@ func (s *TChannelRouter) handleHeader(
 		return ctx, err
 	}
 
-	reqUUID, ok := c.reqHeaders[s.requestUUIDHeaderKey]
-	if !ok {
-		reqUUID = uuid.New()
-	}
+	reqUUID := getRequestUUID(s.requestUUIDHeaderKey, c.reqHeaders)
 	ctx = withRequestUUID(ctx, reqUUID)
 
 	// put request headers on context so that user-provided extractor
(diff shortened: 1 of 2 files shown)


runtime/tchannel_server.go
```diff
@@ -206,7 +206,13 @@ func (s *TChannelRouter) handleHeader(
 		return ctx, err
 	}
 
-	reqUUID, ok := c.reqHeaders[s.requestUUIDHeaderKey]
+	// TODO: Make headers case insensitive. Remove instances of headers being used in a case
```
rpatali: can you write a test?

runtime/tchannel_server.go
```diff
@@ -206,7 +206,13 @@ func (s *TChannelRouter) handleHeader(
 		return ctx, err
 	}
 
-	reqUUID, ok := c.reqHeaders[s.requestUUIDHeaderKey]
```
rpatali, line 209: can you simply not use:
```
reqUUID, ok := c.reqHeaders[textproto.CanonicalMIMEHeaderKey(s.requestUUIDHeaderKey)]
```

sean1006: unfortunately, we need to standardize input map keys

runtime/tchannel_server.go
sean1006: will update

runtime/tchannel_server.go
```diff
@@ -300,3 +297,18 @@ func DecorateWithRecover(
...
+// uuid if no such UUID exists in request headers
+func getRequestUUID(
+	reqUUIDHeaderKey string,
+	reqHeaders map[string]string
+) string {
+	canonicalHeaderKey := textproto.CanonicalMIMEHeaderKey(s.requestUUIDHeaderKey)
+	for key, value := range reqHeaders {
+		if key == canonicalHeaderKey && value != "" {
+			return value
+		}
```
rpatali: `reqHeaders` is a map :)

runtime/tchannel_server_test.go
```diff
@@ -93,3 +94,44 @@ func (m mockRWTStruct) ToWire() (wire.Value, error) {
...
+
+func TestGetRequestUUID(t *testing.T) {
+	uuidHeaderKey := "test-uuid-key"
+	testUuidValue := uuid.New()
+
+	testCases := map[string]struct {
+		reqHeaders 		map[string]string
+		expectedEquals  string
+	}{
+		"Success: return new UUID when no request header map is given": {},
```
rpatali, line 106: when null header map is given as compared to empty map given in the next case is better naming

rpatali (approved):
LGTM

PR #909, opened 2023-10-10
Expose tracer& tracer closer options in gateway

ChenX1993:
Today zanzibar creates its own Jaeger Tracer in the gateway component. However at Uber, we use a more customized tracer, and we want to make sure services that consume zanzibar use the same tracer.

Expose the tracer& tracer closer options to create gateway so that we are able to inject the customized tracer.


chirayu (approved):
lgtm

PR #913, opened 2023-10-27
Event generation 

chirayu:
This change adds support for event generation and sampling. Specifically:

1.  request/response exchanges on various interfaces: incoming http, outgoing thrift, outgoing http
2. User provided sampliing and event handler functions


runtime/router.go
```diff
@@ -128,8 +132,41 @@ func (endpoint *RouterEndpoint) HandleRequest(
...
+		var events []Event
+		if ec != nil {
+			events = append(events, ec.events...)
+		}
+
+		event := &HTTPIncomingEvent{
+			EndpointName: endpoint.EndpointName,
+			HandlerName:  endpoint.HandlerName,
+			HTTPCapture: HTTPCapture{
+				ReqURL:        r.URL.String(),
```
groshu, line 163: i think we should clone req headers before handlerFn

chirayu: Headers weren't being mutated in the endpoint handler. However, it could change in the future. Capturing early is better. Have made the change.

groshu (approved):

PR #928, opened 2024-02-28
Adding more tags for panic metrics 

anirudhraja:
EndpointID and HandlerID tags if present will help debug panic metrics with ease. Adding these tags in this diff.


runtime/router.go
```diff
@@ -301,8 +301,14 @@ func (router *httpRouter) handlePanic(
 	}
 	logger := router.gateway.ContextLogger
 	logger.Error(r.Context(), "A http request handler paniced", zap.Error(err), zap.Int(logFieldResponseStatusCode, http.StatusInternalServerError))
-	router.panicCount.Inc(1)
-
+	m := router.gateway.ContextExtractor.ExtractScopeTags(r.Context())
```
shivam-srivastava28: Can a gateway be nil resulting in nil pointer?

shivam-srivastava28 (approved):
LGTM

PR #929, opened 2024-02-28
grpc outgoing event

shivam-srivastava28:


chirayu (commented):
1. Shouldn't the headers be map[string][]string?
2. How will these events be used?

shivam-srivastava28:
> 1. Shouldn't the headers be map[string][]string?
> 2. How will these events be used?

1. Headers returned from yarpc is in map[string]string format.
2. These events will be used in edge-gateway by a yarpc middleware.

chirayu (approved):

PR #934, opened 2024-04-17
Add rpc.status to client response context for annotated thrift exceptions

deepanshumehndiratta:
This change adds the response status to the returned context from tchannel clients by inspecting system errors, as well as annotated applications errors (thrift exceptions)

### DIFF
--- a/codegen/method.go
+++ b/codegen/method.go
@@ -47,8 +47,32 @@ const (
 	antHTTPResNoBody   = "%s.http.res.body.disallow"
 )
 
+const _errorCodeAnnotationKey = "rpc.code"
+
 const queryAnnotationPrefix = "query."
 const headerAnnotationPrefix = "headers."
+
+var (
+	_gRPCCodeNameToYARPCErrorCodeType = map[string]string{
+		// https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+		"CANCELLED":           "yarpcerrors.CodeCancelled",
+		"UNKNOWN":             "yarpcerrors.CodeUnknown",
+		"INVALID_ARGUMENT":    "yarpcerrors.CodeInvalidArgument",
+		"DEADLINE_EXCEEDED":   "yarpcerrors.CodeDeadlineExceeded",
+		"NOT_FOUND":           "yarpcerrors.CodeNotFound",
+		"ALREADY_EXISTS":      "yarpcerrors.CodeAlreadyExists",
+		"PERMISSION_DENIED":   "yarpcerrors.CodePermissionDenied",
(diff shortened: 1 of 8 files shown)


codegen/templates/tchannel_client.tmpl
```diff
@@ -362,10 +382,11 @@ type {{$clientName}} struct {
 				success, respHeaders, clientErr = c.client.Call(
 					ctx, "{{$svc.Name}}", "{{.Name}}", reqHeaders, args, &result,
 				)
-				if _, isSysErr := clientErr.(tchannel.SystemError); !isSysErr {
+				if isSystemError(clientErr) {
```
hbhoyar-uber, line 385: behaviour is changed here

codegen/method.go
```diff
@@ -415,15 +445,36 @@ func (ms *MethodSpec) setExceptions(
 			)
 		}
 
+		errorCode := ""
+
+		for _, t := range ms.CompiledModuleSpec.Types {
+			if _, ok := t.(*compile.StructSpec); ok {
+				if t.ThriftName() ==  e.Type.ThriftName() {
+					if errorCodeString, ok := t.ThriftAnnotations()[_errorCodeAnnotationKey]; ok {
```
hbhoyar-uber: You could directly do e.Type.ThriftAnnotations()

codegen/templates/tchannel_client.tmpl
hbhoyar-uber: behaviour is changed here

codegen/templates/tchannel_client.tmpl
```diff
@@ -411,3 +441,34 @@ type {{$clientName}} struct {
...
+}
+
+func setContextStatusCode(ctx context.Context, code yarpcerrors.Code) context.Context {
+	if ctx != nil {
+		if statusCode := ctx.Value(_errorCodeAnnotationKey); statusCode == nil {
+			ctx = context.WithValue(ctx, _errorCodeAnnotationKey, code)
+		}
+	}
+	return ctx
+}
```
hbhoyar-uber: Can you move this to runtime folder. This is common code and would just increase binary size being present in all generated code files.

hbhoyar-uber:
Can you also add the test which shows generated code setting up this annotation

deepanshumehndiratta:
> Can you also add the test which shows generated code setting up this annotation

Added a unit test to cover this

hbhoyar-uber (approved):
Nice! Looks like you refactored unnecessary code as well.

hbhoyar-uber (approved):

PR #942, opened 2025-06-26
[bugfix] Fix for data race in PascalCase

dragneelfps:
when using PascalCase in parallel constructs, the existing code can lead to data races. 
Using once.Do will ensure the initialization happens only once.


codegen/casing.go
```diff
@@ -29,7 +29,10 @@ import (
 	"unicode/utf8"
 )
 
-var pascalCaseMap *sync.Map
+var (
+	pascalCaseMap *sync.Map
```
chabbimilind, line 33: Why not just use `pascalCaseMap sync.Map`, which will create a sync.Map and we don't need sync.Once and no need to check for nil.

